Issue #29, opened 2024-03-14
Supporting equality checking in a way xarray understands


TomNicholas:
It would be useful (especially for testing purposes) to be able to compare xarray datasets containing ManifestArrays for equality.

It's clear when two `ManifestArray` objects should be considered equivalent - they have the exact same chunkmanifest and zarray info. However, just simply returning this condition from `.__eq__` has a few issues:

1) The array API standard expects `.__eq__` to return an entire array of booleans rather than a single scalar. I could make that array, but it seems like a weird thing to do, as you can't actually access the individual elements (only indirect representations of individual chunks). Actually, I could even go further and define equality on a per-chunk basis, returning a numpy array of booleans that would have `False` at any locations corresponding to non-equivalent `ChunkEntry` objects in the manifest... But you can't then use this mask to index the array. It also seems like the array API standard expects the returned array type to be `Self`, i.e. another `ManifestArray`, not `np.ndarray`.

2) Xarray checks equivalence of wrapped arrays in a complicated lazy NaN-aware way that gets tripped up by `ManifestArray` objects. Following the stack down from `xarray.testing.assert_equal`, it goes through `Variable.equals`, [`duck_array_ops.array_equiv`](https://github.com/pydata/xarray/blob/80fb5f7626b2273cffe9ccbe425beaf36eba4189/xarray/core/duck_array_ops.py#L284), and [calls `duck_array_ops.isnull`](https://github.com/pydata/xarray/blob/80fb5f7626b2273cffe9ccbe425beaf36eba4189/xarray/core/duck_array_ops.py#L292). At this point as `ManifestArray` doesn't define any form of `isnull`, an `AttributeError` is raised, which is [caught in `Variable.equals` here](https://github.com/pydata/xarray/blob/80fb5f7626b2273cffe9ccbe425beaf36eba4189/xarray/core/variable.py#L1745), and `False` is always returned as the result of the equality assertion. 

The problematic section is

```python
flag_array = (arr1 == arr2) | (isnull(arr1) & isnull(arr2))
return bool(flag_array.all())
```

If this condition were just `(arr1 == arr2)` then we would be fine...

Inside [`duck_array_ops.isnull`](https://github.com/pydata/xarray/blob/80fb5f7626b2273cffe9ccbe425beaf36eba4189/xarray/core/duck_array_ops.py#L142) theres a weird part that I don't understand. Copied verbatim:

```python
scalar_type = data.dtype.type
...
[...]

PR #30, opened 2024-03-14
Equality checking

TomNicholas:
Work towards closing #29.

This makes `ManifestArray.__eq__` return a numpy array of bools, but it falls short of doing a full element-wise comparison. Instead it just returns all Falses if any of the chunk entries are different.

#31 is to implement the full version of element-wise checking.


TomNicholas:
This seems to be enough to make xarray asserts return True or False correctly, so it mostly closes #29.

PR #31, opened 2024-03-14
Partial equality checking

TomNicholas:
WIP towards actually doing full element-wise equality checking between `ManifestArray` objects.


dcherian:
Can this be avoided with sensible default options for `concat`? `data_vars="minimal", coords="minimal", compat="override"`?

TomNicholas:
> Can this be avoided with sensible default options for `concat`? `data_vars="minimal", coords="minimal", compat="override"`?

This `__eq__` method isn't actually used during `concat` I don't think. I just implemented it so that I could call `xrt.assert_equal` on virtual datasets. This PR would make the result more useful in cases where some but not all of the chunks in two manifests equivalent.

Issue #2, opened 2024-03-08
Initial release checklist


TomNicholas:
What's here so far is a work-in-progress. I will consider this issue closed once the example in the Readme runs successfully.

For this we need:

- [x] Definition of a `ChunkManifest`
- [x] Definition of a `ManifestArray`
- [x] Concatenation of `ManifestArrays` through merging their manifests
- [x] Creation of `ManifestArrays` from kerchunk reference dicts
- [x] `ds_from_kerchunk_reference_dict` function
- [x] `open_dataset_via_kerchunk` function to read byte ranges and translate them to `ManifestArray` objects
- [x] Correct handling of concatenation in xarray (#34)
- [x] Option to avoid creating pesky indexes (see #14)
- [x] Xarray accessor to reconstruct kerchunk reference dicts from merged manifests (https://github.com/TomNicholas/VirtualiZarr/pull/28)
- [x] Serialize the dicts to disk as json

This list would get us the ability to do this:

```python
from virtualizarr import open_dataset_via_kerchunk

datasets = [open_dataset_via_kerchunk(filepath, no_indexes=True) for filepath in files]

combined = xr.combine_nested(datasets)

combined.virtualize.to_kerchunk(format='json')

# now can read using kerchunk mapper like normal
```

For the initial version we do _not_ need:
- [ ] Xarray backend entrypoint plugin (this is just syntactic sugar) (#35)
- [ ] Automatic inferring of concatenation order (see #18)
- [x] Writing out to parquet (https://github.com/TomNicholas/VirtualiZarr/issues/72)
- [x] Automatic detection of different file formats (see #10)
- [ ] Anything that requires zarr development, e.g. v3 support, variable-length chunking, concatenation of different codecs

Without this latter list we notably can't do `combine_by_coords` and we can't do `open_dataset(filepath, engine='virtualizarr')`.

TomNicholas:
(I just updated this to be much clearer about what the "MVP" involves)

TomNicholas:
I'm ready to release this, but @MSanKeys963 says we can move this repo into the [`zarr-developers`](https://github.com/zarr-developers) github org, and it would be great to do that move first 🙏

jbusecke:
Adding a +1 here to move and release. I am working with @sashakames to prototype virtualizarr usage [within ESGF](https://github.com/jbusecke/esgf-virtual-zarr-data-access) and It would be great if we could pin versions.

TomNicholas:
[...]

PR #34, opened 2024-03-15
Support xarray concat, including broadcasting

TomNicholas:
Closes part of #2.

To get this to work I had to add broadcasting, to cover times when xarray wants to insert a length-1 dimension. I did appreciate until now that actually we can go further and fully support broadcasting with `ManifestArray` objects!

### DIFF
--- a/virtualizarr/manifests/array_api.py
+++ b/virtualizarr/manifests/array_api.py
@@ -1,4 +1,5 @@
-from typing import TYPE_CHECKING, Callable, Dict, Iterable, List, Tuple, Union
+import itertools
+from typing import TYPE_CHECKING, Callable, Dict, Iterable, List, Tuple, Union, cast
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


virtualizarr/manifests/array_api.py
```diff
@@ -239,6 +240,50 @@ def _check_same_shapes(shapes: List[Tuple[int, ...]]) -> None:
...
+    # this is just a special case of stacking
+    return stack([x], axis=axis)
+
+
+@implements(np.broadcast_to)
+def broadcast_to(x: "ManifestArray", /, shape: Tuple[int, ...]) -> "ManifestArray":
+    """
+    Broadcasts an array to a specified shape, by either manipulating chunk keys or copying chunk manifest entries.
+    """
+    print(f"requested broadcasting array from shape {x.shape} to shape {shape}")
```
rabernat, line 255: Replace with log statement?

TomNicholas: Oops thanks for catching that! Removed in https://github.com/TomNicholas/VirtualiZarr/commit/c0ef78c93b1745218ce4286b25f9d83ee50f860b

Issue #22, opened 2024-03-11
Padding arrays with NaNs


TomNicholas:
We want to be able to pad arrays with `NaN` values (/ zarr array `fill_value`) because some kerchunk use cases involve padding data (e.g. satellite swathes that produce slightly differently shaped data each day such as NASA SWOT).

TomNicholas:
We should probably be able to make the xarray `.pad` method work, at least with very specific arguments, e.g. `ds.pad(x=1, mode='constant')`. We would then intercept the arguments and re-interpret the padding operation for however the underlying manifest actually works.

dcherian:
> satellite swathes that produce slightly differently shaped data

How will this work, since each chunk will require a different padding (presumably)? Are you going to add a new codec?

For example consider three arrays with shapes (1,1), (1,2), (1,3), stacked along axis=0.

ayushnag:
I did use kerchunk for SWOT but I wasn't able to pad them so I only used the granules with the most common swath length. This was for a demo though and ideally there is some way to pad with nan's.

TomNicholas:
> For example consider three arrays with shapes `(1,1), (1,2), (1,3)`, stacked along axis=0.

To be concrete, if the chunksize for all these three arrays is `(1,1)`, then empty(/nonexistent) chunks can be padded to each array in turn no problem such that the shapes match and the stacking becomes possible. I agree that if the chunksizes are actually `(1,1), (1,2), (1,3)` then it's a lot harder.

dcherian:
> I agree that if the chunksizes are actually (1,1), (1,2), (1,3) then it's a lot harder.

IIRC this is a good model for the swath problem. I think the (only?) solution is a new codec since we need to transform the chunk itself.  Something like:
```python
chunk: zarr.Array
out = np.full_like(new_shape, fill_value=chunk.fillvalue)
slicer = tuple(slice(size) for size in chunks.shape)
out[slicer] = np.array(chunk)
```

TomNicholas:
> think the (only?) solution

I guess in this case even variable-length chunks wouldn't save us because this would actually require "ragged" chunks (i.e. no longer a rectilinear array of chunks)...

> IIRC this is a good model for the swath problem. I think the (only?) solution is a new codec since we need to transform the chunk itself. Something like:

[...]

PR #39, opened 2024-03-17
[WIP] Structured array for manifest

### LINKED ISSUE #33: In-memory representation of chunks: array instead of a dict?
Currently chunks are stored as a mapping of chunk keys to chunk entries, i.e. a manifest dict. The `ManifestArray` is just a thin wrapper over this, it's not really an array at all internally. The main purpose of it is to do lazy concatenation, which is implemented via manipulating manifest dicts.

However, as was pointed out by @dcherian in https://github.com/fsspec/kerchunk/issues/377#issuecomment-1799115319, a lazily concatenated array is essentially just a chunked array. We could imagine an alternative design where `ManifestArray` works more like a dask array, which holds a grid of chunks, each of which is itself an array (i.e. numpy arrays).

It might make sense to re-implement `ManifestArray` to organise chunks in the same way dask does, perhaps storing some sort of `ChunkReferenceArray` object in place of numpy arrays. This would then handle concatenation at the chunk level, allow for indexing (as long as the indexers aligned with chunk boundaries), and possibly be implemented by vendoring code from inside dask.

In this design the `ChunkManifest` would be something that could be built from the `ManifestArray` when needed, not the fundamental object. This change could be done without changing the existing API of `ManifestArray`.

TomNicholas:
Aims to close #33 by using a numpy array with a structured dtype to store the `(path, offset, length)` information for each chunk.

EDIT: Should add

- [ ] `broadcast_to`
- [ ] `empty_like`


virtualizarr/manifests/manifest.py
```diff
@@ -42,11 +46,20 @@ def to_kerchunk(self) -> List[Union[str, int]]:
         return [self.path, self.offset, self.length]
 
 
+# TODO we want the path field to contain a variable-length string, but that's not available until numpy 2.0
+# See https://numpy.org/neps/nep-0055-string_dtype.html
+MANIFEST_STRUCTURED_ARRAY_DTYPES = np.dtype(
+    [("path", "<U32"), ("offset", np.int32), ("length", np.int32)]
+)
```
TomNicholas, line 53: Because file paths can be strings of any length, we really need to be using numpy's new [variable-width string dtype](https://numpy.org/neps/nep-0055-string_dtype.html) here.

Unfortunately it's only coming out with numpy 2.0, and although there is a release candidate for numpy 2.0, it's so new that [pandas doesn't support it yet](https://github.com/pandas-dev/pandas/issues/55519). Xarray has a pandas dependency, so currently we can't actually build an environment that let's us try virtualizarr with the variable-length string dtype yet.

virtualizarr/manifests/manifest.py
```diff
@@ -165,99 +228,3 @@ def get_chunk_grid_shape(chunk_keys: Iterable[ChunkKey]) -> Tuple[int, ...]:
         max(indices_along_one_dim) + 1 for indices_along_one_dim in zipped_indices
     )
     return chunk_grid_shape
-
-
-def check_keys_form_grid(chunk_keys: Iterable[ChunkKey]):
-    """Check that the chunk keys collectively form a complete grid"""
```
TomNicholas, line 171: Lines 170-263 are what we get rid of by doing concatenation/stacking via the wrapped structured array.

virtualizarr/manifests/manifest.py
```diff
@@ -165,99 +228,3 @@ def get_chunk_grid_shape(chunk_keys: Iterable[ChunkKey]) -> Tuple[int, ...]:
...
-    all_required_chunk_keys: set[ChunkKey] = set(
-        join(inds) for inds in all_possible_combos
-    )
-
-    # check that every possible combination is represented once in the list of chunk keys
-    if set(chunk_keys) != all_required_chunk_keys:
-        raise ValueError("Chunk keys do not form a complete grid")
-
-
-def concat_manifests(manifests: List["ChunkManifest"], axis: int) -> "ChunkManifest":
```
TomNicholas, line 188: Lines 188-263 are what we get rid of by doing concatenation/stacking via the wrapped structured array.

virtualizarr/manifests/manifest.py
```diff
@@ -14,6 +13,9 @@
 _CHUNK_KEY = rf"^{_INTEGER}+({_SEPARATOR}{_INTEGER})*$"  # matches 1 integer, optionally followed by more integers each separated by a separator (i.e. a period)
 
 
+ChunkDict = NewType("ChunkDict", dict[ChunkKey, dict[str, Union[str, int]]])
+
+
 class ChunkEntry(BaseModel):
```
TomNicholas, line 19: Not sure we really need this class anymore

TomNicholas:
As with this approach array entries with empty strings for paths are interpreted as chunks missing from the zarr array, we could easily solve #22 just by having an `empty_like` creation function that creates an array with only empty paths.

virtualizarr/manifests/manifest.py
TomNicholas: Pandas just released [2.2.2](https://pandas.pydata.org/docs/whatsnew/v2.2.2.html), which is compatible with the upcoming numpy 2.0 release.

Not sure if that will break any part of xarray that we need for VirtualiZarr, but this might now be close enough to test out variable-length dtypes now.

TomNicholas:
Superceded by #107

PR #43, opened 2024-03-20
Adds netCDF3 vs netCDF4 distinction to _automatically_determine_filetype.

norlandrhagen:
Curious on what you think of this @TomNicholas. 

- Allows _automatically_determine_filetype to guess between netCDF3 vs netcdf4.
- Adds a `netCDF4` dependency to tests. 🤷

### DIFF
--- a/virtualizarr/kerchunk.py
+++ b/virtualizarr/kerchunk.py
@@ -1,3 +1,4 @@
+import importlib.util
 from pathlib import Path
 from typing import List, NewType, Optional, Tuple, Union, cast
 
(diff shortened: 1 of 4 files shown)


virtualizarr/kerchunk.py
```diff
@@ -61,10 +62,26 @@ def _automatically_determine_filetype(filepath: str) -> str:
...
+
+        import netCDF4
+
+        with netCDF4.Dataset(filepath, "r") as dataset:
+            if dataset.data_model == "NETCDF4":
+                filetype = "netCDF4"
+            elif dataset.data_model == "NETCDF3_CLASSIC":
+                filetype = "netCDF3"
+            else:
+                raise NotImplementedError(
```
TomNicholas, line 81: This is cool! Presumably this has to open the file to read this `data_model` information? Is that likely to incur any significant cost over the opening that `kerchunk` already needs to do?

norlandrhagen: I think the read done by the NetCDF4 library is only reading attributes and not loading the entire file. Not 100% sure on this though.

rabernat: You can do this without opening the file with netCDF at all. Both file types have a "magic" at the beginning of the file.

```python
def guess_file_type(fp) -> FileType:
    magic = fp.read(4)
    fp.seek(0)
    if magic[:3] == b"CDF":
        return FileType.netcdf3
    elif magic == b"\x89HDF":
        return FileType.hdf5
    else:
        raise ValueError(f"Unknown file type - magic {magic}")
```

PR #44, opened 2024-03-20
Test concat of dimension coordinate not backed by an index

TomNicholas:
Adds a test showing a problem in xarray - if I build two datasets that have dimension coordinates which are explicitly not backed by an index (which is possible by passing the new `xr.Coordinates` object explicitly to the constructor), then upon concatenating these, xarray sees the dimension coordinate and attempts to automatically rebuild a new index (which will fail in my case). I need to be able to turn off that behaviour somehow.


TomNicholas:
The simplest test that shows this issue is described in https://github.com/pydata/xarray/issues/8871

TomNicholas:
This PR now uses a branch on my fork of xarray, which contains the (unmerged upstream) PRs https://github.com/pydata/xarray/pull/8711, https://github.com/pydata/xarray/pull/8714 and https://github.com/pydata/xarray/pull/8872. All three of these are needed to get the new test in this PR to pass.

TomNicholas:
I'm going to merge this, making this package depend on my branch of xarray. Once those changes are merged upstream I can upin that, see https://github.com/TomNicholas/VirtualiZarr/issues/14#issuecomment-2018369470.

PR #45, opened 2024-03-22
Write manifests to zarr store

### LINKED ISSUE #6: Serialization to a Zarr chunk manifest
We can write methods to serialize the in-memory chunk manifest objects to json on-disk, but it's not much use until a Zarr reader can understand them. See https://github.com/zarr-developers/zarr-specs/issues/287.

TomNicholas:
Closes #6

This shows how we could write an xarray Dataset containing `ManifestArray` objects to disk as a new zarr store, where each array's chunk data is written as byte range references in the form of a `manifest.json` file.

This therefore creates an example of the type of zarr store described in https://github.com/zarr-developers/zarr-specs/issues/287 by @jhamman.

(Currently this writes using the V2 spec, which I guess is not correct, but you get the idea.)


virtualizarr/zarr.py
```diff
@@ -96,3 +106,67 @@ def ceildiv(a: int, b: int) -> int:
...
+    See https://github.com/zarr-developers/zarr-specs/issues/287
+    """
+
+    from virtualizarr.manifests import ManifestArray
+
+    _storepath = Path(storepath)
+    Path.mkdir(_storepath, exist_ok=False)
+
+    # TODO should techically loop over groups in a tree but a dataset corresponds to only one group
+    # TODO does this mean we need a group kwarg?
```
jhamman, line 125: As long as your storepath argument is the root of the group, this should work fine.

virtualizarr/zarr.py
```diff
@@ -96,3 +106,67 @@ def ceildiv(a: int, b: int) -> int:
...
+
+    consolidated_metadata: dict = {"metadata": {}}
```
jhamman, line 127: small wrinkle for v3 -- we don't have a consolidated metadata spec extensions yet. No harm in putting this in for now but v3 readers are not going to know what to do here until we formalize this a bit.

jhamman: I think you could also forgo accumulating the consolidated metadata as you do here, and instead, just call `zarr.consolidate_metadata(store)` on the store after you've written all the metadata docs.

virtualizarr/zarr.py
```diff
@@ -96,3 +106,67 @@ def ceildiv(a: int, b: int) -> int:
...
+
+    # write top-level .zattrs
+    with open(_storepath / ".zattrs", "w") as json_file:
+        json.dump(ds.attrs, json_file, indent=4, separators=(", ", ": "))
```
jhamman, line 131: You might consider using `zarr.util.json_dumps` here. It would be a bit cleaner and more or less guarantees that you will be able to encode all valid zarr metadata. For example, it uses a custom number encoder to encode some dtypes.

jhamman (commented):
Very cool @TomNicholas -- This will be very helpful in getting a target for the zarr-python work to come.

xref: https://github.com/zarr-developers/zarr-python/issues/1718

virtualizarr/zarr.py
TomNicholas: I think in that case I should probably take out the consolidated metadata part and just make this into a bare-bones implementation of a "chunk manifest ZEP"-compliant store.

virtualizarr/zarr.py
TomNicholas: @jhamman cool - sounds like this package should have an optional dependence on zarr-python for utility functions like that then.

@cisaacstern why not? 😄 I won't put that in this PR, but it can be a follow-up (tracked by #46).

TomNicholas:
I realized that I could also add a reader for this type of store which creates a ManifestArray-backed dataset from a chunk-manifest-ZEP-compliant store. You could then use VirtualiZarr to combine the chunks from multiple such stores.

TomNicholas:
This PR now also adds the ability to open a zarr v3 store with all arrays as `manifest.json` files via `open_virtual_dataset(storepath, filetype="zarr_v3")`. I did this partly so that then I can test via roundtripping.

virtualizarr/vendor/zarr/utils.py
```diff
@@ -0,0 +1,22 @@
...
+        # See json.JSONEncoder.default docstring for explanation
+        # This is necessary to encode numpy dtype
+        if isinstance(o, numbers.Integral):
+            return int(o)
+        if isinstance(o, numbers.Real):
+            return float(o)
+        return json.JSONEncoder.default(self, o)
+
+
+def json_dumps(o: Any) -> bytes:
```
TomNicholas, line 18: I chose to vendor this because I didn't want to import the zarr-python library when it's in flux, and also this helps make it clear exactly which parts of this package even need `zarr-python` at all.

virtualizarr/xarray.py
```diff
@@ -50,37 +51,88 @@ def open_virtual_dataset(
...
-        filetype=filetype,
-    )
+    if drop_variables is None:
+        drop_variables = []
+
+    if virtual_array_class is not ManifestArray:
+        raise NotImplementedError()
+
+    if filetype == "zarr_v3":
+        # TODO is there a neat way of auto-detecting this?
```
TomNicholas, line 61: This is a bit ugly - I want to automatically distinguish between non-zarr, zarr v2 (to be read using kerchunk) and zarr v3 (to be read using this code). I guess I will just have to search for `.zgroup`/`zarr.json` files explicitly?

virtualizarr/xarray.py
```diff
@@ -50,37 +51,88 @@ def open_virtual_dataset(
...
+    """
+    Read a Zarr v3 store and return an xarray Dataset containing virtualized arrays.
+    """
+    _storepath = Path(storepath)
+
+    ds_attrs = attrs_from_zarr_group_json(_storepath / "zarr.json")
+
+    # TODO recursive glob to create a datatree
+    vars = {}
+    for array_dir in _storepath.glob("*/"):
```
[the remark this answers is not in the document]
TomNicholas: It seems that when run locally (on MacOS), a `pathlib.Path.glob("*/")` call only returns  directories (as the pathlib docs say it will), but for some reason when run in this CI the glob will include files too. I've hacked around this by excluding any paths for which `.is_file()` is True.

virtualizarr/xarray.py
TomNicholas: @norlandrhagen do you have any thoughts on a neat way to handle this?

TomNicholas:
I'm just going to merge this as we can always change it later.

Issue #8, opened 2024-03-08
Better documentation


TomNicholas:
This library is supposed to have an API so small it's practically non-existent (everything being done through xarray instead), but we still probably need documentation that's more than just the Readme.

For example a detailed explanation of when to use `combine_nested` vs `combine_by_coords`, or how to use `preprocess` to order the datasets using the name of the file.

TomNicholas:
I'm thinking about how the documentation of this package should be structured. It's tricky because there are multiple concepts, it requires understanding xarray's model pretty well, and not everything works yet.

We need narrative documentation explaining:

- What Kerchunk reads from files
- Storage of these byte ranges in a chunk manifest
- Wrapping of these in a ManifestArray
- Wrapping of those inside xarray objects
- Concatenation via xarray without indexes
- Concatenation via xarray with indexes
- Doing all the above in one go using open_mfdataset
- Writing out to Kerchunk
- Eventually writing out to a Zarr manifest
- Reading using fsspec

This can be adapted from the existing content in the readme.

We also some example-based docs that can be treated more as recipes. Those should cover:

- Concatenation along one dimension
- Concatenation along multiple dimensions
- Concatenating staggered grids?
- Concatenating in order determined by information in file names/attributes
- Reading the written references using fsspec
- Adjusting metadata in references (e.g. altering paths to be S3 urls rather than local paths if you move the archival files to the cloud)

We also need some developer docs:
- API reference, even for a lot of internals
- Relation to Kerchunk, and motivation for this being a separate project to Kerchunk
- Discussion about Kerchunk file format
- Discussion of virtual concatenation in Zarr as an intended way to eventually concatenate arrays with different codecs
- Discussion of the status of variable-length chunking

Some of those might be better placed under an FAQ page.

okz:
Would add somewhere that "Only local and s3 file protocols are currently supported with fsspec". 

Getting the `NotImplementedError` cought me by suprise.  I was hoping the use of fsspec would guarantee all implementations

TomNicholas:
@okz which implementation were you trying to use?

okz:
@TomNicholas It was azure, `az` or `abfs`.  The exception was so specific I have assumed there is a real requirement for s3, but looking at : 

[...]

PR #48, opened 2024-03-25
Narrative docs

TomNicholas:
Starts working towards #8 by sketching out the structure of a narrative docs page. Makes an effort to keep this separate from the aspirational examples that don't work yet.

Also removes text in README in favour of having it in the actual docs.


TomNicholas:
I've fleshed out half a new page on Usage, written as narrative documentation but covering what actually works already, not what I would like to work. I still have the second half to write but I'll merge this now and do that in a second PR.

PR #52, opened 2024-03-26
open_virtual_dataset with and without indexes

### LINKED ISSUE #18: Inferring concatenation order from coordinate data values
If `xr.combine_by_coords` is used, then indexes must be created for dimension coordinates in order to use them to infer the correct order of concatenation. This requires loading of the data for that coordinate in order to create the index. We can do this, but I think we will then end up with a Dataset which contains some ManifestArrays but also some numpy arrays for the dimension coordinates. This isn't itself a problem - we have that information, we can load it from the files. 

The tricky part is that we presumably don't want to write those actual data values back out during the serialization step, we want to only write the chunk manifest for each dimension coordinate. How can we regain access to this information, or carry it around ready for serialization at the end?

TomNicholas:
Closes #18

This PR ensures that you can either create a virtual dataset with real in-memory pandas indexes using `open_virtual_dataset(..., indexes=None)`, or avoid creating indexes entirely using `open_virtual_dataset(..., indexes={})`. The kwarg signature here was chosen to match what is planned for `xr.open_dataset`, see https://github.com/pydata/xarray/issues/6633.


virtualizarr/xarray.py
```diff
@@ -38,27 +39,38 @@ def open_virtual_dataset(
...
+        ds = xr.open_dataset(filepath)
+        indexes = ds.xindexes
+        ds.close()
+
+    vds = dataset_from_kerchunk_refs(
+        vds_refs,
         drop_variables=drop_variables,
         virtual_array_class=virtual_array_class,
         indexes=indexes,
     )
```
TomNicholas, line 69: We're literally opening the file twice here - once with kerchunk to read all the byte ranges, and then optionally once again to read in the values to use to create defaut pandas indexes with xarray.

Wondering if you have any thoughts on how hard it might be to consolidate these @jhamman ?

PR #55, opened 2024-03-27
Updated NetCDF IO path

norlandrhagen:
The NetCDF3 reader previously imported `SingleHDF5ToZarr`, instead of `NetCDF3ToZarr`.


TomNicholas:
We can worry about proper integration tests later.

PR #56, opened 2024-03-27
API docs

TomNicholas:
Cleans up and exposes a more useful and minimal subset of the API (so defining the public API). 

Also changes some autodoc configuration to get everything to look nice with links and so on, basically just copying the way it was done in [Cubed](https://github.com/cubed-dev/cubed/tree/main/docs).


TomNicholas:
It would be nice to not have all the pydantic-specific methods on the `ChunkManifest` class be included in the API docs, but that can wait.

PR #58, opened 2024-03-27
Refactor test case virtual dataset creation into pytest fixture. 

norlandrhagen:
Moved logic from test_combine_by_coords for creating a virtual dataset into a fixture for re-use.


TomNicholas:
I'm not sure this change makes sense... I can see why it might be nice to factor out the creation of the virtual datasets into a fixture, but the `xr.combine_by_coords` call should definitely still be in the test itself, because if an error is thrown it will likely be from inside that function.

norlandrhagen:
Makes sense! Getting a bit too carried away in fixture land 😅

Issue #65, opened 2024-03-29
Filters should be a list of dictionaries


abarciauskas-bgse:
I believe filters should be an optional list of dictionaries, at least in the case of netcdf4, which is read, [in kerchunk, by the h5py library](https://github.com/fsspec/kerchunk/blob/a0c4f3b828d37f6d07995925b324595af68c4a19/kerchunk/hdf.py#L93). Futher the zarr spec indicates filters should be a [list of json objects](https://zarr.readthedocs.io/en/stable/spec/v2.html)

Without this datatype change, I get pydantic type errors which I first reported in https://github.com/TomNicholas/VirtualiZarr/issues/60.

## Reproducible example

In this example, I created an artificial dataset with filters as well as used the air dataset from the [Usage](https://virtualizarr.readthedocs.io/en/latest/usage.html) docs since I knew that worked. It is interesting how the netcdf4 library appears to read filters from both files and the h5py library only reads filters from the artificially generated dataset. I have not yet tracked down why this is.

```
from netCDF4 import Dataset
import numpy as np
from virtualizarr import open_virtual_dataset
import xarray as xr
import h5py
from netCDF4 import Dataset

# Create some artificial data
data = np.random.rand(100, 100)  # 100x100 array of random numbers

# Create a new NetCDF file
nc_filename = 'artificial_with_filter.nc'
nc_file = Dataset(nc_filename, 'w', format='NETCDF4')

# Define the dimensions of the data
nc_file.createDimension('x', data.shape[0])
nc_file.createDimension('y', data.shape[1])

# Create a variable with zlib compression
data_var = nc_file.createVariable('data', np.float32, ('x', 'y'), zlib=True)

# Assign the data to the variable
data_var[:] = data

# Close the file
nc_file.close()

print(f"NetCDF file '{nc_filename}' created successfully with zlib compression.")

# create an example netCDF4 file from xarray dataset
ds = xr.tutorial.open_dataset('air_temperature')
ds.to_netcdf('air.nc')

files = [('air.nc'), ('artificial_with_filter.nc')]
var_keys = ['air', 'data']
for file in files:
    h5file = h5py.File(file, 'r')
    nc_file = Dataset(file, 'r')
    for group_name in h5file.keys():
        if group_name in var_keys:
            group = h5file[group_name]

            h5filters = group._filters
            print(f"Filters found with hdf5 for {group_name}: {h5filters}")

            var = nc_file.variables[group_name]
[...]

PR #66, opened 2024-03-29
Ab/filters dtype

abarciauskas-bgse:
Addresses #65

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 requires-python = ">=3.9"
 dynamic = ["version"]
 dependencies = [
-    "xarray",
+    "xarray@git+https://github.com/TomNicholas/xarray#egg=concat-no-indexes",
     "kerchunk",
     "pydantic",
     "packaging",
--- a/.gitignore
+++ b/.gitignore
@@ -158,3 +158,4 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+virtualizarr/_version.py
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ classifiers = [
 requires-python = ">=3.9"
 dynamic = ["version"]
 dependencies = [
-    "xarray",
+    "xarray@git+https://github.com/TomNicholas/xarray#egg=concat-no-indexes",
```
abarciauskas-bgse, line 24: This was required to get a few tests to pass. Even with this change I am still getting the error

```
E       NotImplementedError: ManifestArrays can't be converted into numpy arrays or pandas Index objects
```

from virtualizarr/tests/test_xarray.py:175 TestConcat.test_concat_dim_coords_along_existing_dim

TomNicholas: The CI in this PR is not using the correct branch of xarray - the CI show the error is raised on this line

```python
result = type(datasets[0])(result_vars, attrs=result_attrs)
```

but in https://github.com/pydata/xarray/pull/8872 and my `concat-no-indexes` branch [this line is different](https://github.com/TomNicholas/xarray/blob/cb70bec455eae709137136ba8ae9f2044e699cdd/xarray/core/concat.py#L669)

the line should read 

```python
result = type(datasets[0])(result_vars, coords=coords, attrs=result_attrs)
```

so something is up with the pip-installing of that specific branch.

(I'm trying to get https://github.com/pydata/xarray/pull/8872 merged so that we can just use xarray `main`)

TomNicholas:
Thanks for the issue and for the fix @abarciauskas-bgse ! I think this PR is mergeable, but the tests are failing because the wrong version of xarray is being installed by the CI...

pyproject.toml
TomNicholas: Also note that this test passes on the CI on `main` of this package

https://github.com/TomNicholas/VirtualiZarr/actions/runs/8457581269/job/23169875291

abarciauskas-bgse: ok trying with the default xarray branch, like in the main branch of this package.

abarciauskas-bgse:
Tests are now passing @TomNicholas, let me know if you have other comments 👍🏽

TomNicholas:
This looks great, thank you!

PR #67, opened 2024-03-29
Adding `reader_options` kwargs to open_virtual_dataset.

norlandrhagen:
In a step to start reading remote files https://github.com/TomNicholas/VirtualiZarr/issues/61, this PR adds in `reader_options` to the open_virtual_dataset function. 

These `reader_options` are passed into each Kerchunk file reader (SingleHdf5ToZarr, NetCDF3ToZarr, etc..) in `read_kerchunk_references_from_file`. Once `open_virtual_dataset` is replaced with the Xarray backend, we could pass them through: `ds = xr.open_dataset(fp, engine='virtualizarr', backend_kwargs={'reader_options': {'storage_options': {'anon': True}}})`.

This approach relies on the user knowing what options are available in each Kerchunk file reader. 

This example should work off of this PR pointing to a public s3 bucket.

```python

path = 's3://carbonplan-share/virtualizarr/local.nc'

vds = open_virtual_dataset(path)


```
edit: Tests are passing. Index generation and filetype guessing are now working.


TomNicholas:
@norlandrhagen I've merged `main` in here because @jbusecke is using this branch for testing with CMIP6 data in #93. It would be great to get this PR into `main`!

virtualizarr/kerchunk.py
```diff
@@ -57,24 +61,24 @@ def read_kerchunk_references_from_file(
...
         # TODO Grib files should be handled as a DataTree object
         # see https://github.com/TomNicholas/VirtualiZarr/issues/11
         raise NotImplementedError(f"Unsupported file type: {filetype}")
     elif filetype.name.lower() == "tiff":
         from kerchunk.tiff import tiff_to_zarr
 
-        refs = tiff_to_zarr(filepath, inline_threshold=0)
-    elif filetype.name.lower() == "fits":
+        refs = tiff_to_zarr(filepath, inline_threshold=0, **reader_options)
+    elif filetype.lower() == "fits":
```
TomNicholas, line 78:
```suggestion
    elif filetype.name.lower() == "fits":
```

virtualizarr/utils.py
```diff
@@ -0,0 +1,46 @@
+from typing import Optional, Union
+
+# TODO: importing fsspec and s3fs to get typing. Is there a better way incase these are optional deps?
+from s3fs.core import S3File
+from fsspec.implementations.local import LocalFileOpener
```
TomNicholas, line 5: You can use this try except pattern within `TYPE_CHECKING`

https://github.com/pydata/xarray/blob/aaa778cffb89baaece31882e03a7f4af0adfe798/xarray/core/types.py#L87

virtualizarr/utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _fsspec_openfile_from_filepath(*, filepath: str, reader_options: Optional[dict] = {'storage_options':{'key':'', 'secret':'', 'anon':True}}) -> Union[S3File, LocalFileOpener]:
+    """Utility function to facilitate reading remote file paths using fsspec.
+
+    :param filepath: Input filepath
+    :type filepath: str
+    :param reader_options: Dict containing options to pass to fsspec file reader. Default: {'storage_options':{'key':'', 'secret':'', 'anon':True}}
+    :type reader_options: Optional[dict]
+    :rtype: Union[S3File, LocalFileOpener]
```
TomNicholas, line 15: This is not the same docstring style we're using in the rest of the package!

virtualizarr/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    import fsspec
+    from upath import UPath
+
+    universal_filepath = UPath(filepath)
+    protocol = universal_filepath.protocol
+
+    # why does UPath give an empty string for a local file protocol :(
+    # import pdb; pdb.set_trace()
```
TomNicholas, line 24:
```suggestion
```

virtualizarr/xarray.py
```diff
@@ -27,6 +28,7 @@ def open_virtual_dataset(
     loadable_variables: Optional[Iterable[str]] = None,
     indexes: Optional[Mapping[str, Index]] = None,
     virtual_array_class=ManifestArray,
+    reader_options: Optional[dict] = {'storage_options':{'key':'', 'secret':'', 'anon':True}},
```
TomNicholas, line 31: When you normally point `xr.open_dataset` at an S3 url, you don't need to pass `reader_options` do you? Can we try to follow the signature of `xr.open_dataset` as closely as possible? (Maybe this already is as close as we can get)

virtualizarr/tests/test_xarray.py
```diff
@@ -271,6 +272,19 @@ def test_combine_by_coords(self, netcdf4_files):
         assert combined_vds.xindexes["time"].to_pandas_index().is_monotonic_increasing
 
 
+
+
+
+
```
TomNicholas, line 278: There are lots of linting things in this PR like this that our tools should be correcting, see #96

virtualizarr/tests/test_xarray.py
```diff
@@ -271,6 +272,19 @@ def test_combine_by_coords(self, netcdf4_files):
...
+pytest.importorskip("s3fs")
+@pytest.mark.parametrize("filetype", ['netcdf4', None], ids=["netcdf4 filetype", "None filetype"])
+@pytest.mark.parametrize("indexes", [None, {}], ids=["None index", "empty dict index"])
+def test_anon_read_s3(filetype, indexes):
+    #TODO: Switch away from this s3 url after minIO is implemented.
+    fpath = 's3://carbonplan-share/virtualizarr/local.nc'
+    assert open_virtual_dataset(fpath, filetype=filetype, indexes=indexes)
```
TomNicholas, line 286: Can we assert a little more about the structure here? Maybe at least checking that certain variables / indexes are present? Because currently this would pass even if it returned just a non-empty string (somehow).

TomNicholas (commented):
Thanks @norlandrhagen ! I think this is a nice minimal addition to support reading from s3. I like how the fsspec stuff is generally kept separate too.

virtualizarr/utils.py
norlandrhagen: Good catch! Just to check, are we using the `numpy` style?

TomNicholas: Whatever style this is, which I think is numpy?

https://github.com/TomNicholas/VirtualiZarr/blob/3ba30c7c46c725e0e2c637a173445f7a403b06a7/virtualizarr/xarray.py#L38

virtualizarr/xarray.py
norlandrhagen: I could be wrong, but I thought you had to pass in some sort of fsspec/s3fs mapper. 

For me this fails:
``` import xarray as xr
ds = xr.open_dataset('s3://carbonplan-share/virtualizarr/local.nc')
```

virtualizarr/tests/test_xarray.py
norlandrhagen: Good call!

norlandrhagen:
Looks like the docs + CI builds are failing with:

```
Installing pip dependencies: ...working... Pip subprocess error:
  Running command git clone --filter=blob:none --quiet https://github.com/TomNicholas/xarray.git /tmp/pip-install-rmq_i38f/xarray_61220b7470e0456fa17857a7af71f04b
  WARNING: Did not find branch or tag 'concat-avoid-index-auto-creation', assuming revision or ref.
  Running command git checkout -q concat-avoid-index-auto-creation
  error: pathspec 'concat-avoid-index-auto-creation' did not match any file(s) known to git
  error: subprocess-exited-with-error
```

@TomNicholas should we pin another version of Xarray?

TomNicholas:
> should we pin another version of Xarray?

https://github.com/pydata/xarray/pull/8872 was merged yesterday so now we should be able to release xarray, remove the xarray pin in virtualizarr, then release the first version of virtualizarr!

TomNicholas:
This seems very close now @norlandrhagen ?

norlandrhagen:
CI is passing now @TomNicholas!

TomNicholas:
Thanks @norlandrhagen ! One final request: Can we add a quick line to the docs? Something like

> To open remote files as virtual datasets, pass the `reader_kwargs` options, e.g.
> 
> `vds = open_virtual_dataset("s3://fake-bucket/file.nc", reader_kwargs={whatever would be needed})`

This would go on the usage page, either as a quick entry underneath the first `Opening files as virtual datasets` heading, or under a new heading at the bottom.

docs/usage.md
```diff
@@ -61,6 +62,17 @@ Attributes:
...
 
+### Opening remote files
+
+To open remote files as virtual datasets pass the `reader_options` options, e.g.
+
+```python
+
+aws_credentials = {"key": "", "secret": ""}
+vds = open_virtual_dataset("s3://fake-bucket/file.nc", reader_options={'storage_options':aws_credentials})
+
```
TomNicholas, line 73:
```suggestion
```python
aws_credentials = {"key": ..., "secret": ...}
vds = open_virtual_dataset("s3://some-bucket/file.nc", reader_options={'storage_options': aws_credentials})
```

TomNicholas:
Thank you so much @norlandrhagen ! Will merge this now

jbusecke:
Fantastic. Thanks so much. Ill refactor my code in the coming days.

Issue #62, opened 2024-03-28
"Inlining" data when writing references to disk


TomNicholas:
Sometimes we might prefer to write actual data values out into the on-disk kerchunk references file (/ zarr store), because it's more efficient than storing byte ranges to point to very small amounts of data. e.g.

> you wouldn't want to read ~90k time steps from 90k files to construct a 90K long time coordinate

(_Originally posted by @dcherian in https://github.com/TomNicholas/VirtualiZarr/issues/18#issuecomment-2025438773_)
        
Kerchunk calls this "inlining". 

To implement this we need to actually read those data values into memory in the first place. Once #18 is solved we would already be doing that for 1D coordinate indexes (which are the data we most likely want to inline anyway), so the choice of whether or not to "inline" those values could be deferred until the `.to_kerchunk` write step. (But you wouldn't be able to inline if you never created the indexes.)

TomNicholas:
To support this in zarr would it be enough to write a mixed store, i.e. one where some arrays are backed by `manifest.json` files and some by normal zarr chunks? Where the latter is effectively the "inlined" data? cc @jhamman

TomNicholas:
Case (2) would require adding inlining into the manifest spec right? (xref https://github.com/zarr-developers/zarr-specs) Whereas case (1) is just a mixture of "normal" zarr arrays and arrays where every chunk is represented in the manifest but nothing is inlined into the manifest itself.

TomNicholas:
What's the kerchunk equivalent of case (2)? Is it written into the kerchunk references the same as case (1) is?

jhamman:
In the kerchunk story, (1) and (2) are equivalent.

The special thing about (1) is that you don't have to inline the data in the case of a zarr manifest. You could, for specific variables, choose to rewrite the array, perhaps with new chunking/encoding/etc. Why would you want to do this?

Imagine a situation where you concatenate a dataset with a 10k of time variables of size `(1, )`. This is very easy to inline with an effective chunksize of `(1, )` but you may want to just rewrite the concatenated time variable as a real zarr array with shape `(10000, )` and a chunksize of `(1000, )`.

TomNicholas:
So in the kerchunk story, (1) is kind of a poor mans Zarr array? Because you've written in actual data to the `.json` file for every chunk in an array. So in a more zarr-native way of doing things, we might as well just write a real zarr array for that variable instead.

[...]

PR #69, opened 2024-04-01
Load selected variables instead of making them virtual

TomNicholas:
Towards implementing case (1) of #62, at least the opening step. We would still need to be able to save these loaded variables out (to kerchunk and zarr ideally). 

~~Mostly works but there is a subtlety with the difference between a `Variable` and an `IndexVariable` causing the test to fail.~~

### DIFF
--- a/virtualizarr/xarray.py
+++ b/virtualizarr/xarray.py
@@ -1,4 +1,4 @@
-from typing import List, Literal, Mapping, Optional, Union, overload, MutableMapping
+from typing import List, Literal, Mapping, Optional, Union, overload, MutableMapping, Iterable
 
 import ujson  # type: ignore
 import xarray as xr
@@ -21,8 +21,8 @@ class ManifestBackendArray(ManifestArray, BackendArray):
 def open_virtual_dataset(
     filepath: str,
     filetype: Optional[str] = None,
-    drop_variables: Optional[List[str]] = None,
-    load_variables: Optional[List[str]] = None,
+    drop_variables: Optional[Iterable[str]] = None,
+    loadable_variables: Optional[Iterable[str]] = None,
     indexes: Optional[Mapping[str, Index]] = None,
     virtual_array_class=ManifestArray,
 ) -> xr.Dataset:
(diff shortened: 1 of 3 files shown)


virtualizarr/xarray.py
```diff
@@ -21,6 +21,7 @@ def open_virtual_dataset(
     filepath: str,
     filetype: Optional[str] = None,
     drop_variables: Optional[List[str]] = None,
+    load_variables: Optional[List[str]] = None,
```
TomNicholas, line 24: Not sure "load" is the best name for this kwarg, because these variables will still be lazy. But what else is the opposite of "virtual"? "Real" also has incorrect connotations (and "inline" makes no sense here).

virtualizarr/xarray.py
```diff
@@ -92,31 +128,52 @@ def dataset_from_kerchunk_refs(
...
+    vars = {var_name: variable_from_kerchunk_refs(
             refs, var_name, virtual_array_class
-        )
+        ) for var_name in var_names_to_keep}
+
+    return vars
+
+
+
+def dataset_from_kerchunk_refs(
```
TomNicholas, line 147: We're now only using this function in internal tests

virtualizarr/xarray.py
```diff
@@ -41,49 +42,84 @@ def open_virtual_dataset(
...
     )
 
-    if indexes is None:
-        # add default indexes by reading data from file
+    if indexes is None or load_variables != []:
         # TODO we are reading a bunch of stuff we know we won't need here, e.g. all of the data variables...
         # TODO it would also be nice if we could somehow consolidate this with the reading of the kerchunk references
-        ds = xr.open_dataset(filepath)
+        # TODO really we probably want a dedicated xarray backend that iterates over all variables only once
+        ds = xr.open_dataset(filepath, drop_variables=drop_variables)
```
TomNicholas, line 78: At some point it might be nice to have a dedicated function / xarray backend that only opens the legacy file once, instead of opening it once with kerchunk and once with with xarray.

virtualizarr/xarray.py
TomNicholas: Renamed to "loadable", which is marginally better because it doesn't imply that they have already been loaded.

PR #79, opened 2024-04-10
Convert user defined filetype to FileType

norlandrhagen:
Minor PR to catch case where user defined filetype is not a `FileType`


TomNicholas:
Thanks! Should this have a test?

Issue #68, opened 2024-04-01
How to handle encoding


TomNicholas:
Encoding is the part of the virtualization problem that I have the least clear idea of how to handle using this library.

The [discussion in #42](https://github.com/TomNicholas/VirtualiZarr/pull/42#discussion_r1536040324) shows that currently to roundtrip an xarray dataset to disk as kerchunk then open it with fsspec requires ignoring encoding of the time variable. (In general time variables are going to be the most common offenders for encoding mistakes, as different files in the same dataset can easily end up with different encoding options for times.)

Part of the problem is that encoding is kind of a second-class citizen in xarray's data model. It happens magically behind the scenes in the backend file opener system (see https://github.com/pydata/xarray/issues/8548 for a great overview), and then is attached at the `Variable` level as `var.encoding`, but there aren't general rules for propagating it through operations (see https://github.com/pydata/xarray/issues/1614 and https://github.com/pydata/xarray/issues/6323). `xr.concat` will just keep the encoding of the first object passed, which for us could lead to an incorrect result.

There is also some ambiguity around whether certain types of encoding should be handled by xarray or zarr. For example `scale` and `offset` exist in the CF conventions and so are normally handled by xarray, but could also be handled by a Zarr codec.

One practical workaround for a lot of cases would be to load the encoded variables into memory (i.e. "inline" them) - see #62. Then effectively xarray has decoded them, you do the concatenation in-memory, and re-encode them when you save back out to the new zarr store. For dimension coordinates you often might want to do this anyway.

TomNicholas:
>One practical workaround for a lot of cases would be to load the encoded variables into memory (i.e. "inline" them)

See also https://github.com/TomNicholas/VirtualiZarr/pull/73#issuecomment-2042915164, which implies that kerchunk / fsspec doesn't actually do any decoding of times at all.

rabernat:
I think a lot of encoding challenges can be solved by pushing encoding down the stack, in to Zarr and out of Xarray. `scale_factor` / `add_offset` is a perfect minimal example of this. This can be implemented either as a user attribute, which trigger's Xarray encoding / decoding, or as a Zarr codec, in which case Zarr handles the encoding / decoding.

There are two obvious advantages to doing this at the Zarr level:
- You can easily read and write data directly to the Zarr array, bypassing Xarray, if you wish. Zarr users see the same data as Xarray users.
[...]

PR #87, opened 2024-04-22
Non-kerchunk backend for HDF5/netcdf4 files.

sharkinsspatial:
This is a rudimentary initial implementation for https://github.com/TomNicholas/VirtualiZarr/issues/78.  The core code is ported directly from kerchunk's [hdf backend](https://github.com/fsspec/kerchunk/blob/main/kerchunk/hdf.py).  I have not ported the bulk of the kerchunk backend's specialized encoding translation logic but I'll try to do so incrementally so that we can build complete test coverage for the many edge cases it currently covers.


virtualizarr/readers/hdf.py
```diff
@@ -0,0 +1,206 @@
+from typing import List, Mapping, Optional
+
+import fsspec
```
TomNicholas, line 3: Does one need fsspec if reading a local file? Is there any other way to read from S3 without fsspec at all?

virtualizarr/readers/hdf.py
```diff
@@ -0,0 +1,206 @@
...
+    dims = _dataset_dims(dataset)
+    attrs = _extract_attrs(dataset)
+    variable = xr.Variable(data=marray, dims=dims, attrs=attrs)
+    return variable
+
+
+def virtual_vars_from_hdf(
+    path: str,
+    drop_variables: Optional[List[str]] = None,
+) -> Mapping[str, xr.Variable]:
```
TomNicholas, line 191: I like this an a way to interface with the code in `open_virtual_dataset`

TomNicholas (commented):
This is looking great so far @sharkinsspatial !

> kerchunk backend's specialized encoding translation logic

This part I would really like to either factor out, or at a least really understand what it's doing. See #68

virtualizarr/readers/hdf.py
rabernat: Not with a filesystem-like API. You would have to use boto3 or aiobotocore directly.

This is one of the great virtues of fsspec and is not to be under-valued.

rabernat:
This looks cool @sharkinsspatial!

My opinion is that it doesn't make sense to just forklift the kerchunk code into virtualizarr. What I would love to see is an extremely tight, strictly typed, unit-tested total refactor of the parsing logic. I think you're headed down the right path, but I encourage you to push as far as you can in that direction.

sharkinsspatial:
@rabernat Fully agree with your take above 👆 👍 .  I'm trying to work through this incrementally whenever I can find some spare time.  In the spirit of thorough test coverage 🎊 looking through your issue https://github.com/pydata/xarray/issues/7388 and the corresponding PR I'm not sure what the proper incantation of variable encoding configuration is to use `blosc` with the `netcdf4` engine?  Do you have an example of this that you can provide?

sharkinsspatial:
LZF compression support currently blocked by https://github.com/cgohlke/imagecodecs/issues/103

virtualizarr/xarray.py
```diff
@@ -96,26 +101,41 @@ def open_virtual_dataset(
...
-            reader_options=reader_options,
-        )
-        virtual_vars = virtual_vars_from_kerchunk_refs(
-            vds_refs,
-            drop_variables=drop_variables + loadable_variables,
-            virtual_array_class=virtual_array_class,
-        )
-        ds_attrs = kerchunk.fully_decode_arr_refs(vds_refs["refs"]).get(".zattrs", {})
+        if filetype is None:
+            filetype = _automatically_determine_filetype(filepath=filepath)
```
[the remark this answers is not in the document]
sharkinsspatial: 👍 Thanks @ghidalgo3.  Should be fixed by https://github.com/zarr-developers/VirtualiZarr/pull/87/commits/64f446c8d452291548bba2c73a104bf068dc2d7e

virtualizarr/readers/hdf_filters.py
```diff
@@ -0,0 +1,126 @@
...
+            offset = attributes["add_offset"][0]
+        except IndexError:
+            offset = attributes["add_offset"]
+        mapping["add_offset"] = offset
+    else:
+        mapping["add_offset"] = 0
+    if mapping["scale_factor"] != 1 or mapping["add_offset"] != 0:
+        float_dtype = _choose_float_dtype(dtype=dataset.dtype, mapping=mapping)
+        target_dtype = np.dtype(float_dtype)
+        codec = FixedScaleOffset(
```
TomNicholas, line 109: Are you able to make [this test parametrization](https://github.com/zarr-developers/VirtualiZarr/blob/08ca0cd557d108c1ac500b88fd023142fb3294d7/virtualizarr/tests/test_integration.py#L17) pass with this PR? It's currently xfailed because `open_virtual_dataset` doesn't know how to handle scale factor encoding.

sharkinsspatial: I might be misunderstanding, but none of the hdf reader code will be called for `loadable_variables` and this [block](https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/kerchunk.py#L271) would only be entered for a loaded variable.  Is that correct?

virtualizarr/readers/hdf.py
```diff
@@ -0,0 +1,243 @@
...
+            chunk_manifest = ChunkManifest(entries=chunk_entries)
+            return chunk_manifest
+    else:
+        num_chunks = dsid.get_num_chunks()
+        if num_chunks == 0:
+            raise ValueError("The dataset is chunked but contains no chunks")
+
+        shape = tuple(math.ceil(a / b) for a, b in zip(dataset.shape, dataset.chunks))
+        paths = np.empty(shape, dtype=np.dtypes.StringDType)  # type: ignore
+        offsets = np.empty(shape, dtype=np.int32)
```
ghidalgo3, line 55: After https://github.com/zarr-developers/VirtualiZarr/pull/177 , these arrays will need to be `uint64` instead of `int32`.

virtualizarr/readers/hdf.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def _dataset_to_variable(path: str, dataset: h5py.Dataset) -> Optional[xr.Variable]:
+    # This chunk determination logic mirrors zarr-python's create
+    # https://github.com/zarr-developers/zarr-python/blob/main/zarr/creation.py#L62-L66
+
+    manifest = _dataset_chunk_manifest(path, dataset)
+    if manifest:
+        chunks = dataset.chunks if dataset.chunks else dataset.shape
+        codecs = codecs_from_dataset(dataset)
```
ghidalgo3, line 175: - Given the ZarrV3 spec on codecs being non-empty: https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#id12
- this comment: https://github.com/zarr-developers/zarr-python/pull/1944#issuecomment-2150056083 mapping `filters` and `compressor` to v3 concepts
- And empirically we observed that on GOES data this builds of a list of [zlib](https://numcodecs.readthedocs.io/en/stable/zlib.html) and [FixedScaleOffset](https://numcodecs.readthedocs.io/en/stable/fixedscaleoffset.html)

Leaving `compressor=None` causes ambiguity for roundtripping v3 metadata (`ZArray -> disk -> ZArray`) because we can't determine if it's a list of 2 filters or a list of one filter and one compressor. `zlib` is a compression codec and `FixedScaleOffset` is not, but should they both be treated as filters?

sharkinsspatial: @ghidalgo3 My rationale for describing the full codec chain in the `filters` property was the fact that internally HDF5 does not distinguish compressors and filters, the entire encoding chain is represented as `filters`.  Since we don't need to worry about v2 interoperability, I think we can just focus with aligning with v3's api (which still seem to be in a state of flux).  I think I prefer the approach proposed in https://github.com/zarr-developers/zarr-python/pull/1944#issuecomment-2148231227 but I don't know where that leaves me in the interim until a final decision gets made on the v3 api path 🤔.  For v3 compatibility we'll also need to track https://github.com/zarr-developers/numcodecs/pull/524 so we use `numcodecs` which are compatible with the new v3 codec specification.  TLDR I think we might be in flux for some time while upstream v3 decisions get made.

sharkinsspatial:
@ghidalgo3 I also want to address your question from your PR https://github.com/zarr-developers/VirtualiZarr/pull/193
> Also, what happens if a source file uses a codec that is not one of the specified codecs of ZarrV3? Does that mean the file cannot be represented in ZarrV3? Seems rather onerous.
IIUC different v3 implementations will support a codec registry https://github.com/zarr-developers/zarr-python/pull/1588 to make the codec support fully extensible.  Codec discovery and registration has always been a thorny problem (this is a big issue in the HDF space) but I'm hopeful that this approach will be flexible.

sharkinsspatial:
@TomAugspurger I'm trying to merge `main` into my branch so I can also investigate removing the Pydantic dependency in this branch and have my pre-commit working correctly again.  But I seem to be hitting an issue with some change in `zarr.py` from https://github.com/zarr-developers/VirtualiZarr/pull/213.  All of the roundtrip tests are failing on `to_kerchunk_json` with `TypeError: array([xxx]) is not JSON serializable` but I can't seem to find which of your changes is causing this 🤔 and I'm a bit stumped.  I'm packing and moving countries this week 🇲🇽 -> 🇨🇦 so I'm probably missing something obvious 😄 .  Any thoughts?

TomAugspurger:
@sharkinsspatial this is a behavior change in `ZArray.dict` during the move away from pydantic.

Something like this seems to fix the failing tests

```diff
diff --git a/virtualizarr/zarr.py b/virtualizarr/zarr.py
index 824892c..87bb453 100644
--- a/virtualizarr/zarr.py
+++ b/virtualizarr/zarr.py
@@ -106,8 +106,15 @@ class ZArray:
 
     def to_kerchunk_json(self) -> str:
         zarray_dict = self.dict()
-        if zarray_dict["fill_value"] is np.nan:
+
+        fill_value = zarray_dict["fill_value"]
+
+        if fill_value is np.nan:
             zarray_dict["fill_value"] = None
+
+        elif isinstance(fill_value, (np.number, np.ndarray)):
+            zarray_dict["fill_value"] = fill_value.item()
+
         return ujson.dumps(zarray_dict)
 
     # ZArray.dict seems to shadow "dict", so we need the type ignore in
```

I'm not sure what behavior we want `dict()` to have. Whether it should faithfully return the value set, or whether it should attempt to cast to something else.

TomNicholas:
The ZArray class was supposed to be a way to standardize the metadata, allowing the rest of the package to not worry about any differences that kerchunk throws at us.

The .dict() method we just got for free via inheriting from the pydantic base model.

I think we should migrate the interface of ZArray towards providing a unified representation of the metadata, and also try to move its API closer to that of zarr-python's ZMetaData class because really we want to be using that instead.

TomNicholas:
@sharkinsspatial just wondering what the status of this PR is? Right now would be a very useful point in time to be able to generate references without kerchunk - with this PR in we could plausibly consider making kerchunk an optional dependency.

sharkinsspatial:
@TomNicholas Apologies, I definitely let this languish as I've had to prioritize some other work recently.  But the good news is that @abarciauskas-bgse @maxrjones and @norlandrhagen will all be trapped in a lodge together this week for SatCamp 😄 so I'm guessing with this critical mass, we can hack together the remaining bits quickly.

I believe I left things with https://github.com/zarr-developers/VirtualiZarr/pull/213 causing some breaking tests on my branch.  I'll try to get tests passing against the current `main`.  

There are still several outstanding items to achieve feature parity with `kerchunk`.  I had reached out to @ajelenak via Slack earlier this summer to see if folks from the HDF group could review some of this work and also to get more insight into some of the Kerchunk code they contributed but haven't heard back from them.  Specifically, I'm a bit unclear about everything that is covered in https://github.com/fsspec/kerchunk/blob/main/kerchunk/hdf.py#L272.  But I'm guessing that we can try and wade through it this week as a group and try to re-implement it in Virtualizarr.

I think ideally we can get this to a mergeable state and then fix any missing functionality or encoding logic as users encounter and report it.

TomNicholas:
> let this languish as I've had to prioritize some other work recently.

story of my life

> trapped in a lodge together

that's fun!

> I think ideally we can get this to a mergeable state and then fix any missing functionality or encoding logic as users encounter and report it.

If the cases where this fails are well-understood then we should be able to merge it sooner rather than later?

sharkinsspatial:
@TomNicholas Apologies, our internet was not optimal at SatCamp (and we were a bit waylaid by the mountains 😄 ).  I'm working today on getting this aligned with `main` and @abarciauskas-bgse @norlandrhagen and @maxrjones and I have a pairing session planned to review any gaps with the existing kerchunk implementation tomorrow afternoon.  If all looks good I should be ready to merge on Friday.  Apologies for the long delay.

virtualizarr/readers/hdf.py
```diff
@@ -0,0 +1,247 @@
...
+        variable = xr.Variable(data=marray, dims=dims, attrs=attrs)
+    else:
+        variable = None
+    return variable
+
+
+def virtual_vars_from_hdf(
+    path: str,
+    drop_variables: Optional[List[str]] = None,
+    reader_options: Optional[dict] = {
```
norlandrhagen, line 213: The default `reader_options` were updated a bit ago to:
` reader_options: Optional[dict[str, Any]] = None,`

sharkinsspatial:
@TomNicholas I'll try to get #261 incorporated in this branch this week.  Before tackling this, I'm still dealing with [this failing test](https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/tests/test_integration.py#L181).   I'm still a bit unclear on the mechanics here but I think I may understand.   I was previously ignoring empty HDF datasets ([which is the same behavior as kerchunk's reader](https://github.com/fsspec/kerchunk/blob/main/kerchunk/hdf.py#L613-L628)) but to pass the tests introduced in #205 I introduced [empty variable support](https://github.com/zarr-developers/VirtualiZarr/pull/87/commits/9c86e0d2c0f8135b0a53cbf5313bfa11cc2a222e).  But as a side effect, this causes a roundtripping failure since these empty variables now get treated as coordinates.

This is related to @keewis's recent #260.  So I think I have 2 questions
1. Is [this](https://github.com/zarr-developers/VirtualiZarr/blob/hdf5_reader/virtualizarr/readers/hdf.py#L206) the correct way to represent a variable for an empty HDF5 dataset or should I consider some other sort of "empty" `ManifestArray` for the `data`?
2. As I migrate this to align with #261 what approach should I use to have these variables not be treated as coordinates so that we can have successful roundtripping?

TomNicholas:
> Is [this](https://github.com/zarr-developers/VirtualiZarr/blob/hdf5_reader/virtualizarr/readers/hdf.py#L206) the correct way to represent a variable for an empty HDF5 dataset

I actually don't know, but I would find out by creating a HDF file with an empty variable, opening it with xarray, and seeing if that result is represented the same way as what you have done here.

However if the definition of empty variable should just mean "a zarr array with no chunks, just the default `fill_value`" then I would say you should use the empty manifest approach that @keewis used in #260 instead though.

Ultimately what matters is that opening HDF via xarray and opening virtual zarr that points to HDF via xarray give the same result.

---

> As I migrate this to align with https://github.com/zarr-developers/VirtualiZarr/pull/261 what approach should I use to have these variables not be treated as coordinates so that we can have successful roundtripping?

VirtualiZarr has full control over which variables we choose to make coordinates. I've tried to isolate that logic by in https://github.com/zarr-developers/VirtualiZarr/blob/4b7612e5649058475223cdfbb1bf50910b70ff5b/virtualizarr/readers/common.py#L128, where `coord_names` are the names of variables explicitly labeled as `COORDINATES` in the files' variable-level metadata. Does that help?

> these empty variables now get treated as coordinates.

You mean the fact they are not `ManifestArrays` means that if they are 1D they get treated as coordinates?

FYI that re-implementation of determining coordinates is the reason there is a bug that causes an inconsistency with xarray's behaviour, see #224.

sharkinsspatial:
@TomNicholas We're still investigating some failing test associated with inconsistencies around coordinate variable round tripping  and a compatibility issue with nodata and CF time encoding.  But given that this backend is still considered experimental and requires an explicit opt-in I think this PR is ok to review and merge.  These are the test which are currently xfailed and need further investigation

https://github.com/zarr-developers/VirtualiZarr/blob/hdf5_reader/virtualizarr/tests/test_integration.py#L199
https://github.com/zarr-developers/VirtualiZarr/blob/hdf5_reader/virtualizarr/tests/test_readers/test_hdf_integration.py#L14

This test fails with both the kerchunk hdf reader and the experimental reader.
https://github.com/zarr-developers/VirtualiZarr/blob/hdf5_reader/virtualizarr/tests/test_readers/test_hdf_integration.py#L30

pyproject.toml
```diff
@@ -25,7 +25,11 @@ dependencies = [
     "numpy>=2.0.0",
     "packaging",
     "universal-pathlib",
+    "h5py",
+    "hdf5plugin",
     "numcodecs",
+    "imagecodecs",
+    "imagecodecs-numcodecs==2024.6.1",
```
TomNicholas, line 32: I think these should be optional no? (`numcodecs` is still required). That also means they can be removed from the `min-deps.yml` in CI.

virtualizarr/backend.py
```diff
@@ -27,9 +28,9 @@
     "zarr_v3": ZarrV3VirtualBackend,
     "dmrpp": DMRPPVirtualBackend,
     # all the below call one of the kerchunk backends internally (https://fsspec.github.io/kerchunk/reference.html#file-format-backends)
-    "netcdf3": NetCDF3VirtualBackend,
     "hdf5": HDF5VirtualBackend,
     "netcdf4": HDF5VirtualBackend,  # note this is the same as for hdf5
+    "netcdf3": NetCDF3VirtualBackend,
```
TomNicholas, line 33: Shouldn't your reader be added here too? Sounds like it could go under `hdf` rather than `hdf5`? Or maybe that's too subtle...

TomNicholas (commented):
Amazing! I'll take a more detailed look later / tomorrow / next week 😅

virtualizarr/backend.py
sharkinsspatial: Since we discussed treating this new reader as experimental while we do some further battle testing, I'm forcing users to explicitly pass the new backend class `HDFVirtualBackend` to the optional `backend` argument for `open_virtual_dataset` to distinguish it from using the kerchunk reader with filetype auto-detection.

TomNicholas (commented):
This looks great @sharkinsspatial ! Thank you for your patience. My comments are just nits.

virtualizarr/tests/test_readers/test_hdf_integration.py
```diff
@@ -0,0 +1,54 @@
...
+            loadable_variables=["time"],
+            cftime_variables=["time"],
+            backend=hdf_backend,
+        )
+        kerchunk_file = f"{tmpdir}/kerchunk.json"
+        vds.virtualize.to_kerchunk(kerchunk_file, format="json")
+        roundtrip = xr.open_dataset(kerchunk_file, engine="kerchunk", decode_times=True)
+        xrt.assert_allclose(ds, roundtrip)
+
+    @pytest.mark.xfail(reason="Coordinate issue affecting kerchunk and HDF reader.")
```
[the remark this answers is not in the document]
sharkinsspatial: I've included the issue for the xfailed test for CF time.  I haven't created an issue for the coordinate dim xfail yet since this is actually specific to this branch.  This is related to https://github.com/zarr-developers/VirtualiZarr/pull/260 and the way kerchunk serializes 0d hdf datasets though I don't fully understand the mechanics.  Once this branch is merged I can investigate a bit more and file a complete issue.

sharkinsspatial:
@TomNicholas This should be ready for final review now.

virtualizarr/tests/test_readers/conftest.py
```diff
@@ -0,0 +1,324 @@
...
+import xarray as xr
+from packaging.version import Version
+from xarray.tests.test_dataset import create_test_data
+from xarray.util.print_versions import netcdf_and_hdf5_versions
+
+try:
+    import hdf5plugin  # type: ignore
+except ModuleNotFoundError:
+    hdf5plugin = None  # type: ignore
+    warnings.warn("hdf5plugin is required for HDF reader")
```
TomNicholas, line 15: Why wouldn't this either error or just not be a problem?

TomNicholas (approved):
Looks good! Only one question about warnings in the test suite when dependencies aren't installed.

Also let's add a release note saying this is available for experimental use.

virtualizarr/tests/test_readers/conftest.py
TomNicholas: I added a release note in #307, but I'm still curious about this @sharkinsspatial

PR #91, opened 2024-04-24
Update installation instructions

jbusecke:
The file `ci/environment.yml` does not currently exist anymore and AFAICT this is now handled by the pip install.


norlandrhagen:
Nice find @jbusecke!

PR #98, opened 2024-05-06
Bump Ruff version and add formatting

### LINKED ISSUE #96: Add isort / black to pre-commit?
I noticed that although we run ruff on pre-commit, a lot of the time it doesn't tidy up things that it could do, e.g. untidily-placed `import` statements. We probably need to add more pre-commit tools / configure ruff.

jbusecke:
- [x] Closes #96

I added some config options to `pyproject.toml` to make the linting more closely resemble black. Happy to change/revert those.


TomNicholas (approved):
This looks great! Thanks @jbusecke !

virtualizarr/xarray.py
```diff
@@ -1,5 +1,14 @@
...
+from typing import (
+    List,
+    Literal,
+    Mapping,
+    Optional,
+    Union,
+    overload,
+    MutableMapping,
+    Iterable,
+)
```
TomNicholas, line 11: I think we're still not using `isort`, otherwise these would be in alphabetical order.

jbusecke: On it!

jbusecke: https://github.com/TomNicholas/VirtualiZarr/pull/98/commits/2c72571e5968a8390c40ed7fe4fd92a5ba27215e looks promising. Is this what you would like? There are also options to do this case sensitive?

jbusecke: Just if anyone is curious. All I did was add the "I" class as active to the linter: https://github.com/TomNicholas/VirtualiZarr/pull/98/commits/8c128b57cdb67ae4447164b7c595696f2f3db8fb

TomNicholas: Cool! Thanks @jbusecke ❤️

Issue #104, opened 2024-05-09
Performance roadmap


TomNicholas:
We want to be able to create, combine, serialize, and use manifests that point to very large numbers of files. The largest Zarr stores we already see have `O(1e6)` chunks per array, and 10's or 100's of arrays (e.g. the National Water Model dataset).

To be able to virtualize this many references at once, there are multiple places that could become performance bottlenecks.

1. **Reference generation**
   `open_virtual_dataset` should create `ManifestArray` objects from URLs with a minimum memory overhead. It is unlikely that the current implementation is as efficient as possible because it first creates a python dict rather than a more efficient in-memory representation.
2. **In-memory representation**
   It's crucial that our in-memory representation of the manifest be as memory-efficient as possible. What's the smallest that the in-memory representation of e.g. a single chunk manifest containing a million references could be? I think it's only ~100MB, if we use a numpy array (or 3 numpy arrays) and avoid object dtypes. We should be able to test this easily ourselves by creating a manifest with some dummy references data. See #33.
3. **Combining**
   If we can fit all the `ManifestArray` objects we need into memory on one worker, this part is easy. Again using numpy arrays is good because `np.concatenate/stack` should be memory-efficient. 

   If the combined manifest is too big to fit into a single worker's memory then we might want to use dask to create a task graph, with the root tasks generating the references, concatenation via a tree of tasks (not technically a reduction because the output is as big as the input), then writing out chunkwise to some form of storage. Overall the result would be similar to how kerchunk uses `auto_dask`.
4. **Serialization**
   Writing out the manifest to JSON as text would create an on-disk representation that is larger than necessary. Writing out to a compressed / chunked format such as parquet could ameliorate this. This could be kerchunk's parquet format or we could use a compressed format in the Zarr manifest storage transformer ZEP (see discussion in https://github.com/zarr-developers/zarr-specs/issues/287)

We should be trying to work out what the performance of each of these steps is, and they are separate so we can look at them individually.

cc @sharkinsspatial @norlandrhagen @keewis

TomNicholas:
To demonstrate my point about (2), it looks like storing a million references in-memory can be done using numpy taking up only 24MB.

```python
In [1]: import numpy as np

# Using numpy 2.0 for the variable-length string dtype
In [2]: np.__version__
Out[2]: '2.0.0rc1'

In [3]: N_ENTRIES = 1000000

In [4]: SHAPE = (100, 100, 100)

[...]

PR #107, opened 2024-05-10
Use 3 numpy arrays for manifest internally

### LINKED ISSUE #33: In-memory representation of chunks: array instead of a dict?

TomNicholas:
Supercedes #39 as a way to close #33, the difference being that this uses 3 separate numpy arrays to store the path strings, byte offsets, and byte range lengths (rather than trying to put them all in one numpy array with a structured dtype). Effectively implements (2) in #104.

Relies on numpy 2.0 (which is currently only available as a release candidate).


martindurant:
Here is a script which generated a 9GB JSON file across many years of NWM data: https://gist.github.com/rsignell-usgs/d386c85e02697c5b89d0211371e8b944 . I'll see if I can find a parquet version, but you should reckon on 10x in on-disk size (parquet or compressed numpy).

Unfortunately, the references have been deleted, because the whole dataset is now also available as zarr. I may have the chance sometime to regenerate them, if it's important.

martindurant:
Also, a super-simple arrow- or awkward-like string representation as contiguous numpy arrays could look something like
```
class String:
    def __init__(self, offsets, data) -> None:
        self.offsets = offsets
        self.data = data

    def __getitem__(self, item):
        if isinstance(item, int):
            return self.data[self.offsets[item]: self.offsets[item + 1]].decode()
        else:
            return String(self.offsets.__getitem__(item), self.data)
```

TomNicholas:
> Here is a script which generated a 9GB JSON file across many years of NWM data: https://gist.github.com/rsignell-usgs/d386c85e02697c5b89d0211371e8b944 . I'll see if I can find a parquet version, but you should reckon on 10x in on-disk size (parquet or compressed numpy).

That's useful context for #104, thanks Martin!

TomNicholas:
I'm finding it very difficult to build an environment to fully test this, but I can at least make one where all the tests which don't involve IO pass. I think that means that once numpy 2.0 is released and all the various libraries support it officially, this PR should be fine.

virtualizarr/tests/test_manifests/test_array.py
```diff
@@ -124,13 +124,43 @@ def test_partly_equals(self): ...
...
+        expanded = np.broadcast_to(marr, shape=(3, 2))
+        assert expanded.shape == (3, 2)
+        assert expanded.chunks == (1, 2)
+        assert expanded.manifest.dict() == {
+            "0.0": {"path": "file.0.0.nc", "offset": 0, "length": 5},
+            "1.0": {"path": "file.0.0.nc", "offset": 0, "length": 5},
+            "2.0": {"path": "file.0.0.nc", "offset": 0, "length": 5},
+        }
+
+    def test_broadcast_new_axis(self):
```
TomNicholas, line 138: Note to self: this is still buggy and needs more rigorous testing to ensure the shapes of arrays / chunks are correct. I should add some hypothesis property-based tests to bulletproof this.

keewis:
looks like you're waiting on `netcdf4` to support `numpy>=2.0`. If you (temporarily) remove `netcdf4` from the environment and instead install `h5netcdf` and `h5py` (the latter from the scientific python nightly repo), you should be able to get CI to run properly.

virtualizarr/manifests/manifest.py
```diff
@@ -56,25 +62,131 @@ class ChunkManifest(BaseModel):
...
-    See the chunk manifest SPEC proposal in https://github.com/zarr-developers/zarr-specs/issues/287 .
+    (See the chunk manifest SPEC proposal in https://github.com/zarr-developers/zarr-specs/issues/287.)
 
     Validation is done when this object is instatiated, and this class is immutable,
-    so it's not possible to have a ChunkManifest object that does not represent a complete valid grid of chunks.
+    so it's not possible to have a ChunkManifest object that does not represent a valid grid of chunks.
     """
 
-    model_config = ConfigDict(frozen=True)
+    _paths: np.ndarray[Any, np.dtype[np.dtypes.StringDType]]
```
TomNicholas, line 73: So originally I was going to wait for numpy2.0 before merging this, but actually I could just merge it before that by using a fixed-length string dtype here, special-casing whether or not you have numpy 1 or 2+ installed. That would prevent using paths longer than a certain amount but if I set the length to some high value all it would be fine for a couple of weeks and just waste a bit of memory...

TomNicholas: See also the test failures in https://github.com/zarr-developers/VirtualiZarr/pull/139, which I would rather fix once on this branch than twice (once in `main` and again on this branch).

TomNicholas (commented):
> h5py (the latter from the scientific python nightly repo)

Can I do this by adding a special `pip install` command to a conda env?

TomNicholas:
Now builds atop #139

keewis:
you can:
```yaml
- pip
- pip:
  - -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
  - --pre
  - h5py
  - numpy
```
(and if you need multiple packages from different sources, put the definition in `requirements.txt` files and include them in the environment file using `-r <requirements.txt>`)

virtualizarr/manifests/manifest.py
TomNicholas: But... turns out in order to get this to pass the test suite I need to use a 128-bit dtype, in order to have enough characters to fit this extremely long temporary directory that pytest creates: `'/private/var/folders/zt/b84f41811p5_vygj1gjp6qx40000gn/T/pytest-of-tom/pytest-73/test_numpy_arrays_to_inlined_k0/air.nc'`.

rabernat: Does numpy really not allow variable-length strings?

What about using a pyarrow string array instead: https://arrow.apache.org/docs/python/generated/pyarrow.string.html

TomNicholas: > Does numpy really not allow variable-length strings?

A variable-length string dtype was one of the [headline features of numpy 2.0](https://numpy.org/devdocs/release/2.0.0-notes.html#highlights), released yesterday. I think the simplest thing to do here now is just to require numpy 2.0 as soon as possible (xarray already supports it, but I don't know about some packages like netcdf-python), and not worry about fixed-length strings at all.

> What about using a pyarrow string array instead:

I looked at this with @norlandrhagen, and it might yet be useful somewhere, but I think the numpy variable-length string dtype is exactly what we need for use inside the `Manifest`.

TomNicholas: https://github.com/zarr-developers/VirtualiZarr/pull/107/commits/8d2dcd5e452d192b1e931f037c624df8df9fabb2 just makes this PR depend on `numpy>=2.0.0`. I seem to now be able to build a local environment with all the dependencies I need and numpy 2 which is great.

martindurant:
Note that en efficient packing of a string array would be all the character data concatenated, and a separate offsets array. This is what arrow or awkward does. Parquet can also store this way, but alternating length-data is standard. Dense packings like that ssusme immutability https://github.com/martindurant/fastparquet/blob/faster/fastparquet/wrappers.py#L51 has the simplest possible version of this (yes, fastparquet is intending to move to pure numpy within months, if I didn't say this already).

TomNicholas:
@martindurant does numpy's new string dtype not do that too?

martindurant:
No, it allocates on the heap and stores pointers, I believe. That is not (necessarily) memory contiguous, but allows for mutability.

virtualizarr/tests/test_manifests/test_array.py
TomNicholas: Adding hypothesis tests turned out to be a big rabbit whole so I just added some parametrized tests instead.

TomNicholas:
This PR now works (passes all tests locally). The failures are the same as on `main`, tracked in #147 .

> Note that en efficient packing of a string array would be all the character data concatenated, and a separate offsets array. This is what arrow or awkward does.

> What about using a pyarrow string array instead:

@rabernat @martindurant a pyarrow string array might be a bit more memory-efficient, but 

a) I can store millions of chunk references using just a few MB with this new numpy dtype, which seems plenty good enough to me

```python
In [1]: from virtualizarr.tests import create_manifestarray

In [2]: marr = create_manifestarray(shape=(100, 100, 100), chunks=(1, 1, 1))

In [3]: marr
Out[3]: ManifestArray<shape=(100, 100, 100), dtype=float32, chunks=(1, 1, 1)>

In [4]: marr.manifest._paths.nbytes / 1e6
Out[4]: 16.0

In [5]: (marr.manifest._paths.nbytes + marr.manifest._offsets.nbytes + marr.manifest._lengths.nbytes) / 1e6
Out[5]: 24.0
```

b) IIUC pyarrow string arrays are not N-dimensional arrays, and half the point of this PR is that my implementation of `concat/stack/broadcasting` for `ManifestArrays` becomes really simple if I can just delegate those operations to a wrapped set of numpy arrays.

virtualizarr/manifests/manifest.py
```diff
@@ -56,25 +62,131 @@ class ChunkManifest(BaseModel):
...
+        paths: np.ndarray[Any, np.dtype[np.dtypes.StringDType]],
+        offsets: np.ndarray[Any, np.dtype[np.int32]],
+        lengths: np.ndarray[Any, np.dtype[np.int32]],
+    ) -> "ChunkManifest":
+        """
+        Create manifest directly from numpy arrays containing the path and byte range information.
+
+        Useful if you want to avoid the memory overhead of creating an intermediate dictionary first,
+        as these 3 arrays are what will be used internally to store the references.
+
```
TomNicholas, line 139: @ayushnag @sharkinsspatial you might want to try building numpy arrays of references and passing them to this new constructor instead to keep memory usage down.

PR #108, opened 2024-05-10
Unpin kerchunk (set floor) and enable Python 3.12

jsignell:
Closes https://github.com/TomNicholas/VirtualiZarr/issues/77

Given that https://github.com/fsspec/kerchunk/issues/445 seems to be resolved on main. This PR unpins the kerchunk version and uses ~the unreleased version~ newly released 0.2.5 instead. ~To me it seems fair that this library would sometimes require unreleased kerchunk and that feels preferable to pinning down to an old release.~

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
         shell: bash -l {0}
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


TomNicholas:
Thanks for spotting that kerchunk has fixed that bug @jsignell !

I'm hesitant to pin this package to kerchunk `main` because I would like the reliability of knowing we are pinning to a released version... If stuff like this can be fixed on kerchunk `main` without notice it could also get broken...

jsignell:
To be fair it was already fixed on main when the original ticket was opened. But yeah I see the hesitation. Let's ask for a kerchunk release :)

jsignell:
https://github.com/fsspec/kerchunk/issues/457

PR #110, opened 2024-05-13
Write to parquet

### LINKED ISSUE #72: Writing to parquet (following kerchunk format)
It would be great to add the ability to write kerchunk references to parquet files, not just to json.

This should be a nice self-contained feature for anyone who is interested in implementing it - it just goes in the xarray accessor. (@norlandrhagen ? 😁 )

For implementation I'm not very familiar with the options. I see there is fastparquet, but if we already have an in-memory complete ManifestArray object (which could well become a numpy structured array in #39), it looks like we could just write to parquet from a pandas dataframe? 

In fact, could we even go from the `xr.Dataset` to the pandas dataframe directly using [`Dataset.to_dataframe()`](https://docs.xarray.dev/en/latest/generated/xarray.Dataset.to_dataframe.html)?? That would be super neat, but I don't understand the [kerchunk parquet format](https://fsspec.github.io/kerchunk/spec.html#parquet-references) well enough yet to know how easily that would work.

jsignell:
Closes #72 

I'm pretty sure this is not what you had in mind, but wanted to put this up as at least an option. Even if we scrap the implementation, the test should stay the same.

I didn't include remote writing since it seemed like that would be something that we'd want to implement for all the file outputs not just for one.

### DIFF
--- a/virtualizarr/tests/test_integration.py
+++ b/virtualizarr/tests/test_integration.py
@@ -1,4 +1,6 @@
+import pytest
 import xarray as xr
+import xarray.testing as xrt
 
 from virtualizarr import open_virtual_dataset
 
@@ -11,3 +13,24 @@ def test_open_scalar_variable(tmpdir):
 
     vds = open_virtual_dataset(f"{tmpdir}/scalar.nc")
     assert vds["a"].shape == ()
+
+
+@pytest.mark.parametrize("format", ["json", "parquet"])
+def test_kerchunk_roundtrip(tmpdir, format):
+    # set up example xarray dataset
+    ds = xr.tutorial.open_dataset("air_temperature", decode_times=False)
+
+    # save it to disk as netCDF (in temporary directory)
+    ds.to_netcdf(f"{tmpdir}/air.nc")
+
+    # use open_virtual_dataset to read it as references
+    vds = open_virtual_dataset(f"{tmpdir}/air.nc")
+
+    # write those references to disk as kerchunk json
+    vds.virtualize.to_kerchunk(f"{tmpdir}/refs.{format}", format=format)
+
+    # read the dataset from disk via the zarr store
+    roundtrip = xr.open_dataset(f"{tmpdir}/refs.{format}", engine="kerchunk")
+
+    # assert equal to original dataset
+    xrt.assert_equal(roundtrip, ds)
--- a/virtualizarr/tests/test_integration.py
+++ b/virtualizarr/tests/test_integration.py
@@ -24,7 +24,7 @@ def test_kerchunk_roundtrip(tmpdir, format):
     ds.to_netcdf(f"{tmpdir}/air.nc")
 
     # use open_virtual_dataset to read it as references
-    vds = open_virtual_dataset(f"{tmpdir}/air.nc")
+    vds = open_virtual_dataset(f"{tmpdir}/air.nc", indexes={})
 
     # write those references to disk as kerchunk json
     vds.virtualize.to_kerchunk(f"{tmpdir}/refs.{format}", format=format)
--- a/virtualizarr/tests/test_kerchunk.py
+++ b/virtualizarr/tests/test_kerchunk.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pandas as pd
 import pytest
 import ujson  # type: ignore
 import xarray as xr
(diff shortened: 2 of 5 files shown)


virtualizarr/tests/test_kerchunk.py
```diff
@@ -127,6 +128,47 @@ def test_accessor_to_kerchunk_json(self, tmp_path):
...
+        df0 = pd.read_parquet(filepath / "a" / "refs.0.parq")
+
+        assert df0.to_dict() == {
+            "offset": {0: 100, 1: 200},
+            "path": {
+                0: "foo.nc",
+                1: "foo.nc",
+            },
+            "size": {0: 100, 1: 100},
+            "raw": {0: None, 1: None},
```
TomNicholas, line 169: What does `"raw"` mean in this specification?

virtualizarr/xarray.py
```diff
@@ -397,6 +412,18 @@ def to_kerchunk(
...
         elif format == "parquet":
-            raise NotImplementedError()
+            from kerchunk.df import refs_to_dataframe
+
+            if isinstance(filepath, Path):
+                url = str(filepath)
+            elif isinstance(filepath, str):
+                url = filepath
+
+            return refs_to_dataframe(
```
TomNicholas, line 422: I don't think `refs_to_dataframe` actually has a return value? Is that return there for type hinting reasons?

Also worth clarifying with a comment that `refs_to_dataframe` doesn't return a dataframe, it really should be called `refs_to_parquet`.

virtualizarr/xarray.py
```diff
@@ -397,6 +412,18 @@ def to_kerchunk(
...
+                refs,
```
TomNicholas, line 423: Where is this `refs` argument coming from? It's not an argument to the method...

TomNicholas (commented):
> I'm pretty sure this is not what you had in mind, but wanted to put this up as at least an option. Even if we scrap the implementation, the test should stay the same.

Agree that's useful already! I think this implementation is useful to start with, and tests ensure we can change it later without breaking behaviour.

I was wondering if there is a more performant way to go from `ManifestArray` to parquet, but that implementation will depend on how the references are stored in-memory, so should wait until #107 is merged. 

> I didn't include remote writing since it seemed like that would be something that we'd want to implement for all the file outputs not just for one.

:+1:

virtualizarr/tests/test_kerchunk.py
jsignell: I think it means "inlined" in the kerchunk library tests this ref: `"a/6": b"data",` results in a non-null "raw" 

https://github.com/fsspec/kerchunk/blob/a5eae4d45f7f601346d0fdc58c0864911330c5d6/kerchunk/tests/test_df.py#L62

virtualizarr/xarray.py
jsignell: yeah you are right. I was thinking that it might sometime return a future or something, but that isn't how it works. I'll remove the return

virtualizarr/xarray.py
jsignell: It is defined in line 402 before the if/else block starts.

TomNicholas: Sorry yes I just saw that - stupid comment sorry 😅

jsignell:
I just stole the easiest test off that PR for roundtripping. See what you think

virtualizarr/tests/test_integration.py
```diff
@@ -11,3 +13,24 @@ def test_open_scalar_variable(tmpdir):
...
+@pytest.mark.parametrize("format", ["json", "parquet"])
+def test_kerchunk_roundtrip(tmpdir, format):
+    # set up example xarray dataset
+    ds = xr.tutorial.open_dataset("air_temperature", decode_times=False)
+
+    # save it to disk as netCDF (in temporary directory)
+    ds.to_netcdf(f"{tmpdir}/air.nc")
+
+    # use open_virtual_dataset to read it as references
+    vds = open_virtual_dataset(f"{tmpdir}/air.nc")
```
TomNicholas, line 27:
```suggestion
    vds = open_virtual_dataset(f"{tmpdir}/air.nc", indexes={})
```
This shouldn't matter but better to be explicit

TomNicholas:
This looks good to me! Let's get it in and iterate if necessary!

Thanks @jsignell !

PR #113, opened 2024-05-14
open_virtual_dataset with dmr++

### LINKED ISSUE #85: Reading from dmrcp index files?
@betolink suggested today the idea of reading from [`dmrcpp`](https://docs.opendap.org/index.php?title=DMR%2B%2B) files, which are apparently a sidecar filetype that is often already included with certain NASA datasets, and basically is a chunk references format.

Could we imagine ingesting these into `VirtualiZarr` as another way to read byte ranges? Or maybe even imagine writing out to this filetype?

ayushnag:
- [x] Closes #85
- [x] Tests added
- [x] Changes are documented in docs/releases.rst
- [x] New functions/methods are listed in api.rst
- [x] Open dataset via `group=` param
- [x] Use [reader options]( https://github.com/TomNicholas/VirtualiZarr/pull/67) for dmr file opening pattern


virtualizarr/xarray.py
```diff
@@ -18,6 +18,7 @@
 from xarray.core.variable import IndexVariable
 
 import virtualizarr.kerchunk as kerchunk
+from virtualizarr.dmrpp import DMRParser
```
TomNicholas, line 21: In #87 Sean is adding a `readers` submodule - let's follow that pattern here and have `readers/dmrpp`.

virtualizarr/xarray.py
```diff
@@ -99,6 +100,10 @@ def open_virtual_dataset(
         return open_virtual_dataset_from_v3_store(
             storepath=filepath, drop_variables=drop_variables, indexes=indexes
         )
+    if filetype == "dmr++":
+        with open(filepath, "rb") as f:
```
TomNicholas, line 104: The importing of `DMRParser` should happen inside the if statement here, so that the `dmr++`-related libraries can be optional dependencies.

virtualizarr/dmrpp.py
```diff
@@ -0,0 +1,134 @@
...
+        for r in root.iterfind(self.dmr_namespace + "chunk"):
+            chunk_pos = (
+                np.zeros(len(chunks), dtype=int)
+                if "chunkPositionInArray" not in r.attrib
+                else np.asarray(ast.literal_eval(r.attrib["chunkPositionInArray"]))
+            )
+            chunk_num = (
+                chunk_pos // chunks
+            )  # [0,1023,10235] // [1, 1023, 2047] -> [0,1,5]
+            chunk_key = ".".join(map(str, chunk_num))  # [0,0,1] -> "0.0.1"
```
TomNicholas, line 128: I have a `join` function in `virtualizarr.zarr` for doing this.

virtualizarr/dmrpp.py
```diff
@@ -0,0 +1,134 @@
...
+        "Float64": "float64",
+        "String": "str",
+    }
+
+    def __init__(self, dmr: str):
+        self.root = ET.fromstring(dmr)
+        self.data_filepath = self.root.attrib["name"]
+        self.global_dims = {}
+
+    def parse_dataset(self):
```
TomNicholas, line 36:
```suggestion
    def parse_dataset(self) -> xr.Dataset:
```

agoodm:
Thanks for taking a look and giving my suggested changes to the chunk key parsing a try @ayushnag ! 

Continuing the discussion on performance I think the remaining bottlenecks (aside from your point about I/O in the cloud maybe) with this now lie primarily outside the scope of this work, and I don't expect changing XML readers to make a significant improvement.

virtualizarr/xarray.py
```diff
@@ -68,6 +69,8 @@ def open_virtual_dataset(
     virtual_array_class
         Virtual array class to use to represent the references to the chunks in each on-disk array.
         Currently can only be ManifestArray, but once VirtualZarrArray is implemented the default should be changed to that.
+    group : str, default None
+        Group path within the dataset to open. For example netcdf4 and hdf5 groups
```
TomNicholas, line 73: It would be nice to separate out the addition of this kwarg into a separate pull request, and implement it for the existing HDF5 reader. Then this PR wouldn't need to change the API of `open_virtual_dataset`.

virtualizarr/readers/dmrpp.py
```diff
@@ -0,0 +1,331 @@
...
+class DMRParser:
+    """
+    Parses a DMR file and creates a virtual xr.Dataset.
+    Handles groups, dimensions, coordinates, data variables, encoding, chunk manifests, and attributes.
+    """
+
+    _ns = {
+        "dap": "http://xml.opendap.org/ns/DAP/4.0#",
+        "dmr": "http://xml.opendap.org/dap/dmrpp/1.0.0#",
+    }
```
TomNicholas, line 22: Can you explain what these are for with comments?

virtualizarr/readers/dmrpp.py
```diff
@@ -0,0 +1,331 @@
+from typing import Optional
+from xml.etree import ElementTree as ET
```
TomNicholas, line 2: Is this the only extra import required? (And this is a built-in python library module right?)

virtualizarr/readers/dmrpp.py
ayushnag: Yes added

virtualizarr/readers/dmrpp.py
ayushnag: Yes this is the only extra import and it is built in

virtualizarr/manifests/manifest.py
ayushnag, line 1: This change had to be made since I ran into some large offsets with the Atlas ICE-SAT dataset. Here is an example error: `OverflowError: Python integer 6751178683 out of bounds for int32`

ayushnag:
Some questions about writing unit tests:
- How to load test dmrpp’s?
    - These files are available over https but need netrc login (NASA Earthdata authentication)
    - I will check how `earthaccess` gets creds and does testing
- What should I compare my result to?
    - My understanding is that the dmr parsed dataset should match the dataset made by vz.open_dataset(“data.nc”)
    - Should everything match or are there some main portions to check? (dims, attrs, variables)
    - Related to https://github.com/zarr-developers/VirtualiZarr/issues/161

ayushnag:
@TomNicholas this PR should be ready to go now. You can take a look at the code again if you wish and I can add updates. I have also added docs and release notes.

virtualizarr/xarray.py
TomNicholas: see #165

virtualizarr/manifests/manifest.py
TomNicholas: Well that justifies https://github.com/zarr-developers/VirtualiZarr/pull/177!

virtualizarr/readers/dmrpp.py
```diff
@@ -0,0 +1,647 @@
...
+                coord_vars[var_tag.attrib["name"]] = variable
+            else:
+                data_vars[var_tag.attrib["name"]] = variable
+        # Attributes
+        attrs: dict[str, str] = {}
+        for attr_tag in self.root.iterfind("dap:Attribute", self._ns):
+            attrs.update(self._parse_attribute(attr_tag))
+        return xr.Dataset(
+            data_vars=data_vars,
+            coords=xr.Coordinates(coords=coord_vars, indexes={}),
```
TomNicholas, line 335: Should the `indexes` variable here technically be passed through from `open_virtual_dataset`? Right now it doesn't matter, but it would in order to support #18.

virtualizarr/readers/dmrpp.py
```diff
@@ -0,0 +1,647 @@
...
+                if dtype != np.object_
+                else value_tag.text
+            )
+            if val == "*":
+                val = np.nan
+            values.append(val)
+        attr[attr_tag.attrib["name"]] = values[0] if len(values) == 1 else values
+        return attr
+
+    def _parse_filters(
```
TomNicholas, line 545: A lot of these methods don't technically need to be methods, because they don't use `self`. It could all be functional instead. But this isn't very important, just a note.

virtualizarr/readers/dmrpp.py
```diff
@@ -0,0 +1,647 @@
...
+                    for i in range(len(chunks_shape))
+                ]
+            # [0,1,5] -> "0.1.5"
+            chunk_key = ChunkKey(chunk_key_template.format(*chunk_num))
+            chunkmanifest[chunk_key] = {
+                "path": self.data_filepath,
+                "offset": int(chunk_tag.attrib["offset"]),
+                "length": int(chunk_tag.attrib["nBytes"]),
+            }
+        return ChunkManifest(entries=chunkmanifest)
```
TomNicholas, line 647: Do you think you could rewrite this to use `ChunkManifest.from_arrays()`? It would potentially be a lot more performant. (But this could also be left to a later PR.)

TomNicholas (commented):
This looks good, but I'm still unclear if you prefer to merge into EarthAccess or here.

virtualizarr/readers/dmrpp.py
ayushnag: Yes you're right these can be standalone. Even many of the functions that use self are just using the class constants. I will update after the initial merge

virtualizarr/readers/dmrpp.py
ayushnag: Oh nice didn't know about that function. I will give it a try

virtualizarr/readers/dmrpp.py
ayushnag: Yes you're right, I will fix that

docs/api.rst
```diff
@@ -66,3 +66,14 @@ VirtualiZarr's :py:class:`~virtualizarr.ManifestArray` objects support a limited
...
+Readers
+=======
+
+.. currentmodule:: virtualizarr.readers.dmrpp
+.. autosummary::
+    :nosignatures:
+    :toctree: generated/
+
+    DMRParser
+    DMRParser.parse_dataset
```
TomNicholas, line 79: Do these need to be public? What's the scenario where someone actually calls these instead of just using `open_virtual_dataset(..., filetype='dmr++')`?

virtualizarr/readers/dmrpp.py
TomNicholas: This and https://github.com/zarr-developers/VirtualiZarr/pull/113#discussion_r1729309485 are the only comment that I think actually need to be addressed before merging, because they affect public API. Everything else can be left until later if you prefer.

ayushnag: I was trying to get this fixed however I think there is an issue which is that the indexes cannot be created since we don't have the path to the actual data file. Technically the dmrpp file does have a data path but it is only the file name and not the full path. E.g. "data.nc" instead of "s3://.../data.nc". So in that case if I add support for the indexes param it will fail when `indexes=None` when usually in virtualizarr `indexes=None` indicates auto-create indexes.

What I could do instead is to treat `indexes=None` and `indexes={}` as the same but raise a warning if `indexes=None` that the behavior is not as expected. However I can still accept manually created indexes (e.g. {"time": Index})

docs/api.rst
ayushnag: Hmm yes you're right. I think I thought since it is a class on it's own it needs some documentation but I can keep all my example docs and just remove the API docs portion

virtualizarr/readers/dmrpp.py
TomNicholas: You can just raise a `NotImplementedError` on `indexes={}`. But at least something that isn't silently doing an inconsistent behaviour to the other readers.

ayushnag: I thought `indexes={}` would be acceptable since that is how most people indicate that indexes should not be created?

TomNicholas: I meant indexes=None sorry

virtualizarr/readers/dmrpp.py
```diff
@@ -0,0 +1,683 @@
...
+        # Dimension names
+        dim_tags = var_tag.findall("dap:Dim", self._ns)
+        dim_shapes = self._parse_multi_dims(dim_tags, dataset_dims)
+        # convert DAP dtype to numpy dtype
+        dtype = np.dtype(
+            self._dap_np_dtype[var_tag.tag.removeprefix("{" + self._ns["dap"] + "}")]
+        )
+        # Chunks and Filters
+        filters = None
+        shape: tuple[int] = tuple(dim_shapes.values())
```
TomNicholas, line 514: See https://github.com/zarr-developers/VirtualiZarr/actions/runs/10561663864/job/29257886654

```suggestion
        shape: tuple[int, ...] = tuple(dim_shapes.values())
```

TomNicholas (approved):

ayushnag:
@TomNicholas do you know why the test is failing? I have added dmrpp as a valid file type so it should be recognized in the unit test. It also passes when I run network tests locally

TomNicholas:
I'm just looking at that. I suspect it's just a bad merge from `main`. Have you tried pulling the latest changes to this branch and running that locally?

virtualizarr/xarray.py
```diff
@@ -137,6 +137,21 @@ def open_virtual_dataset(
         return open_virtual_dataset_from_v3_store(
             storepath=filepath, drop_variables=drop_variables, indexes=indexes
         )
+    if filetype == "dmrpp":
```
TomNicholas, line 140:
```suggestion
    filetype.name.lower() == "dmrpp":
```

virtualizarr/xarray.py
```diff
@@ -137,6 +137,21 @@ def open_virtual_dataset(
         return open_virtual_dataset_from_v3_store(
             storepath=filepath, drop_variables=drop_variables, indexes=indexes
         )
+    elif filetype.name.lower() == "dmrpp":
```
TomNicholas, line 140: Try another syntax for the `FileType` object
```suggestion
    elif filetype == FileType.dmrpp:
```

TomNicholas:
Amazing piece of work here @ayushnag !!

PR #122, opened 2024-05-20
Decoding `cftime_variables`

### LINKED ISSUE #117: Option to interpret variables using cftime
https://github.com/TomNicholas/VirtualiZarr/issues/68#issuecomment-2113983642 explains how we don't currently have support for virtualizing encoded time variables. 

To handle this it looks like we need to an another option to `open_virtual_dataset`, a `cftime_vars` kwarg, which specifies which variables should be interpreted using cftime. (I think `cftime_vars` should be a subset of `loadable_vars`.) This suggestion comes from how kerchunk deals with this problem:

https://fsspec.github.io/kerchunk/reference.html#kerchunk.combine.MultiZarrToZarr

```
”cf:{var}”, interpret the value of var using cftime, returning a datetime. 
These will be automatically re-encoded with cftime, unless you specify an “M8[*]” 
dtype for the coordinate, in which case a conversion will be attempted.
```

_Originally posted by @dcherian in https://github.com/TomNicholas/VirtualiZarr/issues/73#issuecomment-2047989180_

jsignell:
- [x] Closes #117
- [x] Tests added

This is what I have started working on for the cftime_variables. I am going to be on vacation for a week so I thought I should show what I've gotten up to in case there is some urgency around this, but it is pretty WIP 🥲

### DIFF
--- a/virtualizarr/kerchunk.py
+++ b/virtualizarr/kerchunk.py
@@ -277,11 +277,8 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
                     f"Cannot serialize loaded variable {var_name}, as it is encoded with an offset"
                 )
             if "calendar" in var.encoding:
-                print("JULIA 0 ", var)
                 np_arr = CFDatetimeCoder().encode(var.copy(), name=var_name).values
-                print("JULIA 1 ", var, np_arr)
 
-        print("JULIA 2", var.encoding, var.attrs, np_arr)
         # This encoding is what kerchunk does when it "inlines" data, see https://github.com/fsspec/kerchunk/blob/a0c4f3b828d37f6d07995925b324595af68c4a19/kerchunk/hdf.py#L472
         byte_data = np_arr.tobytes()
         # TODO do I really need to encode then decode like this?
--- a/virtualizarr/kerchunk.py
+++ b/virtualizarr/kerchunk.py
@@ -7,9 +7,10 @@ from typing import Any, NewType, Optional, cast
 import numpy as np
 import ujson  # type: ignore
 import xarray as xr
+from xarray.coding.times import CFDatetimeCoder
 
 from virtualizarr.manifests.manifest import join
-from virtualizarr.utils import _fsspec_openfile_from_filepath, encode_cftime
+from virtualizarr.utils import _fsspec_openfile_from_filepath
 from virtualizarr.zarr import ZArray, ZAttrs
 
 # Distinguishing these via type hints makes it a lot easier to mentally keep track of what the opaque kerchunk "reference dicts" actually mean
@@ -274,8 +275,11 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
                     f"Cannot serialize loaded variable {var_name}, as it is encoded with an offset"
                 )
             if "calendar" in var.encoding:
-                np_arr = encode_cftime(var)
+                print("JULIA 0 ", var)
+                np_arr = CFDatetimeCoder().encode(var.copy(), name=var_name).values
+                print("JULIA 1 ", var, np_arr)
 
+        print("JULIA 2", var.encoding, var.attrs, np_arr)
         # This encoding is what kerchunk does when it "inlines" data, see https://github.com/fsspec/kerchunk/blob/a0c4f3b828d37f6d07995925b324595af68c4a19/kerchunk/hdf.py#L472
         byte_data = np_arr.tobytes()
         # TODO do I really need to encode then decode like this?
--- a/virtualizarr/kerchunk.py
+++ b/virtualizarr/kerchunk.py
@@ -9,7 +9,7 @@ import ujson  # type: ignore
 import xarray as xr
 
 from virtualizarr.manifests.manifest import join
-from virtualizarr.utils import _fsspec_openfile_from_filepath
+from virtualizarr.utils import _fsspec_openfile_from_filepath, encode_cftime
 from virtualizarr.zarr import ZArray, ZAttrs
 
 # Distinguishing these via type hints makes it a lot easier to mentally keep track of what the opaque kerchunk "reference dicts" actually mean
@@ -267,14 +267,7 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
                     f"Cannot serialize loaded variable {var_name}, as it is encoded with an offset"
                 )
             if "calendar" in var.encoding:
-                import cftime
-
-                calendar = var.attrs.get(
-                    "calendar", var.encoding.get("calendar", "standard")
-                )
-                units = var.attrs.get("units", var.encoding["units"])
-
-                np_arr = cftime.date2num(var, calendar=calendar, units=units).ravel()
+                np_arr = encode_cftime(var)
 
         # This encoding is what kerchunk does when it "inlines" data, see https://github.com/fsspec/kerchunk/blob/a0c4f3b828d37f6d07995925b324595af68c4a19/kerchunk/hdf.py#L472
         byte_data = np_arr.tobytes()
--- a/virtualizarr/kerchunk.py
+++ b/virtualizarr/kerchunk.py
@@ -262,9 +262,16 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
                     f"Cannot serialize loaded variable {var_name}, as it is encoded with an offset"
                 )
             if "calendar" in var.encoding:
-                raise NotImplementedError(
-                    f"Cannot serialize loaded variable {var_name}, as it is encoded with a calendar"
+                import cftime
+
+                calendar = var.attrs.get(
+                    "calendar", var.encoding.get("calendar", "standard")
                 )
+                units = var.attrs.get("units", var.encoding["units"])
+
+                data = cftime.date2num(var, calendar=calendar, units=units).ravel()
+                var = var.copy(data=data)
+                np_arr = var.to_numpy()
 
         # This encoding is what kerchunk does when it "inlines" data, see https://github.com/fsspec/kerchunk/blob/a0c4f3b828d37f6d07995925b324595af68c4a19/kerchunk/hdf.py#L472
         byte_data = np_arr.tobytes()
--- a/virtualizarr/kerchunk.py
+++ b/virtualizarr/kerchunk.py
@@ -47,6 +47,8 @@ class NumpyEncoder(json.JSONEncoder):
             return obj.tolist()  # Convert NumPy array to Python list
         elif isinstance(obj, np.generic):
             return obj.item()  # Convert NumPy scalar to Python scalar
+        elif isinstance(obj, np.dtype):
+            return str(obj)
         return json.JSONEncoder.default(self, obj)
 
 
@@ -269,9 +271,7 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
                 )
                 units = var.attrs.get("units", var.encoding["units"])
 
-                data = cftime.date2num(var, calendar=calendar, units=units).ravel()
-                var = var.copy(data=data)
-                np_arr = var.to_numpy()
+                np_arr = cftime.date2num(var, calendar=calendar, units=units).ravel()
 
         # This encoding is what kerchunk does when it "inlines" data, see https://github.com/fsspec/kerchunk/blob/a0c4f3b828d37f6d07995925b324595af68c4a19/kerchunk/hdf.py#L472
         byte_data = np_arr.tobytes()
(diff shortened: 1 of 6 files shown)


virtualizarr/kerchunk.py
```diff
@@ -262,9 +262,16 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
...
+                import cftime
+
+                calendar = var.attrs.get(
+                    "calendar", var.encoding.get("calendar", "standard")
                 )
+                units = var.attrs.get("units", var.encoding["units"])
+
+                data = cftime.date2num(var, calendar=calendar, units=units).ravel()
+                var = var.copy(data=data)
+                np_arr = var.to_numpy()
```
dcherian, line 274: Thinking out loud: Should this be a Zarr Codec instead?

jsignell: naive question: do you mean a new CFTime codec or just a number codec rather than numpy?

jsignell:
Ok! I think the issue was that I was encoding the var to a number but then I wasn't actually storing the encoding information in zarr 😬 . So now that that is resolved things seem to be working better but @jbusecke if you have a test you want to try go for it!

virtualizarr/xarray.py
```diff
@@ -142,6 +162,30 @@ def open_virtual_dataset(
...
+
+                var = ds[name]
+                calendar = var.attrs.get(
+                    "calendar", var.encoding.get("calendar", "standard")
+                )
+                units = var.attrs.get("units", var.encoding["units"])
+
+                # undoing CF recoding in original input
+                values = []
+                for c in var.values:
```
jsignell, line 176: This is a for loop iterating over each value in an array and decoding/encoding each one. Is this really necessary?

jbusecke:
If folks here have a minute, I just tested this PR over at https://github.com/jbusecke/esgf-virtual-zarr-data-access/pull/10 with real world CMIP data. I think it generally works, but the attributes of the resulting time coordinate are different. Would love to get some input on that.

virtualizarr/kerchunk.py
TomNicholas: Would that just be a refactoring in this PR? Like just separate out the logic here into a Codec class that could later be moved upstream?

virtualizarr/xarray.py
```diff
@@ -94,6 +102,18 @@ def open_virtual_dataset(
...
+        cftime_variables = []
+    elif isinstance(cftime_variables, str):
+        cftime_variables = [cftime_variables]
+    else:
+        cftime_variables = list(cftime_variables)
+
+    if not set(cftime_variables).issubset(set(loadable_variables)):
+        raise ValueError(
+            "All ``cftime_variables`` must be included in ``loadable_variables``"
+        )
```
TomNicholas, line 115: Nit: We can make this error message a little more useful by displaying the set of `cftime_variables` which were not in `loadable_variables`.

virtualizarr/xarray.py
```diff
@@ -142,6 +162,30 @@ def open_virtual_dataset(
...
             }
 
+            for name in cftime_variables:
+                import cftime
+
+                var = ds[name]
+                calendar = var.attrs.get(
+                    "calendar", var.encoding.get("calendar", "standard")
+                )
+                units = var.attrs.get("units", var.encoding["units"])
```
TomNicholas, line 172: Can we abstract this out into a function? That would also help make it clearer what the Zarr Codec equivalent would be.

virtualizarr/xarray.py
TomNicholas: 🤷 I don't think this is a massive issue for now, given that most cftime variables are going to be 1D and therefore likely small.

TomNicholas (commented):
I would love to get this merged!

This is an important feature, so deserves a note in the usage documentation too.

virtualizarr/tests/test_integration.py
```diff
@@ -66,7 +66,10 @@ def test_kerchunk_roundtrip_no_concat(self, tmpdir, format):
 
     def test_kerchunk_roundtrip_concat(self, tmpdir, format):
         # set up example xarray dataset
-        ds = xr.tutorial.open_dataset("air_temperature", decode_times=False)
+        ds = xr.tutorial.open_dataset("air_temperature", decode_times=True)
+
+        # delete the encoding since this is now incorrect
+        del ds.time.encoding["units"]
```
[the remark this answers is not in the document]
jsignell: If you leave it with `decode_times=False` then `ds.time.encoding` doesn't have a `units` attr. So I take this to mean the `units` are a default added by xarray. That being said this test no longer passes as written. 

If I un-delete the units we get rounding to the second which is incorrect.

virtualizarr/tests/test_integration.py
```diff
@@ -66,7 +66,10 @@ def test_kerchunk_roundtrip_no_concat(self, tmpdir, format):
 
     def test_kerchunk_roundtrip_concat(self, tmpdir, format):
         # set up example xarray dataset
-        ds = xr.tutorial.open_dataset("air_temperature", decode_times=False)
+        ds = xr.tutorial.open_dataset("air_temperature", decode_times=True)
```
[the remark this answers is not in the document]
jsignell: yes!

virtualizarr/xarray.py
```diff
@@ -68,6 +71,11 @@ def open_virtual_dataset(
     virtual_array_class
         Virtual array class to use to represent the references to the chunks in each on-disk array.
         Currently can only be ManifestArray, but once VirtualZarrArray is implemented the default should be changed to that.
+    cftime_variables : list[str], default is None
+        Interpret the value of specified vars using cftime, returning a datetime.
+        These will be automatically re-encoded with cftime, unless you specify an “M8[*]”
+        dtype for the coordinate, in which case a conversion will be attempted. This list must be a subset
```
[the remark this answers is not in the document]
jsignell: Correct! I copied it over and intended to ammend...

virtualizarr/xarray.py
```diff
@@ -142,6 +162,30 @@ def open_virtual_dataset(
...
+                    "calendar", var.encoding.get("calendar", "standard")
+                )
+                units = var.attrs.get("units", var.encoding["units"])
+
+                # undoing CF recoding in original input
+                values = []
+                for c in var.values:
+                    value = cftime.num2date(
+                        cftime.date2num(
+                            datetime.datetime.fromisoformat(str(c).split(".")[0]),
```
jsignell, line 179: This is looking very suspicious for my second rounding issue 🤦🏻

virtualizarr/xarray.py
jsignell: yes!

virtualizarr/tests/test_integration.py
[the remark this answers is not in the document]
jsignell: I am removing this line for now.

jsignell:
Thanks for the review! I think I have addressed all the comments.

virtualizarr/tests/test_integration.py
```diff
@@ -86,7 +97,9 @@ def test_kerchunk_roundtrip_concat(self, tmpdir, format):
         vds.virtualize.to_kerchunk(f"{tmpdir}/refs.{format}", format=format)
 
         # use fsspec to read the dataset from disk via the zarr store
-        roundtrip = xr.open_dataset(f"{tmpdir}/refs.{format}", engine="kerchunk")
+        roundtrip = xr.open_dataset(
+            f"{tmpdir}/refs.{format}", engine="kerchunk", decode_times=decode_times
+        )
 
         # assert equal to original dataset
         xrt.assert_equal(roundtrip, ds)
```
jsignell, line 105: This test doesn't pass yet, but it's getting a lot closer! It's basically a floating point difference at this point. I tried switching to `xrt.assert_allclose` but I'm not sure it understands times because it still raises. If I do `(roundtrip.time - ds.time).sum()` I get zero so I do think the times are pretty darn close.

virtualizarr/xarray.py
jsignell: I dropped the `.split`

virtualizarr/tests/test_integration.py
jsignell: I tried using numpy directly to get to the bottom of things:

```python
>>> from numpy.testing import assert_allclose
>>> assert_allclose(roundtrip.time.values, ds.time.values)
*** numpy.exceptions.DTypePromotionError: The DType <class 'numpy.dtypes.DateTime64DType'> could not be promoted by <class 'numpy.dtypes._PyFloatDType'>. This means that no common DType exists for the given inputs. For example they cannot be stored in a single array unless the dtype is `object`. The full list of DTypes is: (<class 'numpy.dtypes.DateTime64DType'>, <class 'numpy.dtypes._PyFloatDType'>)
```

jsignell: See 0ffce4f for a rather silly solution

jsignell:
Ok @TomNicholas I am pausing. Let me know how this is looking to you

TomNicholas:
Thanks for going so deep into this @jsignell! To add more confusion - do you think what I'm seeing in https://github.com/zarr-developers/VirtualiZarr/pull/154#discussion_r1650170676 is related to your close-but-not-equal issue?

TomNicholas:
@jsignell you might want to merge in #154 and #158.

Also sorry to push the finish line farther away, but what do you think about the idea in https://github.com/zarr-developers/VirtualiZarr/issues/117#issuecomment-2187406127?

jsignell:
> @jsignell you might want to merge in #154 and #158.

✅ 

> Also sorry to push the finish line farther away, but what do you think about the idea in [#117 (comment)](https://github.com/zarr-developers/VirtualiZarr/issues/117#issuecomment-2187406127)?

No worries, better to do this the right way. I modeled the approach in this PR on kerchunk, but I will take a look at what is going on in xarray.

TomNicholas:
> No worries, better to do this the right way. I modeled the approach in this PR on kerchunk, but I will take a look at what is going on in xarray.

❤️ 

So I was looking at this in more detail last night, and I think we might just want to replace your `encode_cftime` and `recode_cftime` with importing [xarray's `CFDatetimeCoder`](https://github.com/pydata/xarray/blob/b5180749d351f8b85fd39677bf137caaa90288a7/xarray/coding/times.py#L958) (and possibly the [`CFTimeDeltaEncoder`](https://github.com/pydata/xarray/blob/b5180749d351f8b85fd39677bf137caaa90288a7/xarray/coding/times.py#L1001) too). As far as I can tell (and please tell me if I'm wrong) those are basically more mature versions of the functions you've written here. Those classes are private but stable interfaces, and using them here could be an argument for factoring them out of xarray at some point (or exposing them more publicly).

Those encoders are called by `xr.decode_cf` [here](https://github.com/pydata/xarray/blob/56209bd9a3192e4f1e82c21e5ffcf4c3bacaaae3/xarray/conventions.py#L288), but replacing everything with `xr.decode_cf` seems like a longer-term idea that is beyond the scope of this PR (but using xarray's encoders would build towards it).

virtualizarr/kerchunk.py
sharkinsspatial: @TomNicholas I've looked at a tiny portion of upstreaming CF logic to a codec for CF `scale_factor` and `add_offset` but this case has the benefit of a pre-existing `numcodec` [FixedScaleAndOffset](https://numcodecs.readthedocs.io/en/stable/fixedscaleoffset.html).

TomNicholas: There is also a [`CFScaleOffsetCoder`](https://github.com/pydata/xarray/blob/b5180749d351f8b85fd39677bf137caaa90288a7/xarray/coding/variables.py#L453) in xarray too...

jsignell:
Ok I think I've finally got it working.

virtualizarr/kerchunk.py
```diff
@@ -274,10 +277,11 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
                     f"Cannot serialize loaded variable {var_name}, as it is encoded with an offset"
                 )
             if "calendar" in var.encoding:
-                raise NotImplementedError(
-                    f"Cannot serialize loaded variable {var_name}, as it is encoded with a calendar"
-                )
+                print("JULIA 0 ", var)
```
TomNicholas: Nice to know I'm not the only person who still uses print statements as their primary method of debugging 😂

virtualizarr/xarray.py
```diff
@@ -127,7 +148,9 @@ def open_virtual_dataset(
                 filepath=filepath, reader_options=reader_options
             )
 
-            ds = xr.open_dataset(fpath, drop_variables=drop_variables)
+            ds = xr.open_dataset(
+                fpath, drop_variables=drop_variables, decode_times=False
```
jsignell, line 152: This is a change that required changes to tests (see: cebb031) but I _do_ think this is more correct. We don't want to be decoding things if it isn't explicitly requested right?

TomNicholas (approved):

virtualizarr/kerchunk.py
jsignell: 😵‍💫  yeah I'm pretty advanced.

virtualizarr/xarray.py
jsignell: @TomNicholas really want to draw attention to this

TomNicholas: I agree. We can also make this more auto-magic later if we want. I'm not sure any of the existing tests were covering this.

jsignell:
Let me know if you are looking for any more changes or feedback! Otherwise this is done from my perspective 😸

TomNicholas:
Great contribution! Thanks so much @jsignell !

PR #128, opened 2024-05-28
Pass args and add test

abarciauskas-bgse:
I noticed that reader_options were not being passed through to `read_kerchunk_references_from_file` in `open_virtual_dataset` when I was trying out the library with some files on S3 (e.g. `open_virtual_dataset(s3_link, reader_options={'storage_options': aws_creds})` would get a Forbidden error, see https://gist.github.com/abarciauskas-bgse/629702733946b1ec0b008f659dbc3406 for reproducible example).

This PR just adds one line so those args are passed through. I added a test but not sure if that is overkill for such a small change.

### DIFF
--- a/virtualizarr/tests/test_xarray.py
+++ b/virtualizarr/tests/test_xarray.py
@@ -1,4 +1,5 @@
 from collections.abc import Mapping
+from unittest.mock import patch
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


TomNicholas:
This looks great, thanks @abarciauskas-bgse !

PR #133, opened 2024-06-10
Add dunder version to top-level __init__.py

maxrjones:
This PR makes the following work:

```
python -c "import virtualizarr; print(virtualizarr.__version__)"
```

Welcome any feedback, in particular if there's a reason to use setuptools_scm's `write_to` rather than `importlib.metadata`.
- [ ] Closes #xxxx
- [ ] Tests added

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,11 +61,7 @@ requires = [
 ]
 
 [tool.setuptools_scm]
-write_to = "virtualizarr/_version.py"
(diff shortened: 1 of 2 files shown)


TomNicholas:
Thanks!

> in particular if there's a reason to use setuptools_scm's `write_to` rather than `importlib.metadata`

No idea 😅 I'm happy to merge this now though?

PR #134, opened 2024-06-10
Add release flow

maxrjones:
Builds off https://github.com/zarr-developers/VirtualiZarr/pull/133.

This PR adds a GitHub workflow to test distributions using TestPyPI and upload releases to PyPI.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`

PR #136, opened 2024-06-12
Add release workflow

maxrjones:
This PR adds a GitHub workflow to test distributions using TestPyPI and upload releases to PyPI. (Same as #134 but not off a fork to allow testing the workflow)

- [ ] Closes #xxxx
- [ ] Tests added

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,9 +51,9 @@ jobs:
   test-built-dist:
     needs: build-artifacts
     runs-on: ubuntu-latest
-    # environment: # TODO: Have a VirtualiZarr maintainer create a test-release environment
-    #   name: test-release
-    #   url: https://test.pypi.org/p/virtualizarr
+    environment:
+      name: test-release
+      url: https://test.pypi.org/p/virtualizarr
     permissions:
       id-token: write # this permission is mandatory for trusted publishing
     steps:
@@ -87,9 +87,9 @@ jobs:
     needs: test-built-dist
     if: github.event_name == 'release'
     runs-on: ubuntu-latest
-    # environment: # TODO: Have a VirtualiZarr maintainer create a release environment
-    #   name: release
-    #   url: https://pypi.org/p/virtualizarr
+    environment:
+      name: release
+      url: https://pypi.org/p/virtualizarr
     permissions:
       id-token: write # this permission is mandatory for trusted publishing
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,10 +27,18 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install setuptools setuptools-scm wheel twine check-manifest
-      - name: Build tarball and wheels
-        run: |
           git clean -xdf
           git restore -SW .
+      # This step is only necessary for testing purposes and for TestPyPI
+      - name: Fix up version string for TestPyPI
+        if: ${{ !startsWith(github.ref, 'refs/tags') }}
+        run: |
+          # Change setuptools-scm local_scheme to "no-local-version" so the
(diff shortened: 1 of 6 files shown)


maxrjones:
@TomNicholas after merging, this will publish virtualizarr on PyPI when a new tag is pushed to the repo, which can also be done automatically via the GitHub releases page.

.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+        with:
+          name: releases
+          path: dist
+
+  test-built-dist:
+    needs: build-artifacts
+    runs-on: ubuntu-latest
+    environment:
+      name: test-release
+      url: https://test.pypi.org/p/virtualizarr
```
TomNicholas, line 63: Have you reserved the name on pypi? Or will it do that the first time this is run?

TomNicholas (approved):
Incredible @maxrjones ! Thank you so much!

.github/workflows/release.yml
TomNicholas: That's totally fine - yes please add me to reduce the bus factor (TomNicholas) on pypi. Will anyone with the ability to push a tag to the repo be able to release?

maxrjones: I'm going to add a super simple contributing guide that goes over this structure so it's easier to find these answers in the future. To make that, are you planning to use SemVer or year.month versioning for the tags/releases?

TomNicholas: Great idea! 

> are you planning to use SemVer or year.month versioning

I prefer SemVer (or [EffVer](https://jacobtomlinson.dev/effver/)). I don't expect this package to have so many ongoing changes that it needs really frequent releases indefinitely into the future - the scope is pretty targeted so I'm hoping it converges to some stable fixed point, with only bugfix releases after that. (We're a fair way from that now though.)

maxrjones: @TomNicholas can you let me know if https://virtualizarr--136.org.readthedocs.build/en/136/contributing.html#making-a-release answers your questions?

TomNicholas: This is great! It's almost exactly the same as [xarray's release guide](https://github.com/pydata/xarray/blob/main/HOW_TO_RELEASE.md), which is blessedly simple :pray:

I think the only differences are:
1. The need to go and manually approve the deployment
2. Xarray also publishes to conda-forge at the same time

maxrjones: 1. We could change this so there aren't any environment protection rules if you rather
2. True, I'd be glad to set up the conda-forge package which would also be automatic without any more changes needed via GitHub actions. It's easier to do it after it's on PyPI because then you can just use [grayskull](https://pypi.org/project/grayskull/).

TomNicholas: > We could change this so there aren't any environment protection rules if you rather

This already requires the would-be-releaser to have permissions to create tags right? Which is already restricted to `virtualizarr-core-devs` I believe?

> True, I'd be glad to set up the conda-forge package which would also be automatic without any more changes needed via GitHub actions. It's easier to do it after it's on PyPI because then you can more easily use [grayskull](https://pypi.org/project/grayskull/).

That would also be great. Also I love the name of that package.

maxrjones: This already requires the would-be-releaser to have permissions to create tags right? Which is already restricted to virtualizarr-core-devs I believe?

AFAIK anyone with write permissions can create tags. It's possible to restrict tags beyond that to admins (i.e., virtualizarr-core-devs) via protected tags (https://github.com/zarr-developers/VirtualiZarr/settings/tag_protection). I'm not sure who in the zarr-developers organization has write permissions to all repos but not admin permissions, which is who that restriction could apply to. For now, I've just removed the workflow protection and will merge the PR.

PR #138, opened 2024-06-12
Replace np.NaN with np.nan in preparation for numpy 2.0

TomNicholas:
See https://numpy.org/devdocs/numpy_2_0_migration_guide.html#main-namespace

- [ ] ~~Closes #xxxx~~
- [ ] ~~Tests added~~
- [ ] ~~Changes are documented in `docs/releases.rst`~~

### DIFF
--- a/virtualizarr/tests/test_kerchunk.py
+++ b/virtualizarr/tests/test_kerchunk.py
@@ -40,7 +40,7 @@ def test_dataset_from_df_refs():
 
     assert da.data.zarray.compressor is None
     assert da.data.zarray.filters is None
-    assert da.data.zarray.fill_value is np.NaN
+    assert da.data.zarray.fill_value is np.nan
     assert da.data.zarray.order == "C"
 
     assert da.data.manifest.dict() == {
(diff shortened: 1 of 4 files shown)


TomNicholas:
@maxrjones the test failure here looks weird - is something going wrong with the workflows you set up earlier?

maxrjones:
> @maxrjones the test failure here looks weird - is something going wrong with the workflows you set up earlier?

ah sorry about that. in order to test distributions on non-tagged commits with TestPyPI, we need to use 'no-local-version' (e.g., 0.1.dev159) rather than the default of 'guess-next-dev', but also there cannot be multiple releases with the same name. I used up a bunch of the devN names on TestPyPI (see https://test.pypi.org/manage/project/virtualizarr/releases/) when testing out the workflow, so the error is reporting that name conflict. This won't be a problem if we only run the workflow on releases (configured in https://github.com/zarr-developers/VirtualiZarr/pull/140), which is how xarray has their release workflow configured.

In summary, you can ignore the issue and merge in https://github.com/zarr-developers/VirtualiZarr/pull/140 so it doesn't show up if you want.

PR #143, opened 2024-06-14
Use magic bytes to identify file formats

scottyhq:
- [x] Closes #142
- [x] Tests added
- [x] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`

PR #139, opened 2024-06-12
Hypothesis test broadcasting

TomNicholas:
This definitely found a few bugs, but I'm not sure there's any point fixing them in `main` when I'm about to change the entire implementation in #107 anyway...

- [x] Works towards #9 for array functions on the `ManifestArray` class
- [x] Tests added
- [ ] ~~Changes are documented in `docs/releases.rst`~~
- [ ] ~~New functions/methods are listed in `api.rst`~~

PR #144, opened 2024-06-17
Mark tests which require network access

TomNicholas:
Implements @scottyhq's suggestion in https://github.com/zarr-developers/VirtualiZarr/pull/143#issuecomment-2174029069.

Makes the test suite run considerably faster locally by default because you don't have to wait for the tests that access S3 to complete.

- [ ] ~~Closes #xxxx~~
- [ ] ~~Tests added~~

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
 
       - name: Running Tests
         run: |
-          python -m pytest --cov=./ --cov-report=xml --verbose
+          python -m pytest --run-network-tests --cov=./ --cov-report=xml --verbose
 
       - name: Upload code coverage to Codecov
         uses: codecov/codecov-action@v3.1.4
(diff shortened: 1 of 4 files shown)


TomNicholas:
This now works, the test that fails is the hypothesis test I added in #139.

TomNicholas:
Actually this appears to have broken the ability to run tests using `pytest` - I get an error `ValueError: no option named 'run_network_tests'`.

I can still run the tests if I specify the directory to run from, i.e. `pytest -vv ./virtualizarr` works. I have no idea why because I can't see what's different about this setup compared to xarray, which doesn't have this problem 😬

TomNicholas:
Realized the problem is here likely that conftest.py should be in the root directory...

TomNicholas:
(yeah https://github.com/zarr-developers/VirtualiZarr/commit/4d13b35c5c816413a966b0a7f4e60da1d40e55ff fixed it)

Issue #114, opened 2024-05-14
Ugly error from xarray chunkmanager when loading ManifestArray


TomNicholas:
If you create the `air1.nc` and `air2.nc` files in the same way as in the docs, then concatenate them with `compat='equals'`, you get a super ugly error:

```python
In [9]: vds1 = open_virtual_dataset('air1.nc', indexes={})

In [10]: vds2 = open_virtual_dataset('air2.nc', indexes={})

In [11]: xr.concat([vds1, vds2], dim='time', compat='equals')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[11], line 1
----> 1 xr.concat([vds1, vds2], dim='time', compat='equals')

File ~/Documents/Work/Code/xarray/xarray/core/concat.py:276, in concat(objs, dim, data_vars, coords, compat, positions, fill_value, join, combine_attrs, create_index_for_new_dim)
    263     return _dataarray_concat(
    264         objs,
    265         dim=dim,
   (...)
    273         create_index_for_new_dim=create_index_for_new_dim,
    274     )
    275 elif isinstance(first_obj, Dataset):
--> 276     return _dataset_concat(
    277         objs,
    278         dim=dim,
    279         data_vars=data_vars,
    280         coords=coords,
    281         compat=compat,
    282         positions=positions,
    283         fill_value=fill_value,
    284         join=join,
    285         combine_attrs=combine_attrs,
    286         create_index_for_new_dim=create_index_for_new_dim,
    287     )
    288 else:
    289     raise TypeError(
    290         "can only concatenate xarray Dataset and DataArray "
    291         f"objects, got {type(first_obj)}"
    292     )

File ~/Documents/Work/Code/xarray/xarray/core/concat.py:538, in _dataset_concat(datasets, dim, data_vars, coords, compat, positions, fill_value, join, combine_attrs, create_index_for_new_dim)
    532     datasets = [
    533         ds.expand_dims(dim, create_index_for_new_dim=create_index_for_new_dim)
    534         for ds in datasets
    535     ]
    537 # determine which variables to concatenate
--> 538 concat_over, equals, concat_dim_lengths = _calc_concat_over(
    539     datasets, dim, dim_names, data_vars, coords, compat
    540 )
    542 # determine which variables to merge, and then merge them according to compat
    543 variables_to_merge = (coord_names | data_names) - concat_over

[...]

PR #152, opened 2024-06-21
Rename paths in manifest

### LINKED ISSUE #130: Rewrite paths in a manifest
One minor feature that might be useful would be a convenience method for rewriting the string paths in a manifest. The use case is if you move or rename underlying files, you don't want to have to regenerate all the byte ranges when you could just use the same ones and edit the paths.

This is therefore related to #118, as you could open previously-written kerchunk references, change the paths, and then re-save them, without having to find byte ranges again.

I'm imagining adding API something like this:

```python
class Manifest:
    ...

    def rename_paths(
        new: str | Callable[str, str],
    ) -> Manifest:
        """
        Rename paths to chunks in this manifest.

        Accepts either a string, in which case this new path will be used for all chunks, or 
        a function which accepts the old path and returns the new path.

        Parameters
        ----------
        new
            New path to use for all chunks, either as a string, or as a function which accepts and returns strings.

        Returns
        -------
        manifest

        Examples
        --------
        Rename paths to reflect moving the referenced files from local storage to an S3 bucket.

        >>> def local_to_s3_url(old_local_path: str) -> str:
        ...     from pathlib import Path
        ...
        ...     new_s3_bucket_url = "http://s3.amazonaws.com/my_bucket/"
        ...
        ...     filename = Path(old_local_path).name
        ...     return str(new_s3_bucket_url / filename)

        >>> manifest.rename_paths(local_to_s3_url)
        """
        ...
```

This method would be implemented on `Manifest`, but also present on `ManifestArray` and on the [`VirtualiZarrDatasetAccessor`](https://github.com/TomNicholas/VirtualiZarr/blob/1aac3de9d9cff386b9a00158a431f62c1f81dc14/virtualizarr/xarray.py#L339).

The option to set all chunks to have the same path might not be particularly useful, though perhaps more so if we support indexing (https://github.com/TomNicholas/VirtualiZarr/issues/51).

TomNicholas:
- [x] Closes #130
- [x] Tests added

### DIFF
--- a/virtualizarr/xarray.py
+++ b/virtualizarr/xarray.py
@@ -479,13 +479,10 @@ class VirtualiZarrDatasetAccessor:
         ChunkManifest.rename_paths
         """
 
-        def _rename_manifestarray(marr: ManifestArray) -> ManifestArray:
-            return marr.rename_paths(new=new)
+        new_ds = self.ds.copy()
+        for var_name in new_ds.variables:
+            data = new_ds[var_name].data
+            if isinstance(data, ManifestArray):
+                new_ds[var_name].data = data.rename_paths(new=new)
 
-        def _rename_dataarray(da: xr.DataArray) -> xr.DataArray:
-            if isinstance(da.data, ManifestArray):
-                return xr.apply_ufunc(_rename_manifestarray, da)
-            else:
-                return da
-
-        return self.ds.map(_rename_dataarray)
+        return new_ds
--- a/virtualizarr/xarray.py
+++ b/virtualizarr/xarray.py
@@ -1,6 +1,7 @@
 from collections.abc import Iterable, Mapping, MutableMapping
 from pathlib import Path
 from typing import (
+    Callable,
     Literal,
     Optional,
     overload,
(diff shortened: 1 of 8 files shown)


virtualizarr/xarray.py
```diff
@@ -438,3 +439,53 @@ def to_kerchunk(
...
+        def _rename_manifestarray(marr: ManifestArray) -> ManifestArray:
+            return marr.rename_paths(new=new)
+
+        def _rename_dataarray(da: xr.DataArray) -> xr.DataArray:
+            if isinstance(da.data, ManifestArray):
+                return xr.apply_ufunc(_rename_manifestarray, da)
+            else:
+                return da
+
+        return self.ds.map(_rename_dataarray)
```
TomNicholas, line 491: Need to change this because `xr.apply_ufunc` craps out when it sees that the `ManifestArrays` have .chunks. Same problem really as #114, but we can just work around it here.

TomNicholas: Got around this in https://github.com/zarr-developers/VirtualiZarr/pull/152/commits/e099e9891509cb6d7b39dfc5a4f4d787c8f861d5

PR #154, opened 2024-06-23
Ensure attributes on coordinate variables are preserved during round-tripping

### LINKED ISSUE #155: Bugs with preserving variable-level attrs
There are some bugs with preserving variable-level attributes.

You can see this in #154, where [changing `assert_equal` to `assert_identical`](https://github.com/zarr-developers/VirtualiZarr/pull/154/commits/8ddac726a26294c195ba3cc53aa306fd0d857f0d) in the roundtripping tests in `test_integration.py` reveals that the attributes of the `lat` and `lon` coordinate variables [are being dropped](https://github.com/zarr-developers/VirtualiZarr/actions/runs/9635897427/job/26573199921?pr=154) accidentally.

TomNicholas:
- [x] Closes #155
- [x] Tests added
- [x] Tests passing

### DIFF
--- a/virtualizarr/xarray.py
+++ b/virtualizarr/xarray.py
@@ -243,7 +243,6 @@ def virtual_vars_from_kerchunk_refs(
         var_name: variable_from_kerchunk_refs(refs, var_name, virtual_array_class)
         for var_name in var_names_to_keep
     }
-
     return vars
 
 
@@ -306,7 +305,7 @@ def separate_coords(
     """
     Try to generate a set of coordinates that won't cause xarray to automatically build a pandas.Index for the 1D coordinates.
 
-    Currently requires a workaround unless xarray 8107 is merged.
+    Currently requires this function as a workaround unless xarray PR #8124 is merged.
 
     Will also preserve any loaded variables and indexes it is passed.
     """
@@ -322,7 +321,7 @@ def separate_coords(
             # use workaround to avoid creating IndexVariables described here https://github.com/pydata/xarray/pull/8107#discussion_r1311214263
             if len(var.dims) == 1:
                 dim1d, *_ = var.dims
-                coord_vars[name] = (dim1d, var.data)
+                coord_vars[name] = (dim1d, var.data, var.attrs)
 
                 if isinstance(var, IndexVariable):
                     # unless variable actually already is a loaded IndexVariable,
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -25,7 +25,9 @@ Deprecations
 Bug fixes
 ~~~~~~~~~
 
-- Ensure that `_ARRAY_DIMENSIONS` are dropped from variable `.attrs`. (:issue:`150`, :pull:`152`) 
+- Ensure that `_ARRAY_DIMENSIONS` are dropped from variable `.attrs`. (:issue:`150`, :pull:`152`)
+  By `Tom Nicholas <https://github.com/TomNicholas>`_.
+- Ensure that `.attrs` on coordinate variables are preserved during round-tripping. (:issue:`155`, :pull:`154`)
   By `Tom Nicholas <https://github.com/TomNicholas>`_.
 
 Documentation
--- a/virtualizarr/tests/test_integration.py
+++ b/virtualizarr/tests/test_integration.py
@@ -62,7 +62,7 @@ class TestKerchunkRoundtrip:
         roundtrip = xr.open_dataset(f"{tmpdir}/refs.{format}", engine="kerchunk")
 
         # assert equal to original dataset
-        xrt.assert_equal(roundtrip, ds)
+        xrt.assert_identical(roundtrip, ds)
 
     def test_kerchunk_roundtrip_concat(self, tmpdir, format):
         # set up example xarray dataset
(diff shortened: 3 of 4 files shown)


virtualizarr/xarray.py
```diff
@@ -322,7 +321,7 @@ def separate_coords(
             # use workaround to avoid creating IndexVariables described here https://github.com/pydata/xarray/pull/8107#discussion_r1311214263
             if len(var.dims) == 1:
                 dim1d, *_ = var.dims
-                coord_vars[name] = (dim1d, var.data)
+                coord_vars[name] = (dim1d, var.data, var.attrs)
```
TomNicholas, line 324: Adding this fixes the initial issue described in #155 (i.e. the `lat` and `lon` coordinate variables having their `attrs` get dropped) and makes the new test in this PR pass.

However it seems to introduce a new failure involving decoding the `time` variable:

```python
>       xrt.assert_identical(roundtrip, ds)
E       AssertionError: Left and right Dataset objects are not identical
E       
E       Differing coordinates:
E       L * time     (time) datetime64[ns] 23kB 2013-01-01T00:02:06.757437440 ... 201...
E       R * time     (time) float32 12kB 1.867e+06 1.867e+06 ... 1.885e+06 1.885e+06
E           Differing variable attributes:
E               units: hours since 1800-01-01
E               calendar: standard
```

@jsignell wondering if you have any idea what's going on here?

TomNicholas: If I set `decode_times=True` then I get

```python
>       xrt.assert_identical(roundtrip, ds)
E       AssertionError: Left and right Dataset objects are not identical
E       
E       Differing coordinates:
E       L * time     (time) datetime64[ns] 23kB 2013-01-01T00:02:06.757437440 ... 201...
E       R * time     (time) datetime64[ns] 23kB 2013-01-01T00:02:06.757437440 ... 201...
```
which seem extremely close, similar to in https://github.com/zarr-developers/VirtualiZarr/pull/122#discussion_r1649152320.

TomNicholas, line 324: This gets passed into the constructor of [`xr.Coordinates`](https://docs.xarray.dev/en/latest/generated/xarray.Coordinates.html), which calls [`xr.as_variable`](https://docs.xarray.dev/en/latest/generated/xarray.as_variable.html).

jsignell: Just to make sure I am understanding this properly:
- if you don't set `decode_times` then the roundtripped version ends up as a datetime[ns] and the original version ends up with integers
- if you _do_ set `decode_times=True` then the both the roundtripped and the original end up as datetime[ns] but with a floating point error

jsignell: Ok I just tested a little theory. I think this has to do with the default attrs that xarray adds to the ds on load. When you read the saved version back in using `xarray.open_dataset` you end up with datetime[ns] dtype as well:

```diff
diff --git a/virtualizarr/tests/test_integration.py b/virtualizarr/tests/test_integration.py
index b758be6..0b370d5 100644
--- a/virtualizarr/tests/test_integration.py
+++ b/virtualizarr/tests/test_integration.py
@@ -52,6 +52,9 @@ class TestKerchunkRoundtrip:
         # save it to disk as netCDF (in temporary directory)
         ds.to_netcdf(f"{tmpdir}/air.nc")
 
+        # read it back in
+        original = xr.open_dataset(f"{tmpdir}/air.nc")
+
         # use open_dataset_via_kerchunk to read it as references
         vds = open_virtual_dataset(f"{tmpdir}/air.nc", indexes={})
 
@@ -62,12 +65,18 @@ class TestKerchunkRoundtrip:
         roundtrip = xr.open_dataset(f"{tmpdir}/refs.{format}", engine="kerchunk")
 
         # assert equal to original dataset
-        xrt.assert_identical(roundtrip, ds)
+        xrt.assert_identical(roundtrip, original)
 
     def test_kerchunk_roundtrip_concat(self, tmpdir, format):
         # set up example xarray dataset
         ds = xr.tutorial.open_dataset("air_temperature", decode_times=False)
 
+        # save it to disk as netCDF (in temporary directory)
+        ds.to_netcdf(f"{tmpdir}/air.nc")
+
+        # read it back in
+        original = xr.open_dataset(f"{tmpdir}/air.nc")
+
         # split into two datasets
         ds1, ds2 = ds.isel(time=slice(None, 1460)), ds.isel(time=slice(1460, None))
 
@@ -89,7 +98,7 @@ class TestKerchunkRoundtrip:
         roundtrip = xr.open_dataset(f"{tmpdir}/refs.{format}", engine="kerchunk")
 
         # assert equal to original dataset
-        xrt.assert_identical(roundtrip, ds)
+        xrt.assert_identical(roundtrip, original)
```

TomNicholas: So I just realised that if I make sure to set `decode_times=False` in both the original 

```python
ds = xr.tutorial.open_dataset("air_temperature", decode_times=False)
```
and the roundtrip 
```python
# use fsspec to read the dataset from disk via the zarr store
roundtrip = xr.open_dataset(f"{tmpdir}/refs.{format}", engine="kerchunk", decode_times=False)
```
calls then all the tests pass on this PR, even with `assert_identical`. So perhaps this is a red herring and what we should do is merge this PR with `decode_times=False` then rebase #112 and work on `decode_times=True` there?

jsignell: yep that makes sense to me. It seems like xarray might determine whether or not to decode times based on the attrs that are defined and since some attrs get added on load it might bump the version that you store in netcdf over the boundary.

PR #156, opened 2024-06-24
Identify non dimension coords

### LINKED ISSUE #105: Identifying coordinate variables
I'm a little unclear how xarray / netCDF coordinate variables are distinguished from data variables in the Zarr model.

Xarray follows the netCDF convention that 1-dimensional variables with the same name as their only dimension are to be treated as "coordinates", but it is also possible to have additional 1D or multi-dimensional coordinates (which [netCDF calls "auxiliary coordinate variables"](https://www.unidata.ucar.edu/software/netcdf/workshops/2010/datamodels/NcCVars.html)).

I think at the very least we might need an additional line of code inside `open_virtual_dataset` that looks for any metadata attribute named `'coordinates'`.

TomNicholas:
- [x] Closes #105
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -29,6 +29,8 @@ Bug fixes
   By `Tom Nicholas <https://github.com/TomNicholas>`_.
 - Ensure that `.attrs` on coordinate variables are preserved during round-tripping. (:issue:`155`, :pull:`154`)
   By `Tom Nicholas <https://github.com/TomNicholas>`_.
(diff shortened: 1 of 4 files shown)


TomNicholas:
The implementation here could potentially be replaced by one based on `xr.decode_cf` (see #157), but the tests added here will still be useful to define the intended behaviour.

Issue #94, opened 2024-04-25
Is the `compressor` type right? 


jbusecke:
As part of #93 I just ran into an issue with the type of `filters` on `zarr.Codec` and `zarr.Zarray` which I then noticed was already fixed on main by #66. 

I was wondering though if the type of compressor also needs to be changed from str to [Codec](https://numcodecs.readthedocs.io/en/latest/abc.html#numcodecs.abc.Codec) as indicated in the zarr [API docs](https://zarr.readthedocs.io/en/stable/api/creation.html)?

<img width="534" alt="image" src="https://github.com/TomNicholas/VirtualiZarr/assets/14314623/6de32f42-50f6-40cc-aff9-dc8ff69bf017">

TomNicholas:
I think this is zarr v2 vs v3 difference. In v2 there were separate filters and compressors, but in v3 those are both examples of codecs. Internally VirtualiZarr currently uses the word "codec" to refer to things that must be the same between arrays in order for them to be concatenatable, so the `Codec` class I defined in `virtualizarr/zarr.py` is just a way to keep the filter and compressor together.

Ideally all these differences would be handled by a compatibility class that I just import from zarr-python.

maxrjones:
Following https://github.com/zarr-developers/VirtualiZarr/pull/482, we use `zarr.core.metadata.v3.ArrayV3Metadata` and a conversion utility for V2 metadata, so we no longer need to define these types independently.

Issue #17, opened 2024-03-08
Zarr V2 vs V3


TomNicholas:
What needs to be done here?

TomNicholas:
Ideally we would be able to import Zarr objects from zarr-python that help us navigate the differences between v2 and v3

TomNicholas:
So Zarr v3 was released last week ([it's on PyPI](https://pypi.org/project/zarr/3.0.0a0/), and can be installed by specifying --pre with pip install). This means we can start the process of refactoring virtualizarr to use the v3 zarr-python library for some internal abstractions.

In particular I want to replace the `virtualizarr.zarr.ZArray` class. I haven't really looked at the v3 code much at all yet, but I'm hoping there is some drop-in for the `ZArray` class that can smoothly handle the difference between Zarr v2 and v3 arrays for us. (And if there's not we should be making suggestions upstream.)

maxrjones:
The migration to Zarr-Python 3.0 is being discussed in more detail in #392

Issue #5, opened 2024-03-08
Virtual concatenation of arrays with different codecs or dtypes


TomNicholas:
The motivation for this is in https://github.com/zarr-developers/zarr-specs/issues/288. A `VirtualZarrArray` class can be prototyped in this library, but can't be serialized in a way that can be read by any Zarr readers until that upstream work is done.

Eventually we probably want to change the default concatenation behaviour to concatenate using `VirtualZarrArray`s instead of `ManifestArray`s.

TomNicholas:
Currently `np.concat` called on two `ManifestArray` objects with different codecs will raise an error here:

https://github.com/TomNicholas/VirtualiZarr/blob/f226093bcb8ad248b7a2c8cdcadd224747089792/virtualizarr/manifests/array_api.py#L54

TomNicholas:
Virtual concatenation would also allow for concatenation of arrays with different dtypes, see #215 for an example.

TomNicholas:
Note the proposed approach in https://github.com/zarr-developers/zarr-python/issues/2536

TomNicholas:
There's an important potential use case for chunk-dependent codecs: storing LLM model weights in Zarr. Those often have "quantization" (basically bit packing, or scale and offset-like encoding), but sometimes it's on a per-chunk basis.

PR #175, opened 2024-07-03
Replace VirtualiZarr.ZArray with zarr ArrayMetadata

ghidalgo3:
Before going further with these changes, I'd like to discuss if this direction is correct or not. Based on #17, I tried using Zarr-V3's [Array](https://github.com/zarr-developers/zarr-python/blob/ace96f569490882e89c181387d27f20b0babd6a1/src/zarr/array.py#L591C1-L591C13) definition instead of VirtualiZarr's `ZArray` definition. Keeping the names straight is hard, `ZArray` is the current code and `Array` is `zarr`'s type.

The problems I ran into were:

1. The set of allowed `kwargs` for `Array` are different between v2 and v3 for zarr, but `VirtualiZarr.ZArray` I believe is first created with the v2 properties and then transformed to v3 format when `zarr_v3_array_metadata` is called. After the replacement, anywhere VirtualiZarr creates a `Array`, one must choose v2 or v3 `kwargs`, or again introduce something that abstracts over the versions. Unless VirtualiZarr chooses to only support Zarr V3 only, which I believe is the intention.
1. During concatenation, VirtualiZarr checks if the codecs between ZArrays all match, which was easier when ZArray declared the codecs. Now, if you support v2 and v3 then the code checking logic has to branch based on the type of `zarr.array.metadata` because the internal zarr property names have changed. Probably there will be other code-level incompatibilities from metadata property name differences, I think #94 ran into this too.

This is just from a few hours of trying this out.

If VirtualiZarr needs to support both Zarr V2 and V3 stores, then this change needs to handle all the little differences in Zarr's API between 2 and 3. But if only V3 is supported, this becomes easier and replacing `ZArray` with `Array` should not be too difficult.

Thoughts?

Incompatibilities:
- [x] https://github.com/zarr-developers/zarr-python/issues/2009
- [ ] https://github.com/zarr-developers/zarr-python/issues/2021
 
To be filled later:
- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/array.py
+++ b/virtualizarr/manifests/array.py
@@ -49,6 +49,9 @@ class ManifestArray:
                 else:
                     _zarray = ArrayV2Metadata(**zarray)
 
+        if not isinstance(_zarray.chunk_grid, RegularChunkGrid):
+            raise ValueError("Chunk grid must be RegularChunkGrid.")
+
         if isinstance(chunkmanifest, ChunkManifest):
             _chunkmanifest = chunkmanifest
         elif isinstance(chunkmanifest, dict):
@@ -60,7 +63,6 @@ class ManifestArray:
 
         # TODO check that the zarray shape and chunkmanifest shape are consistent with one another
         # TODO also cover the special case of scalar arrays
-
         self._zarray = _zarray
         self._manifest = _chunkmanifest
 
@@ -92,12 +94,7 @@ class ManifestArray:
         """
         Individual chunk size by number of elements.
         """
-        if isinstance(self._zarray.chunk_grid, RegularChunkGrid):
-            return self._zarray.chunk_grid.chunk_shape
-        else:
-            raise NotImplementedError(
-                "Only RegularChunkGrid is currently supported for chunk size"
-            )
+        return self._zarray.chunk_grid.chunk_shape
 
     @property
     def dtype(self) -> np.dtype:
--- a/virtualizarr/manifests/array.py
+++ b/virtualizarr/manifests/array.py
@@ -2,9 +2,9 @@ import warnings
 from typing import Any, Callable, Union
 
 import numpy as np
-from zarr.array import Array as ZArray
 
 from ..kerchunk import KerchunkArrRefs
+from ..zarr import ZArray
 from .array_api import MANIFESTARRAY_HANDLED_ARRAY_FUNCTIONS
 from .manifest import ChunkManifest
 
--- a/virtualizarr/manifests/array.py
+++ b/virtualizarr/manifests/array.py
@@ -2,9 +2,9 @@ import warnings
 from typing import Any, Callable, Union
 
 import numpy as np
+from zarr.array import ArrayMetadata, ArrayV2Metadata, ArrayV3Metadata, RegularChunkGrid
 
 from ..kerchunk import KerchunkArrRefs
-from ..zarr import ZArray
 from .array_api import MANIFESTARRAY_HANDLED_ARRAY_FUNCTIONS
 from .manifest import ChunkManifest
 
@@ -22,11 +22,11 @@ class ManifestArray:
     """
 
     _manifest: ChunkManifest
-    _zarray: ZArray
+    _zarray: ArrayMetadata
 
     def __init__(
         self,
-        zarray: ZArray | dict,
+        zarray: ArrayMetadata | dict,
         chunkmanifest: dict | ChunkManifest,
     ) -> None:
         """
@@ -34,15 +34,20 @@ class ManifestArray:
 
         Parameters
         ----------
-        zarray : dict or ZArray
+        zarray : dict or zarr.array.ArrayMetadata
         chunkmanifest : dict or ChunkManifest
         """
 
-        if isinstance(zarray, ZArray):
-            _zarray = zarray
-        else:
-            # try unpacking the dict
-            _zarray = ZArray(**zarray)
+        match zarray:
+            case ArrayMetadata():
+                _zarray = zarray
+            case dict():
+                zarray = zarray.copy()
+                zarr_format = zarray.pop("zarr_format", None)
+                if zarr_format == 3:
+                    _zarray = ArrayV3Metadata(**zarray)
+                else:
+                    _zarray = ArrayV2Metadata(**zarray)
 
         if isinstance(chunkmanifest, ChunkManifest):
             _chunkmanifest = chunkmanifest
@@ -79,12 +84,20 @@ class ManifestArray:
         return self._manifest
 
     @property
-    def zarray(self) -> ZArray:
+    def zarray(self) -> ArrayMetadata:
         return self._zarray
 
     @property
     def chunks(self) -> tuple[int, ...]:
-        return tuple(self.zarray.chunks)
+        """
+        Individual chunk size by number of elements.
+        """
+        if isinstance(self._zarray.chunk_grid, RegularChunkGrid):
+            return self._zarray.chunk_grid.chunk_shape
+        else:
+            raise NotImplementedError(
+                "Only RegularChunkGrid is currently supported for chunk size"
+            )
 
     @property
     def dtype(self) -> np.dtype:
@@ -93,6 +106,9 @@ class ManifestArray:
 
     @property
     def shape(self) -> tuple[int, ...]:
+        """
+        Array shape by number of elements along each dimension.
+        """
         return tuple(int(length) for length in list(self.zarray.shape))
 
     @property
--- a/virtualizarr/manifests/array.py
+++ b/virtualizarr/manifests/array.py
@@ -2,9 +2,9 @@ import warnings
 from typing import Any, Callable, Union
 
 import numpy as np
+from zarr.array import Array as ZArray
 
 from ..kerchunk import KerchunkArrRefs
-from ..zarr import ZArray
 from .array_api import MANIFESTARRAY_HANDLED_ARRAY_FUNCTIONS
 from .manifest import ChunkManifest
 
--- a/virtualizarr/manifests/array_api.py
+++ b/virtualizarr/manifests/array_api.py
@@ -1,6 +1,9 @@
-from typing import TYPE_CHECKING, Callable, Iterable
+from dataclasses import replace
+from typing import TYPE_CHECKING, Callable, Iterable, Union
 
 import numpy as np
+from zarr.abc.codec import Codec as ZCodec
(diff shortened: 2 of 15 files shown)


rabernat:
Just wanted to say that the Zarr V3 class behaviors are not set in stone yet. If you have feedback on the V3 API and potential incompatibilities between V2, please report them issues on Zarr-python: https://github.com/zarr-developers/zarr-python/issues

jsignell:
My sense is that in terms of file format both zarr v2 and zarr v3 should be supported but in terms of library dependencies it would be fine to mandate that only zarr-python >= 3 is supported.

TomNicholas:
Thanks for taking the initiative here @ghidalgo3 !

> I tried using Zarr-V3's [Array](https://github.com/zarr-developers/zarr-python/blob/ace96f569490882e89c181387d27f20b0babd6a1/src/zarr/array.py#L591C1-L591C13) definition instead of VirtualiZarr's ZArray definition

I was actually hoping we could use some class from `zarr-python` that wasn't a fully-featured `zarr.Array`, instead using something that just represented the metadata for one array. This [`ArrayV3Metadata` class](https://github.com/zarr-developers/zarr-python/blob/22e3fc5070f21c8131f3293723dc1cda85db8665/src/zarr/metadata.py#L160) might be what we need? Is that public API?

> If VirtualiZarr needs to support both Zarr V2 and V3 stores, then this change needs to handle all the little differences in Zarr's API between 2 and 3. But if only V3 is supported, this becomes easier and replacing `ZArray` with `Array` should not be too difficult.

I see two ways to do this:
1. Make VirtualiZarr internally agnostic to v2 vs v3. Then a `ManifestArray` can represent either a v2 array or a v3 array, distinguished via its metadata. For example the attributes of a `ManifestArray` might become
    ```python
    from zarr.metadata import ArrayMetadata

    class ManifestArray:
        _chunkmanifest: ChunkManifest
        _zarray: ArrayMetadata
    ```
    where [ArrayMetadata](https://github.com/zarr-developers/zarr-python/blob/22e3fc5070f21c8131f3293723dc1cda85db8665/src/zarr/metadata.py#L119) is an ABC that needs to be a concrete `ArrayV2Metadata` or `ArrayV3Metadata`.
3. VirtualiZarr internally only uses v3 arrays, and whilst we support reading zarr and writing zarr v2 arrays, they are coerced at reading / writing time.

I'm not sure which of these two is the best approach.

I also completely agree with @abarciauskas-bgse - all of this should be supported through only a dependency on `zarr>=3.0.0`, as this PR adds.

ghidalgo3:
To minimize the code changes needed in VirtualiZarr, I'll attempt the first option.

ghidalgo3:
Adding a dependency on `zarr==3` causes a problem for Kerchunk. Kerchunk expects to be able to `import zarr.meta` but in ZarrV3, that code was moved to the `v2` directory such that Kerchunk has to do `from zarr.v2.meta import encode_fill_value` instead of `from zarr.meta import encode_fill_value`. Then this causes unit tests to fail in VirtualiZarr.

See the reorg that happened here about 3 months ago for ZarrV3: https://github.com/zarr-developers/zarr-python/pull/1809 and some attempts in Kerchunk to support ZarrV3:

1. https://github.com/fsspec/kerchunk/issues/235
2. https://github.com/fsspec/kerchunk/pull/292

Maybe the `v2` module name is temporary in Zarr? But until then, any tests that call into Kerchunk are broken :(.

rabernat:
@ghidalgo3 please please open a Zarr issue to track any API incompatibilities between V2 and V3. 🙏 

https://github.com/zarr-developers/zarr-python/issues

TomNicholas:
@ghidalgo3 it might be a good idea to have a separate first PR that makes virtualizarr pass every test using zarr v3, and then coming back to this PR once that compatibility can be relied upon. Otherwise this might become a rabbit hole.

d-v-b:
> I was actually hoping we could use some class from `zarr-python` that wasn't a fully-featured `zarr.Array`, instead using something that just represented the metadata for one array. This [`ArrayV3Metadata` class](https://github.com/zarr-developers/zarr-python/blob/22e3fc5070f21c8131f3293723dc1cda85db8665/src/zarr/metadata.py#L160) might be what we need? Is that public API?

That class is public API, but it might be changing soon, and while it is intended for this use case it also has some extra stuff that goes beyond "representing metadata". So it should do what you want, eventually :)

virtualizarr/manifests/array.py
```diff
@@ -79,12 +84,20 @@ def manifest(self) -> ChunkManifest:
...
+    def zarray(self) -> ArrayMetadata:
         return self._zarray
 
     @property
     def chunks(self) -> tuple[int, ...]:
-        return tuple(self.zarray.chunks)
+        """
+        Individual chunk size by number of elements.
+        """
+        if isinstance(self._zarray.chunk_grid, RegularChunkGrid):
```
TomNicholas, line 95: this is a nice check, but I think we could actually perform this check at `ManifestArray` construction time, because right now a lot of other things will break if the chunk grid is not regular.

virtualizarr/manifests/array_api.py
```diff
@@ -61,7 +54,20 @@ def _check_same_dtypes(dtypes: list[np.dtype]) -> None:
...
+        raise ValueError("Cannot concatenate arrays with different zarr formats.")
+
+    def to_codec(zarray: ArrayMetadata) -> Union[Codec | tuple[ZCodec, ...]]:
+        match zarray:
+            case ArrayV2Metadata(compressor=compressor, filters=filters):
+                return Codec(compressor=compressor, filters=filters)
+            case ArrayV3Metadata(codecs=codecs):
+                return codecs
+            case _:
+                raise ValueError("Unknown ArrayMetadata type")
```
TomNicholas, line 68: Is there a reason to prefer this `match...case` over a standard `if isinstance()...else` syntax?

virtualizarr/tests/__init__.py
```diff
@@ -46,15 +47,14 @@ def create_manifestarray(
     The manifest is populated with a (somewhat) unique path, offset, and length for each key.
     """
 
-    zarray = ZArray(
+    zarray = ArrayV2Metadata(
         chunks=chunks,
-        compressor="zlib",
+        compressor={"id": "zlib"},
```
TomNicholas, line 52: Is the presence of `"id"` a v2 vs v3 difference? because you don't seem to use it in the other tests

virtualizarr/zarr.py
```diff
@@ -25,90 +25,50 @@
...
+
+    compressor: str | dict[str, Any] | None = None
+    """
+    If it's a string, it's the compressor ID.
+    If it's a dict, it's the full compressor configuration.
+    """
     filters: list[dict] | None = None
 
     def __repr__(self) -> str:
         return f"Codec(compressor={self.compressor}, filters={self.filters})"
```
TomNicholas, line 40: Should this class be moved upstream?

TomNicholas (commented):
Great work so far @ghidalgo3 !

virtualizarr/manifests/array_api.py
ghidalgo3: Mainly a style preference coming from languages with discriminated unions and compilers, but if you'd like the whole code base to be consistent I can rewrite it with `isinstance`.

virtualizarr/tests/__init__.py
ghidalgo3: In V3, the `codecs` array is made up of `(name, configuration)` objects, but in V2 the `id` field serves as the name. In this particular test, I just chose to convert the `ZArray` to an `ArrayV2Metadata`

Actually please ignore any of the changes in `tests` for now because I will need to edit almost every test to excise `ZArray` :) The few changes I added were done up until I hit the `kerchunk` module import issue

virtualizarr/zarr.py
ghidalgo3: I'm not sure, because ZarrV3 now has a notion of [codecs](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#id12) which is different from VirtualiZarr's notion of codecs. A VirtualiZarr codec is really the combination of `filters` and `compressor` from V2 metadata, which are easily transformable to V3 codecs.

Maybe a `zarr.ArrayV2Metadata` can have a property that projects the `compressor` and `filters` into V3 `Codec`s, @d-v-b thoughts on that? In that case, `Codec` could even be a property of the ABC `ArrayMetadata`.

d-v-b: I actually think that `codecs: ListWithInternalStructure[Codec]` (i.e., what zarr v3 metadata uses) is not the best interface. See a writeup on that point of view [here](https://github.com/zarr-developers/zarr-specs/issues/298). 

But if we ignore that concern for a moment, I'm curious about @TomNicholas's question: what could we change in zarr-python so that you could just import a functionally equivalent class from there (or import a dataclass and wrap with pydantic)? I have a selfish interest in this question since I'm likely going to follow in your footsteps when I implement better codec support over in [`pydantic-zarr`](https://github.com/janelia-cellmap/pydantic-zarr/blob/a1f7a978a5c68b8c3425be7d83202986ef828d12/src/pydantic_zarr/v3.py#L117).

ghidalgo3: Thanks for forcing me to read the Zarr code more closely! I found VirtualiZarr could probably just use [create_codec_pipeline](https://github.com/zarr-developers/zarr-python/blob/b8baa6868c8fa95e6b2948c9fd9c725331ca23ec/src/zarr/array.py#L79C1-L87C29 ) to turn V2 or V3 metadata into a `BatchedCodecPipeline` and then VirtualiZarr can just deal with `CodecPipeline` as the abstraction over V2 and V3 . Additionally, VirtualiZarr could change its `Codec` to use [`V2Filters`](https://github.com/zarr-developers/zarr-python/blob/v3/src/zarr/codecs/_v2.py#L65) and `V2Compressor` from `Zarr`, or I might get rid of it entirely.

d-v-b: what does Virtualizarr do with the codecs?

TomNicholas: All virtualizarr currently needs to do with the codecs is:
1) Check that any two arrays you ask it to concatenate have the same codecs (see https://github.com/zarr-developers/VirtualiZarr/issues/5)
2) Make sure the codecs are recorded in the kerchunk references when they get written out to disk.

Issue #78, opened 2024-04-10
Generating references without kerchunk


TomNicholas:
VirtualiZarr + [zarr chunk manifests](https://github.com/TomNicholas/VirtualiZarr/pull/45) re-implement so much of kerchunk that the only part left is [kerchunk's backends](https://fsspec.github.io/kerchunk/reference.html#file-format-backends) - the part that actually generates the byte ranges from a given legacy file. It's interesting to imagine whether we could make virtualizarr work without using kerchunk or fsspec at all.

https://github.com/TomNicholas/VirtualiZarr/issues/61#issuecomment-2047826810 discusses how the rust `object-store` crate might allow us to read actual bytes from zarr v3 stores with chunk manifests over S3, without using fsspec.

The other place we use fsspec (+ kerchunk) is to generate the references in the first place. But can we imagine alternative implementations for generating that byte range information?

Arguments for doing this without using kerchunk + fsspec are essentially:
- **increased reliability**
- **clearer interfaces**
- **not using an overly complex tool** (i.e. fsspec, which can read from all sorts of systems) to read from just two places (local or S3)
- **possible performance increases** during reference generation (though this is unlikely to be a major bottleneck)
- **separation of concerns** - if we can find other libraries that generate the byte range information already for their own purposes (e.g. [h5py]() using the [ros3 driver](https://docs.h5py.org/en/stable/high/file.html#file-drivers), [hidefix](https://github.com/gauteh/hidefix), or [cog3pio](https://github.com/weiji14/cog3pio)), we might be able to avoid bearing that maintainance burden, which would be great.

TomNicholas:
From @sharkinsspatial:

> I just had a quick look at using `hidefix` to support `ChunkManifest` generation without `kerchunk`. Digging in a bit more detail, IIUC `hidefix` is still using the core `HDF5` library to generate an `Index` object which can then be used to bypass the `HDF5` concurrency limitations.
> As we are only iterating over the chunk offsets and not actually reading any data, this doesn’t provide us any advantage over `h5py`.  Given that, I’ll try to focus on just creating an `h5py` based `ChunkManifest` lib for the near term.

sharkinsspatial:
[...]

PR #182, opened 2024-07-10
Add Zarr v3 dependency

### LINKED ISSUE #17: Zarr V2 vs V3

ghidalgo3:
This PR is a pre-cursor to https://github.com/zarr-developers/VirtualiZarr/pull/175 , and will be blocked on at least Kerchunk supporting ZarrV3. 

TODOs:
- [x] https://github.com/fsspec/kerchunk/pull/475
- [x] https://github.com/zarr-developers/zarr-python/issues/2023

Checklist: 
- [ ] Closes #17 
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


docs/releases.rst
```diff
@@ -12,6 +12,9 @@ New Features
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- Requires Zarr 3.0 (for :pull:`182`).
+  By `Tom Nicholas <https://github.com/ghidalgo3>`_.
```
TomNicholas, line 16: !
```suggestion
  By `Gustavo Hidalgo <https://github.com/ghidalgo3>`_.
```

TomNicholas:
Adding zarr as a dependency would also allow us to un-vendor this small bit of code

https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/vendor/zarr/utils.py

ghidalgo3:
I ran into an issue with a behavior change in the call to `zarr.group`, made by Kerchunk when VirtualiZarr calls `SingleHdf5ToZarr`. Here's the problem:
1. VirtualiZarr (either through unit tests tests or a `open_virtual_dataset` call) will call `SingleHdf5ToZarr` and pass a `out=None` argument.
2. In kerchunk, that `out` is passed to `zarr.group(store=out)`. In ZarrV2, that `None` would get converted to an in-memory store that happens to subclass `MutableMapping`. But in ZarrV3, that gets turned into a `StoreLike` which doesn't implement `MutableMapping`.
3. This eventually causes a problem whenever kerchunk calls `_encode_for_JSON` on that store because ZarrV3 stores are not so trivial to turn into a dictionary, there's an async enumeration that needs to happen.

@TomNicholas maybe it might be less work globally if VirtualiZarr didn't depend on Kerchunk? I know there's https://github.com/zarr-developers/VirtualiZarr/issues/78 and https://github.com/zarr-developers/VirtualiZarr/pull/87 as prototypes, maybe they are prerequisites to making VirtualiZarr depend on ZarrV3. 

The process of using Kerchunk to generate ChunkManifests can be delegated to some other package that can continue depending on ZarrV2.

TomNicholas:
@ghidalgo3 that sounds like a gnarly bug!

(It would be nice if python allowed you to have different versions imported for different packages, so that kerchunk could depend on zarr v2 whilst virtualizarr imported zarr v3.)

> maybe it might be less work globally if VirtualiZarr didn't depend on Kerchunk?

@ghidalgo3 many things in this library would be less work if we didn't have to depend on kerchunk, and we're working towards that, but it's not going to be done quickly (especially not for other non-HDF filetypes that kerchunk can also read).

Making kerchunk work with zarr v3 would be the cleanest solution, but it's difficult to know how long that might take too (I know Joe tried it and found it to be a rabbit hole).

> The process of using Kerchunk to generate ChunkManifests can be delegated to some other package that can continue depending on ZarrV2.

I'm not sure I understand this suggestion. Wouldn't that have exactly the same problem as now?

ghidalgo3:
> I'm not sure I understand this suggestion. Wouldn't that have exactly the same problem as now?

Not if you have 2 Python environments:
1. One with Kerchunk and Zarr V2 and it is responsible for reading files and producing `ChunkManifest`s, and you write these to disk or blobs or whatever.
2. One with VirtualiZarr and ZarrV3, which takes the `ChunkManifest`s produced by the first step to create the virtual datasets and ultimately read/write data..

It's an ugly solution yes, but compared to making Kerchunk work with ZarrV3 maybe it's less work? I'm not in a rush so probably makes sense to make Kerchunk work with V3 and V2, and that's only going to happen by upgrading Kerchunk to Zarr>=3.0.0.

> (It would be nice if python allowed you to have different versions imported for different packages, so that kerchunk could depend on zarr v2 whilst virtualizarr imported zarr v3.)

That would be nice! But even before then it would be better if Zarr v2 -> Zarr v3 had easy breaking changes :( This specific one is really subtle, there's no API change to `zarr.group` but an internal change to how `Store`s are handled inside Zarr, which causes a problem for Kerchunk, which causes a problem for VirtualiZarr.

I'll focus my effort on trying to make Kerchunk work with Zarr V3 then.

TomNicholas:
> Not if you have 2 Python environments:

Ohhh right now I get it - you literally split your workflow up into two separate steps using different environments each time. 

> an internal change to how Stores are handled inside Zarr, which causes a problem for Kerchunk

I mean to me that sounds like it's kerchunk's fault for using internal zarr API...

> I'll focus my effort on trying to make Kerchunk work with Zarr V3 then.

Amazing - let us know how that goes (good or bad) so we can re-evaluate here.

virtualizarr/zarr.py
```diff
@@ -12,8 +12,7 @@
 import ujson  # type: ignore
 import xarray as xr
 from pydantic import BaseModel, ConfigDict, field_validator
-
-from virtualizarr.vendor.zarr.utils import json_dumps
+from zarr.v2.util import json_dumps
```
jhamman, line 15: I would avoid this import if possible. We are likely to remove the `v2` namespace before the `3.0` release.

ghidalgo3: I'll revert and continue to vendor this bit of code.

ghidalgo3:
@TomNicholas this work did lead down a pretty deep rabbit hole, probably following in Joe's footsteps. My conclusion is that `VirtualiZarr` should **not** take a code-level dependency on `zarr-python` until all of VirtualiZarr's dependencies explicitly support ZarrV3. `VirtualiZarr` should seek to read/write valid ZarrV3 metadata even if it means duplicating some logic between it and `zarr-python`.

Specifically here is the problem chain this lead me down:
- Adding `zarr==3.0.0a0` to `VirtualiZarr` breaks `kerchunk`, not just because of the module path name changes but also because https://github.com/zarr-developers/zarr-python/issues/2023 means that internally `kerchunk`'s usage of the Store API is broken. 
- Adding a compatibility layer to `kerchunk` to work with both zarrv2 and zarrv3 can be done https://github.com/fsspec/kerchunk/pull/478 _but_ if you actually try to use `kerchunk` with zarrv3 it is broken due to the Store changes in ZarrV3.
- Even if `kerchunk`'s normal usage of `zarr` supported v3, `kerchunk`'s unit tests depend on `xarray.to_zarr` which doesn't work at least because in zarrv3 all the API functions require full keyword arguments, let alone type signature changes. That means `xarray` also needs to support zarrv3.
- I stopped once I discovered that all of `kerchunk`, `fsspec`, and `xarray` all needed changes to work with ZarrV3 and the scope of this change became clear.

Does that make sense? If so, I I'll abandon both this PR and https://github.com/zarr-developers/VirtualiZarr/pull/175

TomNicholas:
That makes complete sense, and you've really done a great job here going down that rabbit hole and clearly communicating all the steps required.

I don't know that we need to close this PR, but it does sound like it will need to be paused for a while before it could be picked up again.

I also agree that in the meantime we can aim for v3 support without a dependency on zarr-python v3

cc @jhamman

TomNicholas:
Coming back to this now as a lot of things have changed, so we might be able to make progress.

> if you actually try to use kerchunk with zarrv3 it is broken

This is still the case.

> xarray also needs to support zarrv3.

This is about to be the case, and there is a branch we can test it on. https://github.com/pydata/xarray/pull/9552

> it might be less work globally if VirtualiZarr didn't depend on Kerchunk

VirtualiZarr now does not depend on Kerchunk! https://github.com/zarr-developers/VirtualiZarr/pull/259 Though if you want to read netCDF files you do have to be able to import kerchunk, the dependency is just optional.

> if you have 2 Python environments:
> 
> 1. One with Kerchunk and Zarr V2 and it is responsible for reading files and producing ChunkManifests, and you write these to disk or blobs or whatever.
> 2. One with VirtualiZarr and ZarrV3, which takes the ChunkManifests produced by the first step to create the virtual datasets and ultimately read/write data..

This is now maybe within reach? The writing to disk could actually be done using the kerchunk format, as virtualizarr can write to and read from the kerchunk json/parquet format without ever importing kerchunk (reading is new, thanks to @norlandrhagen in #251).

At the very least now we should be able to see if new `min-deps` and `upstream` CI workflows (https://github.com/zarr-developers/VirtualiZarr/issues/263) pass.

> we can aim for v3 support without a dependency on zarr-python v3

Adding a reader that can create virtual references from a standard zarr v3 store is tracked in #262.

TomNicholas:
Some more things have now changed:

1) Zarr-python v3 [released](https://github.com/zarr-developers/zarr-python/releases/tag/v3.0.0) a few minutes ago (!)

2) Now that we have a non-kerchunk-based HDF reader it's become much less important that we support kerchunk

3) Nevertheless there is a compatibility branch for kerchunk with zarr v3

maxrjones:
VirtualiZarr now has full support for Zarr-Python 3, so I'm going to close this PR. Please reach out if you run into any issues or have suggestions for improving support. While this PR isn't getting merged, I still really appreciate your work and hope that you'll continue to contribute to VirtualiZarr! Thanks again @ghidalgo3!

PR #186, opened 2024-07-13
Open kerchunk ref as virtual dataset, only json (from PR 119)

kthyng:
This started from https://github.com/zarr-developers/VirtualiZarr/pull/119 and is meant to finish an MVP of that work. All of the work here is essentially from there, which @norlandrhagen did!

@TomNicholas merged main into that branch. Most changes here were brought in from main branch to fix a few leftover errors. Also a change so that if a parquet file is input the NotImplemented exception is raised, and a docstring added for the kerchunk filetype.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`
- [ ] New functionality has documentation

PR #119, opened 2024-05-16
Open Kerchunk refs as Virtual Dataset

norlandrhagen:
- [x] Closes https://github.com/TomNicholas/VirtualiZarr/issues/118
- [ ] Tests added
- [ ] User visible changes (including notable bug fixes) are documented in `changelog.md`
- [ ] New functions/methods are listed in `api.rst`

Start of PR to address https://github.com/TomNicholas/VirtualiZarr/issues/118.

Lots of open questions! 
- How should we read `.parquet`files into `KerchunkStoreRefs` to pass into **dataset_from_kerchunk_refs**
- RT'ing json seems to loose `_ARRAY_DIMENSIONS`

Would love some feedback @jsignell!

PR #187, opened 2024-07-15
Extend refspec support to [path] entries (without offset/length)

maresb:
I'm trying to open a Zarr v2. I couldn't find a good obvious way to do this, so my attempt was:

```python
from kerchunk.zarr import ZarrToZarr
from virtualizarr.xarray import dataset_from_kerchunk_refs

ref = ZarrToZarr(url).translate()
ds = dataset_from_kerchunk_refs(ref)
```

This fails [here in unpacking](https://github.com/zarr-developers/VirtualiZarr/blob/f60707f86f6fa7415463c1a3e886eb134b84bdb0/virtualizarr/manifests/manifest.py#L39) because kerchunk produces reference entries of the form `[path]` instead of `[path, length, offset]`.

I was wondering if I could work around this by `stat`ing the chunks, and sure enough it worked! (See my second commit.)

I'd love to get some feedback on this, including if there's a better way to open v2 zarrs.

In preparation for my second commit I needed to clean up some types. To make things more precise and clean, I tell a minor lie: that the `[path, offset, length]` lists are tuples. This was done in the spirit of duck typing, and I'd argue is more semantically accurate. (Not to mention that tuples serialize to lists in JSON.)

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -299,6 +299,8 @@ class ChunkManifest:
     @classmethod
     def _from_kerchunk_chunk_dict(
         cls,
+        # The type hint requires `Dict` instead of `dict` due to
+        # the conflicting ChunkManifest.dict method.
         kerchunk_chunk_dict: Dict[ChunkKey, str | tuple[str] | tuple[str, int, int]],
     ) -> "ChunkManifest":
         chunk_entries: dict[ChunkKey, ChunkDictEntry] = {}
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -5,6 +5,7 @@ from typing import Any, Callable, NewType, Tuple, TypedDict, cast
 
 import numpy as np
 from pydantic import BaseModel, ConfigDict
+from upath import UPath
 
 from virtualizarr.types import ChunkKey
 
@@ -42,9 +43,14 @@ class ChunkEntry(BaseModel):
 
     @classmethod
     def from_kerchunk(
-        cls, path_and_byte_range_info: tuple[str, int, int]
+        cls, path_and_byte_range_info: tuple[str] | tuple[str, int, int]
     ) -> "ChunkEntry":
-        path, offset, length = path_and_byte_range_info
+        if len(path_and_byte_range_info) == 1:
+            path = path_and_byte_range_info[0]
+            offset = 0
+            length = UPath(path).stat().st_size
+        else:
+            path, offset, length = path_and_byte_range_info
         return ChunkEntry(path=path, offset=offset, length=length)
 
     def to_kerchunk(self) -> tuple[str, int, int]:
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -1,7 +1,7 @@
 import json
 import re
 from collections.abc import Iterable, Iterator
-from typing import Any, Callable, NewType, Tuple, Union, cast
+from typing import Any, Callable, NewType, Tuple, TypedDict, cast
 
 import numpy as np
 from pydantic import BaseModel, ConfigDict
@@ -15,7 +15,13 @@ _SEPARATOR = r"\."
 _CHUNK_KEY = rf"^{_INTEGER}+({_SEPARATOR}{_INTEGER})*$"  # matches 1 integer, optionally followed by more integers each separated by a separator (i.e. a period)
 
 
-ChunkDict = NewType("ChunkDict", dict[ChunkKey, dict[str, Union[str, int]]])
+class ChunkDictEntry(TypedDict):
+    path: str
+    offset: int
+    length: int
+
+
+ChunkDict = NewType("ChunkDict", dict[ChunkKey, ChunkDictEntry])
 
 
 class ChunkEntry(BaseModel):
@@ -35,16 +41,18 @@ class ChunkEntry(BaseModel):
         return f"ChunkEntry(path='{self.path}', offset={self.offset}, length={self.length})"
 
     @classmethod
-    def from_kerchunk(cls, path_and_byte_range_info: list[str | int]) -> "ChunkEntry":
+    def from_kerchunk(
+        cls, path_and_byte_range_info: tuple[str, int, int]
+    ) -> "ChunkEntry":
         path, offset, length = path_and_byte_range_info
         return ChunkEntry(path=path, offset=offset, length=length)
 
-    def to_kerchunk(self) -> list[str | int]:
+    def to_kerchunk(self) -> tuple[str, int, int]:
         """Write out in the format that kerchunk uses for chunk entries."""
-        return [self.path, self.offset, self.length]
+        return (self.path, self.offset, self.length)
 
-    def dict(self) -> dict[str, Union[str, int]]:
-        return dict(path=self.path, offset=self.offset, length=self.length)
+    def dict(self) -> ChunkDictEntry:
+        return ChunkDictEntry(path=self.path, offset=self.offset, length=self.length)
 
 
 class ChunkManifest:
@@ -285,7 +293,7 @@ class ChunkManifest:
     @classmethod
     def _from_kerchunk_chunk_dict(cls, kerchunk_chunk_dict) -> "ChunkManifest":
         chunkentries = {
-            cast(ChunkKey, k): ChunkEntry.from_kerchunk(v).dict()
+            cast(ChunkKey, k): ChunkEntry.from_kerchunk(tuple(v)).dict()
             for k, v in kerchunk_chunk_dict.items()
         }
         return ChunkManifest(entries=cast(ChunkDict, chunkentries))
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -1,7 +1,7 @@
 import json
 import re
 from collections.abc import Iterable, Iterator
-from typing import Any, Callable, NewType, Tuple, TypedDict, cast
+from typing import Any, Callable, Dict, NewType, Tuple, TypedDict, cast
 
 import numpy as np
 from pydantic import BaseModel, ConfigDict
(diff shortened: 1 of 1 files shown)


virtualizarr/manifests/manifest.py
```diff
@@ -35,16 +42,23 @@ def __repr__(self) -> str:
...
     @classmethod
-    def from_kerchunk(cls, path_and_byte_range_info: list[str | int]) -> "ChunkEntry":
-        path, offset, length = path_and_byte_range_info
+    def from_kerchunk(
+        cls, path_and_byte_range_info: tuple[str] | tuple[str, int, int]
+    ) -> "ChunkEntry":
+        if len(path_and_byte_range_info) == 1:
+            path = path_and_byte_range_info[0]
+            offset = 0
+            length = UPath(path).stat().st_size
```
TomNicholas, line 51: This trick might be really useful in general, because it seems kerchunk regularly returns a single string path instead of a tuple here. See for example https://github.com/zarr-developers/VirtualiZarr/pull/186#issuecomment-2229532557

virtualizarr/manifests/manifest.py
```diff
@@ -35,16 +42,23 @@ def __repr__(self) -> str:
         return f"ChunkEntry(path='{self.path}', offset={self.offset}, length={self.length})"
 
     @classmethod
-    def from_kerchunk(cls, path_and_byte_range_info: list[str | int]) -> "ChunkEntry":
-        path, offset, length = path_and_byte_range_info
+    def from_kerchunk(
+        cls, path_and_byte_range_info: tuple[str] | tuple[str, int, int]
+    ) -> "ChunkEntry":
+        if len(path_and_byte_range_info) == 1:
```
TomNicholas, line 48: I think this could be an `isinstance(..., str)` check instead perhaps? Kerchunk's [docs on its specification](https://fsspec.github.io/kerchunk/spec.html) indicate that it should return either a bare `str` or a `tuple[str, int, int]`. We can't predict when kerchunk will randomly decide to return the former, but we should be as explicit as possible when checking to see if that is what has happened. (I'm worried that there may be other situations in which kerchunk returns a length-1 something that isn't actually a string...)

virtualizarr/manifests/manifest.py
```diff
@@ -285,7 +299,7 @@ def to_zarr_json(self, filepath: str) -> None:
     @classmethod
     def _from_kerchunk_chunk_dict(cls, kerchunk_chunk_dict) -> "ChunkManifest":
         chunkentries = {
-            cast(ChunkKey, k): ChunkEntry.from_kerchunk(v).dict()
+            cast(ChunkKey, k): ChunkEntry.from_kerchunk(tuple(v)).dict()
```
TomNicholas, line 302: I _think_ if you don't add this `tuple()` call then my suggestion of using `isinstance(..., str)` above will work.

virtualizarr/manifests/manifest.py
```diff
@@ -15,7 +16,13 @@
...
 
 
-ChunkDict = NewType("ChunkDict", dict[ChunkKey, dict[str, Union[str, int]]])
+class ChunkDictEntry(TypedDict):
+    path: str
+    offset: int
+    length: int
+
+
+ChunkDict = NewType("ChunkDict", dict[ChunkKey, ChunkDictEntry])
```
TomNicholas, line 25: This is nice and probably what I should have used in the first place if I was better with python typing 😅

TomNicholas: I wonder if we should do a similar change for any of the other types in the codebase...

TomNicholas (commented):
> `ref = ZarrToZarr(url).translate()`

When I saw this line I thought your PR was going to add another possible filetype to `open_virtual_dataset` that called `kerchunk.zarr.ZarrToZarr`. My response was going to be that I actually don't want to add any further dependency on kerchunk because we're trying to depend on kerchunk as little as possible (see #87).

But it seems your PR doesn't do that yet, instead it adds a useful little workaround that I didn't know was possible! Amazing.

To support reading Zarr v2 with `open_virtual_dataset` I would prefer that (in a follow-up PR) we wrote code to manually parse the contents of a zarr v2 store ourselves, using the new `ManifestArray.from_array` constructor for maximum efficiency.

virtualizarr/manifests/manifest.py
maresb: Ooh, excellent catch! Kerchunk doesn't like to inline Zarr variable data, effectively masking the case you point out.

I just lazily turned it into a `NotImplementedError` for now since it would require a bit more thought. (Probably best to follow up in a subsequent PR.)

virtualizarr/manifests/manifest.py
maresb: Ya, with my last commit the validation takes care of this.

maresb:
@TomNicholas, I think I'm ready for another round of review, whenever you are.

virtualizarr/manifests/manifest.py
```diff
@@ -283,12 +297,18 @@ def to_zarr_json(self, filepath: str) -> None:
...
     @classmethod
-    def _from_kerchunk_chunk_dict(cls, kerchunk_chunk_dict) -> "ChunkManifest":
-        chunkentries = {
-            cast(ChunkKey, k): ChunkEntry.from_kerchunk(v).dict()
-            for k, v in kerchunk_chunk_dict.items()
-        }
-        return ChunkManifest(entries=cast(ChunkDict, chunkentries))
+    def _from_kerchunk_chunk_dict(
+        cls,
+        kerchunk_chunk_dict: Dict[ChunkKey, str | tuple[str] | tuple[str, int, int]],
```
TomNicholas, line 302:
```suggestion
        kerchunk_chunk_dict: dict[ChunkKey, str | tuple[str] | tuple[str, int, int]],
```
Pretty sure in a recent enough version of python you can just use the built-in `dict` for typing, same as how you're using `tuple`.

virtualizarr/manifests/manifest.py
TomNicholas: Sorry, when does the inline case come up here exactly? I feel like it is also relevant for #119?

virtualizarr/manifests/manifest.py
```diff
@@ -283,12 +297,18 @@ def to_zarr_json(self, filepath: str) -> None:
...
+    ) -> "ChunkManifest":
+        chunk_entries: dict[ChunkKey, ChunkDictEntry] = {}
+        for k, v in kerchunk_chunk_dict.items():
+            if isinstance(v, (str, bytes)):
+                raise NotImplementedError("TODO: handle inlined data")
+            elif not isinstance(v, (tuple, list)):
```
TomNicholas, line 308: Isn't this incorrect typing? You allow `v` to be a `list`, then pass the list into a method which only accepts `tuple`s (i.e. `ChunkEntry.from_kerchunk()`.

TomNicholas:
Sorry for the slow review @maresb ! Just some very small comments here.

virtualizarr/manifests/manifest.py
maresb: I agree that this looks like a mistake, but strangely it's necessary. I added an explanatory comment to prevent others from attempting this obvious change.

virtualizarr/manifests/manifest.py
maresb: I'm pulling some really sketchy tricks with typing here.

My rationale is that kerchunk can and probably will give us weird types, and we should be robust with handling them at runtime. While `ChunkEntry.from_kerchunk` expects `tuple`s it can still handle lists if need be, even though I don't tell the typing system.

From the type checking perspective, the instance check for `list` is vacuous since the type of `v` is `str | tuple[str] | tuple[str, int, int]`. So basically what I'm saying is, "I want to raise a type error if I don't get a `tuple`, but if I happen to find a `list` I'll allow it."

Do you find this an acceptable strategy, or would you like to change it?

Ideally we'd add types to kerchunk, but I don't have time for that right now.

maresb:
> Sorry for the slow review @maresb ! Just some very small comments here.

No worries, no rush, and thanks so much for maintaining!!!

virtualizarr/manifests/manifest.py
maresb: Here's a simple reproducer:

```python
import kerchunk.zarr
import xarray as xr

ds = xr.Dataset({}, coords={"time": [0, 1, 2]})
ds.to_zarr("/tmp/ds.zarr", mode="w")
ref = kerchunk.zarr.ZarrToZarr("/tmp/ds.zarr").translate()
ref["refs"]["time/0"]  # ['file:///tmp/ds.zarr/time/0']
```

I think it's a kerchunk bug that Zarrs are never inlined, even when `inline_threshold` is set.

TomNicholas: > I think it's a kerchunk bug that Zarrs are never inlined, even when inline_threshold is set.

That doesn't surprise me - we already had to deal with https://github.com/fsspec/kerchunk/issues/445.

But we actually don't use kerchunk's inlining feature at all in virtualizarr - instead if we want eager access to some variable we just load it as a numpy array-backed xarray variable instead (
https://virtualizarr.readthedocs.io/en/latest/usage.html#loading-variables). So we just set `inline_threshold=0` whenever we call any kerchunk reader internally.

But I'm a little confused - that's all separate from how kerchunk will sometimes return a `str` instead of `tuple[str, int, int]`. Your example `ref["refs"]["time/0"]  # ['file:///tmp/ds.zarr/time/0']` is not an example of inlining, it's just kerchunk using a different way to implicitly represent the byte range reference to an entire chunk.

Also as I said above I don't think we should call `kerchunk.zarr.ZarrToZarr` inside `virtualizarr.open_virtual_dataset`, we should rewrite that function ourselves (but not in this PR).

virtualizarr/manifests/manifest.py
TomNicholas: Okay! So it's intentional.

I think this is fine - it still deals fine with the types the hints claim it will, plus you're saying that it will also silently succeed on some other types too.

> Ideally we'd add types to kerchunk, but I don't have time for that right now.

Yep. In the long term the plan is to re-implement kerchunk's readers without going via the intermediate kerchunk reference format, at which point we will hopefully be able to remove all the code in `kerchunk.py` anyway.

virtualizarr/manifests/manifest.py
maresb: Oh, sorry, I gave you an example for the wrong case.

I don't actually have any particular case in mind where data is inlined. I was assuming that inlined data is possible because kerchunk does it. Also I was hoping that I'd be able to use VirtualiZarr to open references that I've already kerchunked, in which case it would certainly occur.

If the only intended use of kerchunk is internal, should I remove that check? Or at least reword the text so that it's not a TODO?

TomNicholas: > Also I was hoping that I'd be able to use VirtualiZarr to open references that I've already kerchunked, in which case it would certainly occur.

Currently the only intended use case of kerchunk is internal, but this case:

> Also I was hoping that I'd be able to use VirtualiZarr to open references that I've already kerchunked, in which case it would certainly occur.

we would like to support, which is being (intermittently) worked on in #119 and recently #186.

I think that the minimal version of this PR does not need to handle kerchunk-inlined data though, and we should merge this useful PR before worrying about that.

maresb:
I think I'm all caught up with your comments.

I don't think you're asking for a change regarding the `NotImplementedError` on inlined data, but I'm not certain about this.

Please let me know if you'd like any further changes, or if you'd like me to rebase.

Thanks @TomNicholas!

TomNicholas:
I think this looks good, thank you @maresb !

Issue #163, opened 2024-06-27
Transposition via codec


TomNicholas:
In https://github.com/fsspec/kerchunk/issues/377#issuecomment-1798683304 @ivirshup suggested that virtual zarr arrays could support

> Transpose/ permute dims (re-order chunks + add a transpose codec)

I had forgotten about this suggestion until re-reading just now. We should be able to do that in virtualizarr pretty easily.

juntyr:
I just ran into this - it would be very useful to be able to e.g. `da.transpose("time", "lon", "lat")` to a common coordinate order

Issue #86, opened 2024-04-19
Arbitrary chunking of uncompressed files (e.g. netCDF3)


TomNicholas:
@rabernat made the interesting point to me that uncompressed files (e.g. netCDF3 files) have no specific chunking, as you can start reading bytes from any point in the file immediately, with no minimum unit of decompression to do first.

I'm not totally sure what this implies for VirtualiZarr generating references from netCDF3 files, as it's still meaningful to talk about one chunk per file concatenated together in a manifest. 

Perhaps this is something that should ultimately be taken account of in zarr readers: that reading bytes from an uncompressed array does not require loading an entire chunk into memory first.

TomNicholas:
It seems great minds think alike https://medium.com/pangeo/using-kerchunk-with-uncompressed-netcdf-64-bit-offset-files-cloud-optimized-access-to-hycom-ocean-9008ba6d0d67

That post uses [kerchunk.utils.subchunk](https://fsspec.github.io/kerchunk/_modules/kerchunk/utils.html#subchunk) (which I did not know existed), but I wonder if a neater API to use for a similar feature in virtualizarr could just be to add a `chunks` argument to `open_virtual_dataset`, that can only be used for uncompressed files?

TomNicholas:
In fact one could imagine a `.rechunk` method on a `ManifestArray`, but it would raise unless the `ManifestArray` pointed to contiguous-on-disk uncompressed data.

TomNicholas:
Which PR?

> Is that still the approach that you'd take with virtualizarr 2?


I would make this a Parser-specific kwarg. It only makes sense for file formats that are uncompressed, or which could potentially be uncompressed (e.g Zarr). I would just add an option to the NetCDF3 parser specifically.

jsignell:
> You could probably steal the logic from that PR and use it inside the parser

This is 100% my plan :)

jsignell:
Glancing at the kerchunk docs I originally thought that I'd be able to set `max_chunk_size` on the parser init under reader_options, but setting  `max_chunk_size` had no impact on the chunks and upon closer inspection it looks like that kwarg was never actually hooked up in kerchunk (this error got lost later on, but the behavior was never implemented: https://github.com/fsspec/kerchunk/blame/89b3fd5af5d8d672e6c79025cd2cafc7169806b2/kerchunk/netCDF3.py#L127-L128). 

[...]

PR #193, opened 2024-07-17
Conformant ZarrV3 codecs and fill values

ghidalgo3:
This change serializes the `ZArray` codec attributes `filters`, `compressor`, and `order` according to the ZarrV3 specification, along with the default fill value of the array. This should hopefully allow a theoretical ZarrV3 reader to read a zarr store produced by VirtualiZarr, one day.

I would consider this a breaking change because any existing VirtualiZarr stores that used the "before" format shown here would be unreadable now, but given that those metadata files were invalid zarrv3 anyway I'm not sure if it's worth handling that at possibility at read time. If you think it necessary, reviewer, I will implement it.

I'm also not too sure if adding the `transpose` and `bytes` codec unconditionally is necessary. I think readers can assume that no transpose codec means no data transposition obviously, but it's unclear to me if the codec pipeline _must_ declare either one of:

- https://zarr-specs.readthedocs.io/en/latest/v3/codecs/sharding-indexed/v1.0.html or 
- https://zarr-specs.readthedocs.io/en/latest/v3/codecs/bytes/v1.0.html

Also, what happens if a source file uses a codec that is not one of the specified codecs of ZarrV3? Does that mean the file cannot be represented in ZarrV3? Seems rather onerous.

It may make sense to also change `_check_same_codecs` to use the `codec_pipeline` list declared in this PR.

## Before
```json
{
    "attributes": {},
    "chunk_grid": {
        "configuration": {
            "chunk_shape": [
                2,
                3
            ]
        },
        "name": "regular"
    },
    "chunk_key_encoding": {
        "configuration": {
            "separator": "/"
        },
        "name": "default"
    },
    "codecs": null,
    "data_type": "int64",
    "dimension_names": [
        "x",
        "y"
    ],
    "fill_value": null,
    "node_type": "array",
    "shape": [
        2,
        3
    ],
    "storage_transformers": [
        {
            "configuration": {
                "manifest": "./manifest.json"
            },
            "name": "chunk-manifest-json"
        }
    ],
    "zarr_format": 3
}
```
## After
```json
{
    "attributes": {},
    "chunk_grid": {
        "configuration": {
            "chunk_shape": [
                2,
                3
            ]
        },
        "name": "regular"
    },
    "chunk_key_encoding": {
        "configuration": {
            "separator": "/"
        },
        "name": "default"
    },
    "codecs": [
        {
            "configuration": {
                "order": [
                    0,
                    1
                ]
            },
            "name": "transpose"
        },
        {
            "configuration": {},
            "name": "bytes"
        },
        {
            "configuration": {
                "level": 1
            },
            "name": "gzip"
        }
    ],
    "data_type": "int64",
    "dimension_names": [
        "x",
        "y"
    ],
    "fill_value": 0,
    "node_type": "array",
    "shape": [
        2,
        3
    ],
    "storage_transformers": [
        {
            "configuration": {
                "manifest": "./manifest.json"
            },
            "name": "chunk-manifest-json"
        }
    ],
    "zarr_format": 3
}
```

- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

Issues:
- #163
- #94


TomNicholas:
> I would consider this a breaking change because any existing VirtualiZarr stores that used the "before" format shown here would be unreadable now, but given that those metadata files were invalid zarrv3 anyway I'm not sure if it's worth handling that at possibility at read time. If you think it necessary, reviewer, I will implement it.

That was just a proof-of-principle, I say that anyone bold enough to use that before virtualizarr even released was doing so at their own risk 😁 So don't bother trying to account for that.

TomNicholas:
> Also, what happens if a source file uses a codec that is not one of the specified codecs of ZarrV3? Does that mean the file cannot be represented in ZarrV3? Seems rather onerous.

I'm going to tag @sharkinsspatial here who I think has thought about this more than I have

virtualizarr/zarr.py
```diff
@@ -134,6 +138,80 @@ def replace(
...
+        https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#fill-value
+        """
+        # numpy dtypes's hierarchy lets us avoid checking for all the widths
+        # https://numpy.org/doc/stable/reference/arrays.scalars.html
+        if self.dtype is np.dtype("bool"):
+            return False
+        elif self.dtype is np.dtype("int"):
+            return 0
+        elif self.dtype is np.dtype("float"):
+            return "NaN"
```
TomAugspurger, line 154: It seems like the *default* fill value for float is 0.0:

```python
import zarr
import json
store = zarr.store.MemoryStore(mode="w")
z = zarr.empty((1, 1), store=store)
z[:]

array([[0.]])
```

(I'm not sure where on the Array / Store / Other object that information lives.)

It'd be nice if zarr-python had this as a constant that we could reuse. Would that make sense, or is there some reason not to I'm missing?

virtualizarr/zarr.py
```diff
@@ -134,6 +138,80 @@ def replace(
...
+        ).get_config()
+        compressor_id = compressor_codec.pop("id")
+        compressor = dict(name=compressor_id, configuration=compressor_codec)
+
+        # https://zarr-specs.readthedocs.io/en/latest/v3/codecs/transpose/v1.0.html#transpose-codec-v1
+        # Either "C" or "F", defining the layout of bytes within each chunk of the array.
+        # "C" means row-major order, i.e., the last dimension varies fastest;
+        # "F" means column-major order, i.e., the first dimension varies fastest.
+        if self.order == "C":
+            order = tuple(range(len(self.shape)))
```
TomAugspurger, line 199: nitpick:

```suggestion
            order = tuple(enumerate(self.shape))
```

virtualizarr/zarr.py
```diff
@@ -134,6 +138,80 @@ def replace(
...
+        elif self.order == "F":
+            order = tuple(reversed(range(len(self.shape))))
```
TomAugspurger, line 201: nitpick x 2:

```suggestion
            order = tuple(reversed(enumerate(self.shape)))
```

virtualizarr/zarr.py
TomAugspurger: I also see a reference to NaN as a default at https://github.com/zarr-developers/VirtualiZarr/pull/193/files#diff-f5a7b84b3378d903e91ebf06f2db06dca5ad55d12e7c3bf8537a9b9bb1c4cfa0R361 (present on main).

ghidalgo3: I think the specification doesn't require a specific number, just that it not be null. See the note at the bottom of the `fill_value` section https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#fill-value . However `zarr-python` does default to 0 eventually and not NaN https://github.com/zarr-developers/zarr-python/blob/37a8441c20dae3b284803bb1b0d2e6c8f040fb3e/src/zarr/array.py#L231C9-L235C31 . I may have had some trouble with the unit tests, but I think it's better to be as similar as possible to `zarr-python`, I'll change the defaults to 0s.

virtualizarr/zarr.py
ghidalgo3: Actually these are different:
```python
>>> s = [1,2,3]
>>> a = tuple(range(len(s)))
>>> b = tuple(enumerate(s))
>>> a
(0, 1, 2)
>>> b
((0, 1), (1, 2), (2, 3))
```
I believe I want `a` and not `b`: https://zarr-specs.readthedocs.io/en/latest/v3/codecs/transpose/v1.0.html#configuration-parameters

TomAugspurger: Whoops, sorry 😅

virtualizarr/zarr.py
```diff
@@ -134,6 +148,61 @@ def replace(
...
+            ]
+        else:
+            filters = []
+
+        # Noting here that zarr v3 has very few codecs specificed in the official spec,
+        # and that there are far more codecs in `numcodecs`. We take a gamble and assume
+        # that the codec names and configuration are simply mapped into zarrv3 "configurables".
+        compressor_codec = numcodecs.get_codec(
+            # default to gzip because it is officially specified in the zarr v3 spec
+            dict(id=self.compressor or "gzip")
```
TomAugspurger, line 180: Just making sure I understand the meaning of `compressor` here. Is this required to be the compressor the actual array chunks were written with, by whoever wrote the NetCDF / Zarr / etc. file? If so, I don't think the concept of a "default compressor" makes sense here... IIUC, zarr-python will come along looking for some data, follow the ChunkManifest, make a range request for some bytes, and then try to decode them, at which point it would fail (since the data isn't actually gzipped?).

But I see you say that `compressor: ArrayBytesCodec` is mandatory, so we have to have something?

TomAugspurger (approved):
Looks good to me overall. One question about the default compressor in the new `codecs` pipeline.

virtualizarr/zarr.py
ghidalgo3: >  Is this required to be the compressor the actual array chunks were written with, by whoever wrote the NetCDF / Zarr / etc. file?

Yes. When VirtualiZarr constructs a `ZArray` by reading a valid store, the compressor is read from the store. Even in ZarrV2, [compressors are mandatory](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html#metadata), so unless people are reading malformed metadata, this line we're discussing should never fall through to the default. 

The only time a `None` compressor is defaulted into a `gzip` is when someone constructs a `ZArray` in memory and sets `compressor=None`. The unit tests do this. But so does Sean's PR in the `hdf.py` https://github.com/zarr-developers/VirtualiZarr/pull/87/files now I'm thinking VirtualiZarr should throw if someone passes in a `None` compressor to conform to both zarrv2 and zarrv3 semantics.

> compressor: ArrayBytesCodec is mandatory, so we have to have something?

The store we're reading also had to have something, so unless everyone is misinterpreting the spec I think we can assume `compressor` will be set for us. In that case, I think having a default compressor is not necessary and we should probably disallow None compressors.

virtualizarr/tests/test_zarr.py
```diff
@@ -15,18 +20,63 @@ def test_zarr_v3_roundtrip(tmpdir):
...
+    assert isinstance(metadata["shape"], list) and all(
+        isinstance(dim, int) for dim in metadata["shape"]
+    )
+    assert isinstance(metadata["data_type"], str) or isconfigurable(
+        metadata["data_type"]
+    )
+    assert isconfigurable(metadata["chunk_grid"])
+    assert isconfigurable(metadata["chunk_key_encoding"])
+    assert any(
+        isinstance(metadata["fill_value"], t) for t in (bool, int, float, str, list)
```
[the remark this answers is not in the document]
ghidalgo3: Yes, thanks for the hint, more pythonic.

virtualizarr/zarr.py
TomAugspurger: > The only time a None compressor is defaulted into a gzip is when someone constructs a ZArray in memory and sets compressor=None. 

Thanks for that clarification.

> now I'm thinking VirtualiZarr should throw if someone passes in a None compressor to conform to both zarrv2 and zarrv3 semantics.

That sounds right, but probably best as a separate PR.

TomAugspurger: cc @sharkinsspatial, for whether this will affect the compressor stuff in your PR at https://github.com/zarr-developers/VirtualiZarr/pull/87/files. https://github.com/zarr-developers/VirtualiZarr/pull/87/files#diff-d404da2ebd34ec3696e439dfb0ccbb16214cbb9b4f6c5bdcc2083b3eec270f4dR190 specifically I think.

ghidalgo3: TODO: setting default `gzip` is a bad idea, real life Kerchunk stores have compressor=None

ghidalgo3:
Thanks @t-mcneely for pairing on this with me! I'll write up the results of our PoC with GOES data.

virtualizarr/zarr.py
TomNicholas: I think this could also be deferred to a later PR, especially if the true solution is to make it clearer what the default is upstream.

virtualizarr/zarr.py
```diff
@@ -82,8 +98,8 @@ def from_kerchunk_refs(cls, decoded_arr_refs_zarray) -> "ZArray":
         compressor = decoded_arr_refs_zarray["compressor"]
         # deal with an inconsistency in kerchunk's tiff_to_zarr function
         # TODO should this be moved to the point where we actually call tiff_to_zarr? Or ideally made consistent upstream.
-        if compressor is not None and "id" in compressor:
-            compressor = compressor["id"]
+        # if compressor is not None and "id" in compressor:
+        #     compressor = compressor["id"]
```
TomNicholas, line 102: Is this now unnecessary? In which case should it be deleted?

virtualizarr/zarr.py
```diff
@@ -277,21 +345,71 @@ def metadata_from_zarr_json(filepath: Path) -> tuple[ZArray, list[str], dict]:
...
+
+def _num_codec_config_to_configurable(num_codec: dict) -> dict:
+    """
+    Convert a numcodecs codec into a zarr v3 configurable.
+    """
+    num_codec_copy = num_codec.copy()
+    return {"name": num_codec_copy.pop("id"), "configuration": num_codec_copy}
+
+
+def _default_fill_value(dtype: np.dtype) -> Union[bool, int, float, str, list]:
```
TomNicholas, line 398: Could this function just be a global variable? Of type `dict[np.dtype: Union[bool, int, float, str, list]]`?

(Also we should be able to use `|` instead of `Union`)

TomNicholas: Then use `ZARR_DEFAULT_FILLVALUES.get(dtype, default=0.0)`?

TomNicholas:
Thanks for going into these weeds guys! This is amazing. Just a couple of extremely minor comments then I will merge it.

virtualizarr/zarr.py
TomNicholas: > we should probably disallow None compressors.

FYI I think https://github.com/zarr-developers/VirtualiZarr/issues/86 represents a legitimate real use case for having no compressor!

virtualizarr/zarr.py
TomNicholas: Oh sorry looks like `.get(dtype, 0.0)` is the right syntax, as `default` is positional-only.

TomNicholas:
Thanks @ghidalgo3 !!

PR #196, opened 2024-07-19
https access fix

### LINKED ISSUE #195: open_virtual_dataset HTTPS access fails
Currently `open_virtual_dataset` cannot open a dataset when given a public HTTPS url. This happens because [`reader_options` is set on this line](https://github.com/zarr-developers/VirtualiZarr/blob/5d0851976778ebaf52eb6df57970b0706a892d98/virtualizarr/xarray.py#L128) to s3-like syntax which probably messes up the https request. The functionality is correct in `_fsspec_openfile_from_filepath` so it's a pretty simple fix to just update the linked line in `xarray.py`

Also noticed by @kthyng in https://github.com/zarr-developers/VirtualiZarr/pull/186#issuecomment-2231450332

Here is an example:
```python
>>> from virtualizarr import open_virtual_dataset
>>> open_virtual_dataset("https://github.com/pydata/xarray-data/raw/master/ROMS_example.nc", indexes={})
Traceback (most recent call last): ...
FileNotFoundError: https://github.com/pydata/xarray-data/raw/master/ROMS_example.nc
```

```python
>>> import fsspec
>>> z = fsspec.filesystem("https").open("https://github.com/pydata/xarray-data/raw/master/ROMS_example.nc")
>>> z.readline()
>>> b'\x89HDF\r\n'
```

ayushnag:
- [x] Closes #195
- [x] Tests passing

### DIFF
--- a/virtualizarr/tests/test_xarray.py
+++ b/virtualizarr/tests/test_xarray.py
@@ -376,7 +376,7 @@ class TestReadFromURL:
             with pytest.raises(NotImplementedError):
                 vds = open_virtual_dataset(url, reader_options={}, indexes={})
         else:
-            vds = open_virtual_dataset(url, reader_options={}, indexes={})
+            vds = open_virtual_dataset(url, indexes={})
             assert isinstance(vds, xr.Dataset)
 
 
--- a/virtualizarr/xarray.py
+++ b/virtualizarr/xarray.py
@@ -124,11 +124,6 @@ def open_virtual_dataset(
             storepath=filepath, drop_variables=drop_variables, indexes=indexes
         )
     else:
-        if reader_options is None:
-            reader_options = {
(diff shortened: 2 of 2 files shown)


TomAugspurger:
Thanks for tracking this down. I think that at https://github.com/zarr-developers/VirtualiZarr/blob/5d0851976778ebaf52eb6df57970b0706a892d98/virtualizarr/tests/test_xarray.py#L379, the `reader_options={}` can be removed and act as a unit test for this.

TomNicholas:
Thanks @ayushnag! Yes I agree with @TomAugspurger - any bugfix should really be accompanied by a regression test, and Tom's suggestion seems like an easy way to add this.

TomNicholas:
Thanks @ayushnag !

PR #197, opened 2024-07-19
WIP: Add rechunking example

thodson-usgs:
This PR adds an example script demonstrating how to rechunk a VirtualiZarr dataset with Cubed.

However, this is still a WIP. I'm creating the PR to elicit feedback about what changes might be necessary in order for the script to run as intended. @TomNicholas and @norlandrhagen might have some thoughts. 

After creating the combined virtual dataset, I specify the source chunking before passing it off the Cubed for rechunking
```python
source_chunks = {'Time':1, 'south_north':250, 'west_east':320}

combined_chunked = combined_ds.chunk(
    chunks = source_chunks,
)

combined_chunks
```
returns
```
Frozen({'Time': (1, 1, 1, 1), 'south_north': (250,), 'west_east': (320,), 'interp_levels': (9,), 'soil_layers_stag': (4,)})
```
The virtual dataset contains four files, indicated by `'Time': (1, 1, 1, 1)`.

Then I attempt to rechunk:
```python
from cubed.primitive.rechunk import rechunk

target_chunks = {'Time':5, 'south_north':25, 'west_east':32}

rechunk(
    combined_chunked['TMAX'], # requires shape attr, so can't pass full Dataset
    target_chunks=target_chunks,
    source_array_name='virtual',
    int_array_name='temp',
    allowed_mem=2000,
    reserved_mem=1000,
    target_store="test.zarr",
    #temp_store="s3://cubed-thodson-temp",
)
```
which errors with
```bash
TypeError: can't multiply sequence by non-int of type 'tuple'
```
Apparently, Cubed won't tolerate the Time chunk tuple `'Time': (1, 1, 1, 1)`. Is there a simple way to convert it to `Time': (1, )`?  Alternatively, I could prepare a PR to Cubed, which would set the memory constraint around the largest chunk size when chunks are variable.

## kerchunk
I also tested this workflow with `kerchunk` but I ran into a bug while following the [Pythia cookbook example](https://projectpythia.org/kerchunk-cookbook/notebooks/foundations/02_kerchunk_multi_file.html):
```
/home/runner/miniconda3/envs/kerchunk-cookbook/lib/python3.10/site-packages/kerchunk/combine.py:370: UserWarning: Concatenated coordinate 'Time' contains less than expected number of values across the datasets: [0]
  warnings.warn(
```

### DIFF
--- a/examples/rechunking/Dockerfile_virtualizarr
+++ b/examples/rechunking/Dockerfile_virtualizarr
@@ -1,0 +1,59 @@
+# Python 3.11
+FROM python:3.11-slim-buster
+
+COPY requirements.txt requirements.txt
+
+RUN apt-get update \
+    # Install aws-lambda-cpp build dependencies
+    && apt-get install -y \
+      g++ \
+      make \
+      cmake \
+      unzip \
+    # cleanup package lists, they are not used anymore in this image
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-cache search linux-headers-generic
+
+ARG FUNCTION_DIR="/function"
+
+# Copy function code
+RUN mkdir -p ${FUNCTION_DIR}
+
+# Update pip
+# NB botocore/boto3 are pinned due to https://github.com/boto/boto3/issues/3648
+#    using versions from https://github.com/aio-libs/aiobotocore/blob/72b8dd5d7d4ef2f1a49a0ae0c37b47e5280e2070/setup.py
+#    due to s3fs dependency
+RUN pip install --upgrade --ignore-installed pip wheel six setuptools \
+    && pip install --upgrade --no-cache-dir --ignore-installed \
+        awslambdaric \
+        botocore==1.29.76 \
+        boto3==1.26.76 \
+        redis \
+        httplib2 \
+        requests \
+        numpy \
+        scipy \
+        pandas \
+        pika \
+        kafka-python \
+        cloudpickle \
+        ps-mem \
+        tblib
+
+# Set working directory to function root directory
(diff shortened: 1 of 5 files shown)


TomNicholas:
Great to see you trying this! First time someone has tried to use these two
libraries together!

I don't think you should import the rechunk primitive from Cubed. I think
instead you should open the kerchunked dataset as an Xarray dataset using
cubed-xarray, then call Xarray's chunk method with the desired chunks. This
might smooth out low-level chunk type considerations for you, and if it
doesn't that's a bug.

thodson-usgs:
> I don't think you should import the rechunk primitive from Cubed. I think instead you should open the kerchunked dataset as an Xarray dataset using cubed-xarray, then call Xarray's chunk method with the desired chunks.

Yes, that seems to work, but I'm still working through several errors when I write out to Zarr. I'll report more in a day or two.

TomNicholas:
> Yes, that seems to work, but I'm still working through several errors when I write out to Zarr. I'll report more in a day or two.

Great - very curious to see the details.

I think what you're doing here should live in the cubed repo though - once you have the kerchunk reference files on disk virtualizarr is out of the picture, all of the rechunking is about using cubed. I do think that this use case would make an important example to have in the cubed docs though - as its basically showing how the original `rechunker` package is just a special case of cubed (cc @tomwhite).

tomwhite:
Sounds great. Happy for this to be added as a Cubed example.

thodson-usgs:
Closing and opening a PR on cubed cubed-dev/cubed#520.

PR #199, opened 2024-07-22
Rechunk method for uncompressed arrays

### LINKED ISSUE #86: Arbitrary chunking of uncompressed files (e.g. netCDF3)

TomNicholas:
@rsignell this is inspired by your blog post (still a WIP for now)

The idea is that you can simply do
```python
vds = open_virtual_dataset('uncompressed_netcdf3.nc')
subchunked_vds = vds.chunk(time=1)
```

- [x] Closes #86
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/array.py
+++ b/virtualizarr/manifests/array.py
@@ -6,7 +6,7 @@ import numpy as np
 from ..kerchunk import KerchunkArrRefs
 from ..zarr import ZArray
 from .array_api import MANIFESTARRAY_HANDLED_ARRAY_FUNCTIONS, _isnan
-from .manifest import ChunkManifest
+from .manifest import ChunkManifest, subchunk
 
 
 class ManifestArray:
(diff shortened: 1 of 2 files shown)


TomNicholas:
Note to self: we should add more validation to the `ZArray` class to check that the `chunks` attribute is a tuple of positive integers, and move the `zarray.replace` call to the start of the method to catch invalid input early.

rsignell:
> we can rechunk in however many dimensions we want!

I love that aspect!

TomNicholas:
> but the new integration test fails because Dataset.chunk is dispatching to dask's version of .rechunk somewhere. This may require a change to xarray's ChunkManagerEntrypoint upstream to fix.

https://github.com/pydata/xarray/pull/9286 has now progressed far enough that this PR works for me at least! Passing all tests locally 🟢

Issue #123, opened 2024-05-21
Serverless parallelization of reference generation


TomNicholas:
Finding byte ranges in every file in an archival dataset is an embarrassingly-parallel problem, which might be a good fit for serverless.

This step is analogous to the `parallel=True` option to `xr.open_mfdataset`, which wraps `xr.open_dataset` in `@dask.delayed` for each file to parallelize the opening step (totally separate from any dask.array tree reduction after the arrays have been created). 

With that motivation, it's been suggested that we add a delayed-like function to Cubed https://github.com/cubed-dev/cubed/issues/311, which could in theory plug in to `xarray.open_mfdataset`.

A simpler way to test this idea would be just to skip the cubed layer and use a lithops map-gather (or whatever the correct primitive is) to try out serverless generation of references. I think for this to work the resulting virtual datasets need to be able to be gathered onto one node, but the discussion in https://github.com/TomNicholas/VirtualiZarr/issues/104#issuecomment-2103039189 indicates that this should be okay (at least after #107 is complete).

xref https://github.com/TomNicholas/VirtualiZarr/issues/95

cc @tomwhite @rabernat

TomNicholas:
I'm reading about how reduce jobs work in lithops but I don't fully get it. Is it the case that we need each serverless worker to write out the references it read to some globally-accessible storage? So that they can all be accessed by a single node/worker for the concatenation step?

dcherian:
Yes though you can also tree-reduce. IIUC serverless tasks don't communicate between each other, and must communicate through storage.

Though perhaps lithops has something like https://modal.com/docs/guide/dicts-and-queues to enable cross-task communication.

tomwhite:
> I'm reading about how reduce jobs work in lithops but I don't fully get it. Is it the case that we need each serverless worker to write out the references it read to some globally-accessible storage? So that they can all be accessed by a single node/worker for the concatenation step?

I haven't used Lithops `map_reduce` (we just use `map` in Cubed), but the reduce function runs in a single container and will receive all the map outputs so they can be combined in the way you specify. (You don't need to write anything to storage, Lithops will do that part for you.) Given that the map output is a small amount of metadata Lithops `map_reduce` sounds like a good fit for the problem.

TomNicholas:
[...]

PR #203, opened 2024-07-29
Add example to create a virtual dataset using lithops

thodson-usgs:
At the suggestion of @TomNicholas, I created a simple example using lithops (and serverless compute) to create a virtual dataset from a list of netcdf files hosted on s3.

This PR depends on the fix provided in #206 (now merged).

## Resolved with #206
The workflow was broken in the latest version of VirtualiZarr.
 The example runs fine on  zarr-developers/VirtualiZarr@5d0851976778ebaf52eb6df57970b0706a892d98.
However, using zarr-developers/VirtualiZarr@179bb2ab42664546dd243a2e04fab8737846229a the workflow will run, but complains about `ValueError: Could not convert object to NumPy datetime` when I open the dataset using `xarray`:

```bash
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/zarr/meta.py", line 127, in decode_array_metadata
    fill_value = cls.decode_fill_value(meta["fill_value"], dtype, object_codec)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/zarr/meta.py", line 260, in decode_fill_value
    return np.array(v, dtype=dtype)[()]
           ^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Could not convert object to NumPy datetime

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/xarray/backends/api.py", line 571, in open_dataset
    backend_ds = backend.open_dataset(
                 ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/kerchunk/xarray_backend.py", line 12, in open_dataset
    ref_ds = open_reference_dataset(
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/kerchunk/xarray_backend.py", line 46, in open_reference_dataset
    return xr.open_dataset(m, engine="zarr", consolidated=False, **open_dataset_options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/xarray/backends/api.py", line 571, in open_dataset
    backend_ds = backend.open_dataset(
                 ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/xarray/backends/zarr.py", line 1182, in open_dataset
    ds = store_entrypoint.open_dataset(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/xarray/backends/store.py", line 43, in open_dataset
    vars, attrs = filename_or_obj.load()
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/xarray/backends/common.py", line 221, in load
    (_decode_variable_name(k), v) for k, v in self.get_variables().items()
                                              ^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/xarray/backends/zarr.py", line 563, in get_variables
    return FrozenDict(
           ^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/xarray/core/utils.py", line 443, in FrozenDict
    return Frozen(dict(*args, **kwargs))
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/xarray/backends/zarr.py", line 563, in <genexpr>
    return FrozenDict(
                     ^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/zarr/hierarchy.py", line 691, in _array_iter
    yield _key if keys_only else (_key, self[key])
                                        ~~~~^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/zarr/hierarchy.py", line 467, in __getitem__
    return Array(
           ^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/zarr/core.py", line 170, in __init__
    self._load_metadata()
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/zarr/core.py", line 193, in _load_metadata
    self._load_metadata_nosync()
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/zarr/core.py", line 207, in _load_metadata_nosync
    meta = self._store._metadata_class.decode_array_metadata(meta_bytes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thodson/micromamba/envs/cubed-dev/lib/python3.11/site-packages/zarr/meta.py", line 141, in decode_array_metadata
    raise MetadataError("error decoding metadata") from e
zarr.errors.MetadataError: error decoding metadata

```


TomNicholas:
This is awesome @thodson-usgs ! Thanks for trying this out!

> I'm marking this WIP, because the workflow is broken in the latest version of VirtualiZarr.

I'll raise a separate issue to track whatever regression has occurred there, so we can talk about the cool serverless `map_reduce` you've done here! (which is relevant to #123)

My conception of what `lithops.map_reduce` is actually doing is quite fuzzy. Is it saving out any intermediate results to a storage layer? Is it doing multiple rounds of reduction (i.e. a tree-reduce)? Do you foresee any scaling issues with this approach?

thodson-usgs:
Hmm. I pulled the lastest version of VirtualiZarr in my testing environment, but I neglected to rebuild the runtime image, so I'll double check that.

> My conception of what `lithops.map_reduce` is actually doing is quite fuzzy. Is it saving out any intermediate results to a storage layer? Is it doing multiple rounds of reduction (i.e. a tree-reduce)? Do you foresee any scaling issues with this approach?

Right, I set up s3 storage for cubed, but I think this workflow is entirely in memory.  So, we'll invariably hit scaling issues unless avoid the reduce and by writing to disk during the map operation. Nevertheless, I'm really excited by how easy this was to set up, and I hope others will help improve upon it.

thodson-usgs:
> Hmm. I pulled the lastest version of VirtualiZarr in my testing environment, but I neglected to rebuild the runtime image, so I'll double check that.

No, I rebuilt the image to use the latest commit, and the error persisted, so I believe it's real.

TomNicholas:
> Right, I set up s3 storage for cubed

You're not using cubed at all here, that's for the actual rechunking.

> I think this workflow is entirely in memory.

Lithops does have the ability to persist things - did you set up the storage layer for that?

> So, we'll invariably hit scaling issues unless avoid the reduce and by writing to disk during the map operation. Nevertheless, I'm really excited by how easy this was to set up, and I hope others will help improve upon it.

My plan for scaling this to arbitrary size is actually to use cubed for the virtualizarr array reduction too - see https://github.com/zarr-developers/VirtualiZarr/issues/123#issuecomment-2244001155. I expect this to be pretty complicated to achieve though - I'm not even sure if it's possible yet.

TomNicholas:
> No, I rebuilt the runtime image using the latest VirtualiZarr commit, and the error persisted, so I believe it's real.

Presumably this error can be reproduced without lithops involved at all?

thodson-usgs:
> Presumably this error can be reproduced without lithops involved at all?

Good point. I'll try that next, though something is fishy or else how did this work with previous versions?

> My plan for scaling this to arbitrary size is actually to use cubed for the virtualizarr array reduction too - see [#123 (comment)](https://github.com/zarr-developers/VirtualiZarr/issues/123#issuecomment-2244001155). I expect this to be pretty complicated to achieve though - I'm not even sure if it's possible yet.

Ah, okay. So maybe my hope of simply creating a skeleton zarr, then writing the meta-chunks during the map with to_zarr(region=) is half baked.

TomNicholas:
> I'll try that next, though something is fishy or else how did this work with previous versions?

My current guess is that we simply introduced some accidental regression in virtualizarr recently. The way to find it is to (1) reproduce the error without all the lithops stuff (opening and concatenating 2 files should be enough), then (2) use `git bisect` to find the offending commit.

> Ah, okay. So maybe my hope of simply creating a skeleton zarr, then writing the meta-chunks during the map with `to_zarr(region=)` is half baked. (The current workflow writes to json, but I'll test cloud optimized formats once this is running.)

If you write a manifest to zarr right now (i.e. a "virtual zarr store") you have no way of opening or loading the data via xarray/zarr, because no zarr reader understands what a `manifest.json` is yet.

My model of what we're trying to do is:
- We need to get to one in-memory virtual `xr.Dataset`, on one worker, containing references to all the netcdf files. Once we have that we can save it to kerchunk json / kerchunk parquet / zarr manifests / whatever and we would be done (for a large number of references we should currently save to kerchunk parquet format).
- We know that that virtual dataset will fit on one worker, because we estimate the memory requirements in https://github.com/zarr-developers/VirtualiZarr/issues/104#issuecomment-2103039189.
- Problem is that inspecting the netCDF files to generate chunk references is slow, and we want to parallelize that across many workers (i.e. call `open_virtual_dataset(`some_file.nc')` on each worker).
- This step is embarrassingly parallel and hence can be done nicely with serverless functions, i.e. lithops map.
- Issue is that once we open them we have to combine them in-memory via concatenation. If we used dask we could communicate them all to one worker, but with serverless the only way to communicate is to write to persistent storage then read from that storage.
- I believe lithops `map_reduce` uses a storage layer to hold intermediate results https://lithops-cloud.github.io/docs/source/design.html#computation-flow. So it does a `map`, then `reduce`, each of which must read from and write to that storage layer.
- Using that might be enough for a 1D reduce job, especially if there isn't too many files. But to do it in N-dimensions at large scale we really want a N-dimensional tree-reduce using serverless functions.
- That's what cubed implements, for the case of N-D arrays. Then the question is how to get cubed to understand what to do with a `ManifestArray` instead of a `np.ndarray`?

thodson-usgs:
Updated with the changes made in #206. Also fine to close this example, if `cubed` can do this better.

examples/virtualizarr-with-lithops/README.md
```diff
@@ -0,0 +1,41 @@
+# Generate a virtual zarr dataset using lithops
+
+This example walks through how to create a virtual dataset from a collection of
+necdf files on s3 using lithops to open each file in parralel then concatenate
```
TomNicholas, line 4:
```suggestion
necdf files on s3 using lithops to open each file in parallel then concatenate
```
```suggestion
necdf files on s3 using lithops to open each file in parallel then concatenate
```

TomNicholas, line 4:
```suggestion
netCDF files on s3 using lithops to open each file in parallel then concatenate
```

examples/virtualizarr-with-lithops/README.md
```diff
@@ -0,0 +1,41 @@
...
+lithops runtime build -b aws_lambda -f Dockerfile_virtualizarr virtualizarr-runtime
+```
+
+1. Run the script
+```bash
+python virtualizarr-with-lithops.py
+```
+
+## Cleaning up
+To rebuild the Litops image, delete the existing one by running
```
TomNicholas, line 38:
```suggestion
To rebuild the Lithops image, delete the existing one by running
```

examples/virtualizarr-with-lithops/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+1. Set up a Python environment
+```bash
+conda create --name virtualizarr-lithops -y python=3.11
+conda activate virtualizarr-lithops
+pip install -r requirements.txt
+```
+
+1. Configure compute and storage backends for [lithops](https://lithops-cloud.github.io/docs/source/configuration.html).
```
TomNicholas, line 22:
```suggestion
2. Configure compute and storage backends for [lithops](https://lithops-cloud.github.io/docs/source/configuration.html).
```

TomNicholas (approved):
Thanks @thodson-usgs !

> Also fine to close this example, if cubed can do this better.

I think doing this will Cubed will be quite involved. If this works effectively for you then that's already awesome!

TomNicholas:
Thank you @thodson-usgs !!

examples/virtualizarr-with-lithops/virtualizarr-with-lithops.py
```diff
@@ -0,0 +1,59 @@
...
+files_paths = fs_read.glob("s3://wrf-se-ak-ar5/ccsm/rcp85/daily/2060/*")
+file_pattern = sorted(["s3://" + f for f in files_paths])
+
+# optionally, truncate file_pattern while debugging
+# file_pattern = file_pattern[:4]
+
+print(f"{len(file_pattern)} file paths were retrieved.")
+
+
+def map_references(fil):
```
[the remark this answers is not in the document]
thodson-usgs: I must've copied that directly from an example. I'll need to check whether this follows some convention or just a typo.

douglatornell:
Just a guess... `file` was a reserved word in Python2. If the example code was old enough, or written by someone with  habits formed in the Python2 days (I'm one of those folks :grin: ), that might explain things.

thodson-usgs:
> Just a guess... `file` was a reserved word in Python2. If the example code was old enough, or written by someone with habits formed in the Python2 days (I'm one of those folks 😁 ), that might explain things.

Ah, @douglatornell! I thought that was the case but then I didn't see `file` among the reserved words. Thanks for clarifying.

douglatornell:
Yeah @thodson-usgs, I did a bit of a double take when I looked at the reserved words list and realized that `file` isn't reserved anymore in Python3.

abarciauskas-bgse:
@thodson-usgs excited to see the lithops integration!

PR #205, opened 2024-07-30
Handle scalar dataset variables

### LINKED ISSUE #194: NetCDF files with many empty variables require large `drop_variables` declaration
I'm not sure if this is a real issue or working as intended, but I wanted to get clarification on it. The [GOES](https://planetarycomputer.microsoft.com/dataset/goes-cmi) dataset's NetCDF files contain many non-array variables such that the call to `virtualizarr.open_virtual_dataset` needs a large number of `drop_variables` to work. All of the dropped variables have in common a `chunk_shape` of `()`, because they're not arrays.

If [`validate_chunk_shape`](https://github.com/zarr-developers/VirtualiZarr/blob/5d0851976778ebaf52eb6df57970b0706a892d98/virtualizarr/manifests/manifest.py#L360) skipped validating empty dicts, it would be possible to open this dataset without having to pass such a `drop_variables` list, but more importantly we could preserve variable metadata when we write a Zarr store for this file and others. Is there a reason VirtualiZarr shouldn't read empty variables like this?

## Repro
```python
import planetary_computer
import pystac_client
import virtualizarr as vz

catalog = pystac_client.Client.open("https://planetarycomputer.microsoft.com/api/stac/v1/")
search = catalog.search(
    collections=["goes-cmi"],
    datetime=["2019-08-01T12:00:00Z", "2019-08-01T13:00:00Z"],
    query={"goes:image-type": {"eq": "FULL DISK"}},
)
items = sorted(search.item_collection(), key=lambda x: x.datetime)
item = items[0]
vds = vz.open_virtual_dataset(
    planetary_computer.sign(item.assets["CMIP_C15-nc"].href),
    # drop_variables=[
    #     "processing_parm_version_container",
    #     "earth_sun_distance_anomaly_in_AU",
    #     "goes_imager_projection",
    #     "esun",
    #     "algorithm_product_version_container",
    #     "std_dev_brightness_temperature",
    #     "maximum_focal_plane_temperature",
    #     "focal_plane_temperature_threshold_increasing",
    #     "focal_plane_temperature_threshold_exceeded_count",
    #     "y_image",
    #     "nominal_satellite_height",
    #     "outlier_pixel_count",
    #     "focal_plane_temperature_threshold_decreasing",
    #     "min_brightness_temperature",
    #     "planck_bc2",
    #     "nominal_satellite_subpoint_lon",
    #     "nominal_satellite_subpoint_lat",
    #     "valid_pixel_count",
    #     "planck_bc1",
    #     "mean_brightness_temperature",
    #     "planck_fk2",
    #     "percent_uncorrectable_GRB_errors",
    #     "geospatial_lat_lon_extent",
    #     "kappa0",
    #     "percent_uncorrectable_L0_errors",
    #     "t",
    #     "max_brightness_temperature",
    #     "x_image",
    #     "planck_fk1",
    #     "algorithm_dynamic_input_data_container",
    #     "total_number_of_points"],
    reader_options={},
)
```

ghidalgo3:
Some real-world NetCDF files have scalar variables that `kerchunk` parsers to a ChunkDict of `{}`. Previously, these variables would need to be explicitly loaded or dropped to make a `open_virtual_dataset` call work. With this change, the variable will be loaded into `xr.Variables` of dimension 0 and their attributes passed up to the XArray dataset.

- [x] Closes #194
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,4 @@
+import h5py
 import pytest
 import xarray as xr
 
@@ -50,3 +51,21 @@ def netcdf4_files(tmpdir):
     ds2.close()
 
     return filepath1, filepath2
+
+
+@pytest.fixture
+def hdf5_empty(tmpdir):
+    filepath = f"{tmpdir}/empty.nc"
+    f = h5py.File(filepath, "w")
+    dataset = f.create_dataset("empty", shape=(), dtype="float32")
+    dataset.attrs["empty"] = "true"
+    return filepath
+
+
+@pytest.fixture
+def hdf5_scalar(tmpdir):
+    filepath = f"{tmpdir}/scalar.nc"
+    f = h5py.File(filepath, "w")
+    dataset = f.create_dataset("scalar", data=0.1, dtype="float32")
+    dataset.attrs["scalar"] = "true"
+    return filepath
--- a/virtualizarr/tests/test_xarray.py
+++ b/virtualizarr/tests/test_xarray.py
@@ -419,6 +419,32 @@ class TestLoadVirtualDataset:
         }
         mock_read_kerchunk.assert_called_once_with(**args)
 
+    @patch("virtualizarr.kerchunk.parse_array_refs")
+    def test_open_dataset_with_scalars(self, mock_parse_array_refs, tmpdir):
+        expected_attrs = {"units": "K", "long_name": "scalar"}
+        mock_parse_array_refs.return_value = (
+            None,
+            ZArray(
+                chunks=(),
+                shape=(),
+                compressor={},
+                dtype=np.dtype("float32"),
+                fill_value=0.0,
+                filters=None,
+                order="C",
+                zarr_format=2,
+            ),
+            {**expected_attrs, **{"_ARRAY_DIMENSIONS": []}},
+        )
+
+        ds = xr.Dataset()
+        ds["scalar"] = xr.Variable(dims=(), data=None, attrs=expected_attrs)
+        ds.to_netcdf(f"{tmpdir}/scalar.nc")
+        vds = open_virtual_dataset(f"{tmpdir}/scalar.nc")
+        assert vds.scalar.dims == ()
+        assert vds.scalar.data == 0.0
+        assert vds.scalar.attrs == expected_attrs
+
 
 class TestRenamePaths:
     def test_rename_to_str(self, netcdf4_file):
--- a/virtualizarr/tests/test_xarray.py
+++ b/virtualizarr/tests/test_xarray.py
@@ -419,31 +419,15 @@ class TestLoadVirtualDataset:
         }
         mock_read_kerchunk.assert_called_once_with(**args)
 
-    @patch("virtualizarr.kerchunk.parse_array_refs")
-    def test_open_dataset_with_scalars(self, mock_parse_array_refs, tmpdir):
-        expected_attrs = {"units": "K", "long_name": "scalar"}
-        mock_parse_array_refs.return_value = (
-            None,
-            ZArray(
-                chunks=(),
-                shape=(),
-                compressor={},
-                dtype=np.dtype("float32"),
-                fill_value=0.0,
-                filters=None,
-                order="C",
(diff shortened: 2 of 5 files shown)


virtualizarr/tests/test_xarray.py
```diff
@@ -419,6 +419,32 @@ def test_open_virtual_dataset_passes_expected_args(
         }
         mock_read_kerchunk.assert_called_once_with(**args)
 
+    @patch("virtualizarr.kerchunk.parse_array_refs")
```
TomNicholas, line 422: Wait what exactly is this mocking doing?

ghidalgo3: I couldn't get a repro of the issue by just creating an xarray dataset with a 0-dim variable so I had to debug the problem I was getting on a real netCDF file and mock out the return value of `parse_array_refs` to match what I was seeing. Here's a repro if you'd like to give it shot:
```python
import xarray as xr
import planetary_computer
import virtualizarr

def _href_to_fsspec(href: str) -> str:
    url = urlparse(href)
    return f"az://{url.path.lstrip('/')}"

def open_planetary_computer_netcdf(href):
    account_name = href.split("/")[2].split(".")[0]
    container_name = href.split("/")[3]
    fs = planetary_computer.get_adlfs_filesystem(account_name, container_name)
    return fs.open(_href_to_fsspec(href))

def _get_storage_options(href: str, use_sas_as_credential: bool = False) -> dict:
    url = urlparse(href)
    if use_sas_as_credential:
        return {
            "account_name": url.hostname.split(".")[0],
            "account_key": url.query,
            "anon": False,
        }
    else:
        return {
        "account_name": url.hostname.split(".")[0],
        "anon": False,
    }


href = "https://landcoverdata.blob.core.windows.net/esa-cci-lc/netcdf/C3S-LC-L4-LCCS-Map-300m-P1Y-2016-v2.1.1.nc"
# Next line works, you can inspect the `crs` variable
ds = xr.open_dataset(open_planetary_computer_netcdf(href))
# Next line will raise a `ValueError` unpacking a list of no elements when handling the `crs` kerchunk attrs
vds = virtualizarr.open_virtual_dataset(
    _href_to_fsspec(href),
    reader_options={
        "storage_options": _get_storage_options(planetary_computer.sign(href), use_sas_as_credential=True)
    })


```

sharkinsspatial: @ghidalgo3 I just met up with some folks from the HDF group at a conference last week so I can ask them the proper incantation to use so we can create an hdf fixture with a scalar 👍

TomNicholas: Thank you for the explanation! I'm happy to merge this as-is or wait if you wanted to test it some other way.

sharkinsspatial:
@ghidalgo3 Here is an HDF5 fixture you could use for direct testing of this behavior without mocking.

ghidalgo3:
Thanks @sharkinsspatial for the fixtures! I added test cases for both cases without that ugly mock. @TomNicholas if it looks good to you, it's ready to merge.

ghidalgo3:
I'm running into the problem described by #62 but I think I don't want to solve it in this PR.

You can open files with scalar variables, and empty variables, but if you call `vds.virtualize.to_zarr("store.zarr")` then that will error because one of the variables is a numpy array and not a `ManifestArray`. Thats the behavior today anyway, and should be solved once #62 is implemented. If you truly must write the `vds` to disk, just skip loading the variables with `drop_variables`.

Given that, this PR is good to merge.

conftest.py
```diff
@@ -1,3 +1,4 @@
+import h5py
```
TomNicholas, line 1: I think `h5py` should be added to the test requirements?

ghidalgo3: Yes, good catch. Is there a way to make the CI conda install be constructed starting from the `test` extra  + additional conda packages?

TomNicholas: Probably, but let's do that in a follow-up PR.

PR #207, opened 2024-08-02
Handle uncompressed variables in parsing metadata

ghidalgo3:
When reading in a ZarrV3 store, it is very possible that a variable is uncompressed. The code that assigns codecs to `compressors` and `filters` doesn't handle the case when there are no codecs to assign. This PR addresses that gap.

- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/virtualizarr/zarr.py
+++ b/virtualizarr/zarr.py
@@ -368,9 +368,14 @@ def metadata_from_zarr_json(filepath: Path) -> tuple[ZArray, list[str], dict]:
         for codec in metadata["codecs"]
         if codec["name"] not in ("transpose", "bytes")
     ]
-    compressor, *filters = [
+    num_codec_configs = [
         _configurable_to_num_codec_config(_filter) for _filter in all_codecs
     ]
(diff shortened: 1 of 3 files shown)


virtualizarr/zarr.py
```diff
@@ -368,9 +368,14 @@ def metadata_from_zarr_json(filepath: Path) -> tuple[ZArray, list[str], dict]:
         for codec in metadata["codecs"]
         if codec["name"] not in ("transpose", "bytes")
     ]
-    compressor, *filters = [
+    num_codec_configs = [
         _configurable_to_num_codec_config(_filter) for _filter in all_codecs
     ]
+    # Uncompressed variables may have no codecs
+    if num_codec_configs:
```
ghidalgo3, line 375: Is `len(num_codec_configs) > 0` better? Or `not empty(num_codec_configs)`?

TomNicholas: Does this actually cover all cases? What if you have an uncompressed variable that does have other codecs?

PR #208, opened 2024-08-03
Fix for typo? in filetype docstring

jbusecke:
I just ran into a minor, but confusing issue with the docstrings for the `filetype` input for `open_virtual_dataset`. 

When I look at the docstring (specifically [this line](https://github.com/zarr-developers/VirtualiZarr/blob/b34a1ee47572901ed6a3623c97d340157a0fa62a/virtualizarr/xarray.py#L61)), it suggests that this would work:

```python
vds = open_virtual_dataset(
        ...,
        filetype='netCDF3',
) 
```

but it raises the following error

<details>

</details>

the following works:
```python
vds = open_virtual_dataset(
        ...,
        filetype='netcdf3',
) 
```

Looking at the [tests](https://github.com/zarr-developers/VirtualiZarr/blob/b34a1ee47572901ed6a3623c97d340157a0fa62a/virtualizarr/tests/test_xarray.py#L341), I suspect this is simply a typo in the docstring (upercase `'CDF'` vs lowercase `'cdf'`), but maybe I am missing something?

Happy to add more if desired.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/xarray.py
+++ b/virtualizarr/xarray.py
@@ -58,7 +58,7 @@ def open_virtual_dataset(
         File path to open as a set of virtualized zarr arrays.
     filetype : FileType, default None
         Type of file to be opened. Used to determine which kerchunk file format backend to use.
-        Can be one of {'netCDF3', 'netCDF4', 'HDF', 'TIFF', 'GRIB', 'FITS', 'zarr_v3'}.
+        Can be one of {'netcdf3', 'netcdf4', 'hdf4', 'tiff', 'grib', 'fits', 'zarr_v3'}.
         If not provided will attempt to automatically infer the correct filetype from header bytes.
     drop_variables: list[str], default is None
         Variables in the file to drop before returning.


virtualizarr/xarray.py
```diff
@@ -58,7 +58,7 @@ def open_virtual_dataset(
         File path to open as a set of virtualized zarr arrays.
     filetype : FileType, default None
         Type of file to be opened. Used to determine which kerchunk file format backend to use.
-        Can be one of {'netCDF3', 'netCDF4', 'HDF', 'TIFF', 'GRIB', 'FITS', 'zarr_v3'}.
+        Can be one of {'netcdf3', 'netcdf4', 'hdf4', 'tiff', 'grib', 'fits', 'zarr_v3'}.
```
jbusecke, line 61: I am not entirely shure about the other cases here, specifically `'hdf4'` and `'hdf5'`?

TomNicholas:
This should work, because we call `str.lower()` here

https://github.com/zarr-developers/VirtualiZarr/blob/b34a1ee47572901ed6a3623c97d340157a0fa62a/virtualizarr/kerchunk.py#L90

What error did you get? You didn't actually paste it into the comment above.

jbusecke:
Ughh I am a moron, haha. Let me change the test here to demonstrate.

jbusecke:
**If** I remember correctly the error was happening when the `FileType` class is instantiated [here](https://github.com/zarr-developers/VirtualiZarr/blob/a28b21052271c34ce0a99545387e69bdf9ec5029/virtualizarr/xarray.py#L134). I have to admit I have 0 clue what is actually happening within that [class](https://github.com/zarr-developers/VirtualiZarr/blob/a28b21052271c34ce0a99545387e69bdf9ec5029/virtualizarr/kerchunk.py#L39) 😆.

jbusecke:
Hmm that is odd, I cannot reproduce this error here. I might have some time later to try to surface this again, sorry for the messyness..

maxrjones:
FYI I think this will be fixed by #361. I didn't notice this PR until afterwards when I was looking for the TIFF draft implementation(s)

jbusecke:
Thanks @maxrjones. Ill close this then.

PR #206, opened 2024-07-31
Set ZArray default fill_value as NaT for datetime64

thodson-usgs:
This PR fixes an error introduced in 10bd53dc3dae08303e57fe5aefe49804d9c4517d, which broke the default fill_value for datetime64 fields. Closes #201.

To replicate the error:
```bash
aws s3api get-object --region us-west-2 --no-sign-request --bucket wrf-se-ak-ar5 --key ccsm/rcp85/daily/2060/WRFDS_2060-01-01.nc WRFDS_2060-01-01.nc
aws s3api get-object --region us-west-2 --no-sign-request --bucket wrf-se-ak-ar5 --key ccsm/rcp85/daily/2060/WRFDS_2060-01-02.nc WRFDS_2060-01-02.nc
```
then run this `example.py`
```python
import xarray as xr
import glob
from virtualizarr import open_virtual_dataset

virtual_datasets = [
    open_virtual_dataset(filepath,
                         indexes={},
                         loadable_variables=['Time'],
                         cftime_variables=['Time'])
    for filepath in glob.glob('*.nc')
]

virtual_ds = xr.combine_nested(virtual_datasets, concat_dim=["Time"], coords="minimal", compat="override")
virtual_ds.virtualize.to_kerchunk('combined.json', format='json')

ds = xr.open_dataset('combined.json', engine="kerchunk")
```

The PR implements Tom's suggestion below: https://github.com/zarr-developers/VirtualiZarr/pull/206#issuecomment-2260955421

PR #210, opened 2024-08-05
Implement pydantic models as dataclasses

### LINKED ISSUE #204: Nit: `__post_init__` vs pydantic `@model_validator(mode="after")`
[In the `ZArray` class](https://github.com/zarr-developers/VirtualiZarr/blob/179bb2ab42664546dd243a2e04fab8737846229a/virtualizarr/zarr.py#L76) then after #193 we now have multiple types of validation checks, which together look like this:

```python
from pydantic import (
    BaseModel,
    field_validator,
    model_validator,
)

class ZArray(BaseModel):
    ...

    @field_validator("dtype")
    @classmethod
    def validate_dtype(cls, dtype) -> np.dtype:
        # Your custom validation logic here
        # Convert numpy.dtype to a format suitable for Pydantic
        return np.dtype(dtype)

    def __post_init__(self) -> None:
        if len(self.shape) != len(self.chunks):
            raise ValueError(
                "Dimension mismatch between array shape and chunk shape. "
                f"Array shape {self.shape} has ndim={self.shape} but chunk shape {self.chunks} has ndim={len(self.chunks)}"
            )

    @model_validator(mode="after")
    def _check_fill_value(self) -> Self:
        if self.fill_value is None:
            self.fill_value = ZARR_DEFAULT_FILL_VALUE.get(self.dtype, 0.0)
        return self
```

It seems to me like having both `__post_init__` and `@model_validator(mode="after")` is unnecessarily complicated. I think `_check_fill_value` should just be rewritten using `@field_validator` like `validate_dtype` is. 

The check inside `__post_init__` can't be done using `field_validator` as it involves comparing two correctly-instatiated fields against one another. It is not clear to me whether best practice with pydantic is to use `__post_init__` or `@model_validator(mode="after")` - [both seem supported](https://docs.pydantic.dev/latest/concepts/dataclasses/#initialization-hooks), and to do similar things.

@ghidalgo3 is there some subtlety I've missed here or can I clean this up a bit? (I know eventually this should all be replaced by `ZarrV3Metadata` in #175.)

TomAugspurger:
This removes our pydantic dependency by reimplementing them with dataclasses. There are a few breaking changes:

1. The classes won't automatically cast the argumetns to the declared type. IMO, that's the preferable behavior. Some backwards compatability shims have been added for np.dtype, but perhpas we want to remove that too.
2. The models won't have any of the methods they previously inherited from pydantic.BaseModel. This is probably good for user-facing objects, we now have full control over the public API.
3. We had to reorder some of the fields on ZArray, since dataclasses is stricter about positional arguments. I've aligned the order with `zarr.create`.

The tests did need to be updated in a few places (we compare some strings, and the order changed, and we don't automatically cast lists to tuples in the init for shape and size).

- [x] Closes #204
- [ ] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -17,6 +17,9 @@ Breaking changes
 
 - Serialize valid ZarrV3 metadata and require full compressor numcodec config (for :pull:`193`)
   By `Gustavo Hidalgo <https://github.com/ghidalgo3>`_.
+- VirtualiZarr's `ZArray`, `ChunkEntry`, and `Codec` no longer subclasses `zarr.Array`, no longer inherit from `pydantic.BaseModel` (:pull:`xxx`)
+- `ZArray`'s `__init__` signature has changed to match `zarr.Array`'s (:pull:`xxx`)
+
 
 Deprecations
 ~~~~~~~~~~~~
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -17,7 +17,8 @@ Breaking changes
 
 - Serialize valid ZarrV3 metadata and require full compressor numcodec config (for :pull:`193`)
   By `Gustavo Hidalgo <https://github.com/ghidalgo3>`_.
-- VirtualiZarr's `ZArray`, `ChunkEntry`, and `Codec` no longer subclasses `zarr.Array`, no longer inherit from `pydantic.BaseModel` (:pull:`xxx`)
+- VirtualiZarr's `ZArray`, `ChunkEntry`, and `Codec` no longer subclass
+  `pydantic.BaseModel` (:pull:`210`)
 - `ZArray`'s `__init__` signature has changed to match `zarr.Array`'s (:pull:`xxx`)
 
 
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -54,7 +54,7 @@ class ChunkEntry:
         return (self.path, self.offset, self.length)
 
     def dict(self) -> ChunkDictEntry:
-        return dataclasses.asdict(self)
+        return ChunkDictEntry(dataclasses.asdict(self))
 
 
 class ChunkManifest:
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -1,10 +1,10 @@
+import dataclasses
 import json
 import re
 from collections.abc import Iterable, Iterator
 from typing import Any, Callable, Dict, NewType, Tuple, TypedDict, cast
 
 import numpy as np
-from pydantic import BaseModel, ConfigDict
 from upath import UPath
 
 from virtualizarr.types import ChunkKey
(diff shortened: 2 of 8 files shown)


virtualizarr/manifests/manifest.py
```diff
@@ -58,7 +54,7 @@ def to_kerchunk(self) -> tuple[str, int, int]:
         return (self.path, self.offset, self.length)
 
     def dict(self) -> ChunkDictEntry:
-        return ChunkDictEntry(path=self.path, offset=self.offset, length=self.length)
+        return dataclasses.asdict(self)
```
TomAugspurger, line 57:
```suggestion
        return ChunkDictEntry(dataclasses.asdict(self))
```

virtualizarr/zarr.py
```diff
@@ -133,28 +112,12 @@ def to_kerchunk_json(self) -> str:
...
         self,
-        chunks: Optional[tuple[int, ...]] = None,
-        compressor: Optional[dict] = None,
-        dtype: Optional[np.dtype] = None,
-        fill_value: Optional[float] = None,  # float or int?
-        filters: Optional[list[dict]] = None,  # type: ignore[valid-type]
-        order: Optional[Literal["C"] | Literal["F"]] = None,
-        shape: Optional[tuple[int, ...]] = None,
-        zarr_format: Optional[Literal[2] | Literal[3]] = None,
+        **kwargs: Any,
```
TomAugspurger, line 115: Do we want the list of fields that can be replaced here in the signature?

TomAugspurger:
This will cause a conflict with https://github.com/zarr-developers/VirtualiZarr/pull/206, so waiting on that to go first.

virtualizarr/zarr.py
ghidalgo3: As opposed to letting users call [dataclasses.replace](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace)? I think `ZArray` shouldn't have fields that aren't replaceable, are there any fields like that?

docs/releases.rst
```diff
@@ -17,6 +17,9 @@ Breaking changes
 
 - Serialize valid ZarrV3 metadata and require full compressor numcodec config (for :pull:`193`)
   By `Gustavo Hidalgo <https://github.com/ghidalgo3>`_.
+- VirtualiZarr's `ZArray`, `ChunkEntry`, and `Codec` no longer subclasses `zarr.Array`, no longer inherit from `pydantic.BaseModel` (:pull:`xxx`)
```
TomNicholas, line 20:
```suggestion
- VirtualiZarr's `ZArray`, `ChunkEntry`, and `Codec` no longer inherit from `pydantic.BaseModel` (:pull:`xxx`)
```

It never subclassed `zarr.Array`!

virtualizarr/zarr.py
TomNicholas: I think if `dataclasses.replace` exists already then explicitly writing the list of fields isn't a good enough reason to overwrite the default method.

TomNicholas: Oh no wait `dataclasses.replace` is a module-level function, not a method on the dataclass type. In that case I change my mind - we should list out the arguments and call `dataclasses.replace` from within the `ZArray.replace` method.

TomNicholas (commented):
This is great @TomAugspurger !

virtualizarr/zarr.py
TomAugspurger: > I think ZArray shouldn't have fields that aren't replaceable, are there any fields like that?

Just to be clear, the behavior here (and on `main`) is to return a new instance of `ZArray` with the same values as the old instance, but new values specified in the keyword arguments.

> we should list out the arguments

Sounds good. I think we'll be able to do that (we'll need to remember to keep the two signatures in step though).

TomAugspurger:
Planning to merge this tonight.

Issue #209, opened 2024-08-04
Issues with mypy / other CI checks


TomAugspurger:
I think our static type checking configuration might not be 100% correct.

1. The `mypy` configuration in pyproject.toml at https://github.com/zarr-developers/VirtualiZarr/blob/b34a1ee47572901ed6a3623c97d340157a0fa62a/pyproject.toml#L75 must be under a `tool.mypy` prefix, not `mypy` when you're using pyproject.toml
2. I think that the mypy configuration in `.pre-commit-config.yaml` isn't , at least for my machine. Mine fails with `ERROR: No matching distribution found for types-pkg_resources`. We aren't using pkg-resources, so we shouldn't need that.
3. IMO, it might make sense to move mypy checks out of pre-commit and to a separate check? I think that mypy requires that 3rd party dependencies be installed into the environment so that those boundaries are type checked. Seems wasteful to install all those deps for testing, and then again for type checking.
4. I think that pre-commit isn't being run in CI

I have a branch starting to fix these up.

Expanding on issue 3, on main I have two failures:

```
virtualizarr/zarr.py:98: error: Argument "zarr_format" to "ZArray" has incompatible type "int"; expected "Literal[2, 3]"  [arg-type]
virtualizarr/tests/test_zarr.py:43: error: Argument "filetype" to "open_virtual_dataset" has incompatible type "str"; expected "FileType | None"  [arg-type]
```

On a branch with that moved to my dev env (and a few modules suppressed, since they don't provide type hints), I get more

```
virtualizarr/zarr.py:137: error: Function "virtualizarr.zarr.ZArray.dict" is not valid as a type  [valid-type]
virtualizarr/zarr.py:137: note: Perhaps you need "Callable[...]" or a callback protocol?
virtualizarr/zarr.py:253: error: Unsupported operand types for / ("Path" and "Hashable")  [operator]
virtualizarr/zarr.py:289: error: Argument 1 to "list" has incompatible type "tuple[Hashable, ...]"; expected "Iterable[str]"  [arg-type]
virtualizarr/kerchunk.py:226: error: Argument 2 to "variable_to_kerchunk_arr_refs" has incompatible type "Hashable"; expected "str"  [arg-type]
virtualizarr/kerchunk.py:239: error: Argument 1 to "join" of "str" has incompatible type "list[Hashable]"; expected "Iterable[str]"  [arg-type]
[...]

PR #213, opened 2024-08-05
Update static typing

TomAugspurger:
This updates how we do static type checking, and fixes some issues on `main`.

The biggest change is moving the `mypy` linting check out of `.pre-commit`. As mentioned in https://github.com/zarr-developers/VirtualiZarr/issues/209, we need *all* the typed dependencies in the environment running the check. The one managed by .pre-commit was lacking several dependencies (like xarray and Zarr) and so we weren't benefiting from static type checking at those boundaries. We can move it back into pre-commit, but we need to make sure that all our dependencies are duplicated there.

I'll explain some of the type changes inline.

- [x] Closes https://github.com/zarr-developers/VirtualiZarr/issues/209
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,10 +42,6 @@ jobs:
           conda env list
           conda list
 
-      - name: Lint
-        run: |
-          pre-commit run --all-files
-
       - name: Type check
         run: |
           mypy virtualizarr
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,6 +42,14 @@ jobs:
           conda env list
           conda list
 
+      - name: Lint
+        run: |
+          pre-commit run --all-files
+
+      - name: Type check
+        run: |
+          mypy virtualizarr
+
       - name: Running Tests
         run: |
           python -m pytest ./virtualizarr --run-network-tests --cov=./ --cov-report=xml --verbose
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,18 +34,20 @@ dependencies = [
 [project.optional-dependencies]
 test = [
     "codecov",
-    "pre-commit",
-    "ruff",
-    "pytest-mypy",
-    "pytest-cov",
-    "pytest",
-    "pooch",
-    "scipy",
-    "netcdf4",
-    "fsspec",
-    "s3fs",
     "fastparquet",
-    "h5py"
+    "fsspec",
+    "h5py",
+    "mypy",
+    "netcdf4",
+    "pandas-stubs",
+    "pooch",
+    "pre-commit",
+    "pytest-cov",
+    "pytest-mypy",
+    "pytest",
+    "ruff",
+    "s3fs",
+    "scipy",
 ]
 
 
@@ -70,11 +72,21 @@ exclude = ["docs", "tests", "tests.*", "docs.*"]
 [tool.setuptools.package-data]
 datatree = ["py.typed"]
 
-
-
-[mypy]
+[tool.mypy]
 files = "virtualizarr/**/*.py"
 show_error_codes = true
+
+[[tool.mypy.overrides]]
+module = "fsspec.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "numcodecs.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "kerchunk.*"
+ignore_missing_imports = true
 
 [tool.ruff]
 # Same as Black.
--- a/virtualizarr/kerchunk.py
+++ b/virtualizarr/kerchunk.py
@@ -303,7 +303,7 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
             shape=np_arr.shape,
             dtype=np_arr.dtype,
             order="C",
-            fill_value=None
+            fill_value=None,
         )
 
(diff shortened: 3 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -42,6 +42,14 @@ jobs:
           conda env list
           conda list
 
+      - name: Lint
```
TomAugspurger, line 45: This can maybe be removed? I guess the pre-commit CI thing runs it.

pyproject.toml
```diff
@@ -34,18 +34,20 @@ dependencies = [
 [project.optional-dependencies]
 test = [
     "codecov",
+    "fastparquet",
```
TomAugspurger, line 37: Sorry, mixed two changes here (sorting, and adding some additional dependencies like pandas-stubs and mypy)

virtualizarr/kerchunk.py
```diff
@@ -223,7 +224,7 @@ def dataset_to_kerchunk_refs(ds: xr.Dataset) -> KerchunkStoreRefs:
 
     all_arr_refs = {}
     for var_name, var in ds.variables.items():
-        arr_refs = variable_to_kerchunk_arr_refs(var, var_name)
+        arr_refs = variable_to_kerchunk_arr_refs(var, str(var_name))
```
TomAugspurger, line 227: All of these `str` conversions need to be looked at more closely. We do a lot of stuff with `Dataset.variables` / `.items` / `.coords` / etc. In xarray, `Variable.name` is just `Hashable`. Lots of types in virtualzarr are `str` (which is stricter, since `str` is hashable, but not all `Hashable` are str).


We might be able to loosen this a bit, but there's probably no benefit to.

virtualizarr/kerchunk.py
```diff
@@ -302,6 +303,7 @@ def variable_to_kerchunk_arr_refs(var: xr.Variable, var_name: str) -> KerchunkAr
             shape=np_arr.shape,
             dtype=np_arr.dtype,
             order="C",
+            fill_value=None
```
TomAugspurger: `fill_value` is currently required. My other PR is changing that to be a default of None.

virtualizarr/manifests/array.py
```diff
@@ -131,7 +131,7 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs) -> Any:
             return _isnan(self.shape)
         return NotImplemented
 
-    def __array__(self) -> np.ndarray:
+    def __array__(self, dtype: np.typing.DTypeLike = None) -> np.ndarray:
```
TomAugspurger, line 134: Have to match the base class's signature (but we raise NotImplementedError anyway, so it doesn't matter)

virtualizarr/manifests/manifest.py
```diff
@@ -57,7 +57,7 @@ def to_kerchunk(self) -> tuple[str, int, int]:
         """Write out in the format that kerchunk uses for chunk entries."""
         return (self.path, self.offset, self.length)
 
-    def dict(self) -> ChunkDictEntry:
+    def dict(self) -> ChunkDictEntry:  # type: ignore
```
TomAugspurger, line 60: I... don't know why I ignored this. I'll double check.

virtualizarr/tests/test_zarr.py
```diff
@@ -40,7 +41,7 @@ def isconfigurable(value: dict) -> bool:
 def test_zarr_v3_roundtrip(tmpdir, vds_with_manifest_arrays: xr.Dataset):
     vds_with_manifest_arrays.virtualize.to_zarr(tmpdir / "store.zarr")
     roundtrip = open_virtual_dataset(
-        tmpdir / "store.zarr", filetype="zarr_v3", indexes={}
+        str(tmpdir / "store.zarr"), filetype=FileType.zarr_v3, indexes={}
```
TomAugspurger, line 44: Not sure why the `str` was needed, will double check.

virtualizarr/xarray.py
```diff
@@ -265,7 +281,7 @@ def virtual_vars_from_kerchunk_refs(
     refs: KerchunkStoreRefs,
     drop_variables: list[str] | None = None,
     virtual_array_class=ManifestArray,
-) -> Mapping[str, xr.Variable]:
+) -> dict[str, xr.Variable]:
```
TomAugspurger, line 283: For these, it's probably better to mark these as returning a `dict` (since that's what we return).

In general, I think the guidance is to be broad with what you accept (so something like a `Mapping`) and narrow in what you return.

.github/workflows/main.yml
norlandrhagen: I think the pre-commit CI will run without this.

virtualizarr/kerchunk.py
TomNicholas: I'm not sure we really need to worry about this. It's a little used feature in xarray at all, and if netCDF / zarr groups don't support `Hashable` then it won't be relevant for VirtualiZarr.

virtualizarr/manifests/manifest.py
TomAugspurger: Ah, the `dict()` method is incompatible with the base class. I've tightened up the ignore, but will remove it entirely in #210

TomAugspurger:
CI is passing now.

TomNicholas (approved):
This is great! I looked through but couldn't see any issues :)

PR #227, opened 2024-08-23
Add nbytes property

### LINKED ISSUE #167: What should nbytes return?
Xarray uses the optional property `.nbytes` to indicate the size of wrapped arrays. Currently we don't implement `.nbytes` on `ManifestArrays`, so xarray defaults to estimating the size as basically `arr.size *  arr.dtype.itemsize`. I.e. currently it returns what the full size of the dataset would be if you loaded every referenced chunk into memory at once. But does this make sense for an array that can never be loaded into memory?

There is another completely different size to consider - that of the in-memory representation of the references - see discussion in #104. This is a known fixed number, it's not lazy-loading, but it's much smaller than the current nbytes.

This latter number is what's relevant if you're trying to estimate RAM usage whilst manipulating references, so it's possibly related to the `__sizeof__` discussion in https://github.com/pydata/xarray/issues/5764.

TomNicholas:
```python
In [1]: import virtualizarr as vz

In [2]: vds = vz.open_virtual_dataset('air.nc', indexes={})

In [3]: vds.nbytes
Out[3]: 128

In [4]: vds
Out[4]: 
<xarray.Dataset> Size: 128B
Dimensions:  (time: 2920, lat: 25, lon: 53)
Coordinates:
    time     (time) float32 32B ManifestArray<shape=(2920,), dtype=float32, c...
    lon      (lon) float32 32B ManifestArray<shape=(53,), dtype=float32, chun...
    lat      (lat) float32 32B ManifestArray<shape=(25,), dtype=float32, chun...
Data variables:
    air      (time, lat, lon) int16 32B ManifestArray<shape=(2920, 25, 53), d...
Attributes:
    Conventions:  COARDS
    description:  Data is from NMC initialized reanalysis\n(4x/day).  These a...
    platform:     Model
    references:   http://www.esrl.noaa.gov/psd/data/gridded/data.ncep.reanaly...
    title:        4x daily NMC reanalysis (1948)
```

- [x] Closes #167
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


virtualizarr/manifests/manifest.py
```diff
@@ -223,6 +223,17 @@ def shape_chunk_grid(self) -> tuple[int, ...]:
...
         return f"ChunkManifest<shape={self.shape_chunk_grid}>"
 
+    @property
+    def nbytes(self) -> int:
+        """
+        Size required to hold these references in memory in bytes.
+
+        Note this is not the size of the referenced chunks if they were actually loaded into memory,
+        this is only the size of the pointers to the chunk locations.
+        If you were to load the data into memory it would be ~1e6x larger for 1MB chunks.
```
TomNicholas, line 233: Can probably be a little more precise than this, though it does depend on the length of the stored paths.

Issue #168, opened 2024-07-01
combine_by_coords with loaded cftime variable demotes dtype


TomNicholas:
I realized that `xr.combine_by_coords` should actually already work fine - if you are willing to load the relevant coordinates into memory (and therefore also have those values saved into the resultant references on-disk).

```python
In [1]: from virtualizarr import open_virtual_dataset

In [2]: import xarray as xr

In [3]: ds1 = open_virtual_dataset('air1.nc', loadable_variables=['time', 'lat', 'lon'], cftime_variables=['time'])

In [4]: ds2 = open_virtual_dataset('air2.nc', loadable_variables=['time', 'lat', 'lon'], cftime_variables=['time'])

In [5]: xr.combine_by_coords([ds2, ds1], coords='minimal', compat='override')
Out[5]: 
<xarray.Dataset> Size: 8MB
Dimensions:  (time: 2920, lat: 25, lon: 53)
Coordinates:
  * lat      (lat) float32 100B 75.0 72.5 70.0 67.5 65.0 ... 22.5 20.0 17.5 15.0
  * lon      (lon) float32 212B 200.0 202.5 205.0 207.5 ... 325.0 327.5 330.0
  * time     (time) float32 12kB 1.867e+06 1.867e+06 ... 1.885e+06 1.885e+06
Data variables:
    air      (time, lat, lon) int16 8MB ManifestArray<shape=(2920, 25, 53), d...
Attributes:
    Conventions:  COARDS
    description:  Data is from NMC initialized reanalysis\n(4x/day).  These a...
    platform:     Model
    references:   http://www.esrl.noaa.gov/psd/data/gridded/data.ncep.reanaly...
    title:        4x daily NMC reanalysis (1948)
```

the only issue with this result is that somehow the dtype of `time` has been changed from `datetime64[ns]` to `float32`.

---

This approach doesn't solve the original issue, but it might also be fine in a lot of cases. `xr.combine_by_coords` can only auto-order along dimensions that have a `1D` coordinate, and 1D variables are small, so if they are split across many files its likely that you wanted to include them in `loadable_variables` anyway.

_Originally posted by @TomNicholas in https://github.com/zarr-developers/VirtualiZarr/issues/18#issuecomment-2200498037_

TomNicholas:
@jsignell not sure if this is a bug with the loading of cftime variables, the implementation of `xr.combine_by_coords`, or somewhere else.

TomNicholas:
[...]

PR #232, opened 2024-08-26
Let Xarray handle `decode_times`

norlandrhagen:
- [x] (https://github.com/zarr-developers/VirtualiZarr/issues/168)
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


docs/usage.md
```diff
@@ -325,13 +325,15 @@ Loading variables can be useful in a few scenarios:
 
 ### CF-encoded time variables
 
-Notice that the `time` variable that was loaded above does not have the expected dtype. To correctly decode time variables according to the CF conventions (like `xr.open_dataset` does by default), you need to include them in an additional keyword argument `cftime_variables`:
+To correctly decode time variables according to the CF conventions, you need to pass `time` to `loadable_variables` and ensure the `decode_times` argument of `open_virtual_dataset` is set to True ( like the default behavior of `xr.open_dataset`, `decode_times` defaults to `True`).
```
TomNicholas, line 328: If that's the default behaviour of xarray.open_dataset then we should make it the default behaviour of open_virtual_dataset too!

virtualizarr/tests/test_xarray.py
```diff
@@ -479,8 +479,22 @@ def test_mixture_of_manifestarrays_and_numpy_arrays(self, netcdf4_file):
...
+            "lat": np.arange(-90, 90, 10),
+            "lon": np.arange(-180, 180, 10),
+        },
+    )
+    ds.to_netcdf(f"{tmpdir}/tmp.nc")
+    vds = open_virtual_dataset(
+        f"{tmpdir}/tmp.nc", loadable_variables=["time", "lat", "lon"], indexes={}
+    )
+    # TODO use xr.testing.assert_identical(vds.indexes, ds.indexes) instead once class supported by assertion comparison, see https://github.com/pydata/xarray/issues/5812
+    assert index_mappings_equal(vds.xindexes, ds.xindexes)
```
TomNicholas, line 500: Should probably assert everything else is the same too!

virtualizarr/xarray.py
```diff
@@ -46,7 +45,7 @@ def open_virtual_dataset(
     filetype: FileType | None = None,
     drop_variables: Iterable[str] | None = None,
     loadable_variables: Iterable[str] | None = None,
-    cftime_variables: Iterable[str] | None = None,
```
TomNicholas, line 49: We might want to leave the argument in for now, and raise an informative error

TomNicholas (commented):
Nice. This makes a lot more sense than having the cftime_variables kwarg.

docs/usage.md
norlandrhagen: Looks like I was wrong! 
```decode_times ([bool](https://docs.python.org/3/library/functions.html#bool) or [dict-like](https://docs.python.org/3/glossary.html#term-mapping), optional) – If True, decode times encoded in the standard NetCDF datetime format into datetime objects.``` In this case, do you think we should we default to True or False?

Comment (TomNicholas) [reply docs/usage.md:328]: We should default to whatever Xarray defaults to.

Comment (TomNicholas) [reply docs/usage.md]: (Because once we build the VirtualiZarr Xarray backend, the kwargs will be one and the same.)

Comment (TomNicholas) [reply docs/usage.md]: (Because once we build the VirtualiZarr Xarray backend, the kwargs will be one and the same.)

File: virtualizarr/backend.py [hunk 4]
```diff
@@ -79,10 +79,9 @@ def open_virtual_dataset(
     loadable_variables: list[str], default is None
         Variables in the file to open as lazy numpy/dask arrays instead of instances of virtual_array_class.
         Default is to open all variables as virtual arrays (i.e. ManifestArray).
-    cftime_variables : list[str], default is None
-        Interpret the value of specified vars using cftime, returning a datetime.
-        These will be automatically re-encoded with cftime. This list must be a subset
-        of ``loadable_variables``.
+    decode_times: bool | None, default is None
+        Bool that is passed into Xarray's open_dataset. Allows time to be decoded into a datetime object.
+    cftime_variables: Iterable[str] | None = None
```
TomNicholas, line 84: Let's just remove this from the docstring entirely.

```suggestion
```

PR #233, opened 2024-08-27
Virtualizarr + Coiled Serverless Example Notebook

norlandrhagen:
Inspired by @thodson-usgs's [Lithops example](https://github.com/zarr-developers/VirtualiZarr/pull/203), I've created a Virtualizarr example using [coiled](https://www.coiled.io/) [serverless](https://docs.coiled.io/user_guide/functions.html#api) functions.  

- 1TB virtual dataset from 924 NetCDF files
- 9 minutes and ~$0.24 of cloud cost on coiled

Would love some feedback if anyone has thoughts.

image

### DIFF
--- a/examples/coiled/terraclimate.ipynb
+++ b/examples/coiled/terraclimate.ipynb
@@ -48,7 +48,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -56,8 +56,7 @@
     "import numpy as np\n",
     "import xarray as xr\n",
     "\n",
-    "from virtualizarr import open_virtual_dataset\n",
-    "from virtualizarr.backend import FileType"
+    "from virtualizarr import open_virtual_dataset"
    ]
   },
   {
@@ -73,7 +72,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


TomNicholas:
Awesome!!

I'm a bit unclear where the in-memory datasets live at each point in the computation when using coiled functions. You do the reference generation on a bunch of separate instances, but in order to to `combine_by_coords` they all have to be on the same instance. At what point does that transfer occur?

norlandrhagen:
> I'm a bit unclear where the in-memory datasets live at each point in the computation when using coiled functions. You do the reference generation on a bunch of separate instances, but in order to to `combine_by_coords` they all have to be on the same instance. At what point does that transfer occur?

I'm pretty sure the .map returns a generator of all the virtual datasets to my local laptop. 
Coiled allows you to run notebooks in the cloud. I successfully tried running all the reference generation serverless functions from a coiled cloud notebook and that also worked great. It might be a good option for a larger dataset, but I think since the manifest arrays are so memory efficient, I didn't see any issues doing the reduce on a laptop. 
I also tried starting the reference generation severless functions from a larger serverless function (that was meant to be the reduce machine) and it was kinda wacky.

examples/coiled/terraclimate.ipynb
```diff
@@ -0,0 +1,249 @@
...
+    "    idle_timeout=\"10 minutes\",\n",
+    "    n_workers=[100, 300],\n",
+    "    name=\"parallel_reference_generation\",\n",
+    ")\n",
+    "def process(filename):\n",
+    "    vds = open_virtual_dataset(\n",
+    "        filename,\n",
+    "        decode_times=True,\n",
+    "        loadable_variables=[\"time\", \"lat\", \"lon\", \"crs\"],\n",
+    "        filetype=FileType(\"netcdf4\"),\n",
```
TomNicholas, line 145: I dont' think you need to import `FileType`, you should just be able to pass `filetype='netcdf4'`.

TomNicholas (approved):

PR #235, opened 2024-08-31
Store test datasets in repo

### LINKED ISSUE #226: Store test datasets in this repo
Our current approach to testing involves a bunch of fixtures which each download a tutorial dataset from xarray (and cache it because it uses pooch), saves them to a temporary directory, then open that dataset from disk. This is not ideal for a few reasons:

1. The datasets aren't minimal, so they contain more complexity than is really needed to test a single bug / feature. This can make debugging more complicated.
2. We're using the network when we don't need to be.
3. `vz.open_virtual_dataset` calls `xr.open_dataset`, but because of our test setup `xr.open_dataset` can potentially be called more than once in the same test invocation, even if the code we are testing only calls it once. This again can make debugging more confusing than it needs to be.

We do need to test our ability to read files from disk, but it might be better just to make some really tiny netCDF files and save them in this repo.

norlandrhagen:
Adds a way to build test netcdf files and store them within the repo. 

- [x] Closes #226
- [x] Tests passing


norlandrhagen:
This test `test_numpy_arrays_to_inlined_kerchunk_refs` is failing.

```
refs["refs"]["lon/0"]
'base64:AABIQwCASkMAAE1DAIBPQwAAUkMAgFRDAABXQwCAWUMAAFxDAIBeQwAAYUMAgGNDAABmQwCAaEMAAGtDAIBtQwAAcEMAgHJDAAB1QwCAd0MAAHpDAIB8QwAAf0MAwIBDAACCQwBAg0MAgIRDAMCFQwAAh0MAQIhDAICJQwDAikMAAIxDAECNQwCAjkMAwI9DAACRQwBAkkMAgJNDAMCUQwAAlkMAQJdDAICYQwDAmUMAAJtDAECcQwCAnUMAwJ5DAACgQwBAoUMAgKJDAMCjQwAApUM='
```

TomNicholas:
Great idea! We could make the file even smaller by down sampling spatially presumably.

The failure shows that the variable is not being inlined when it previously was being inlined. This makes sense - which variables are inclined in the test is set by kerchunk's `inline_threshold` kwarg. These are then compared against the variables manually specified with `loadable_variables`. The parameter values (500.0 etc) were just chosen to be bigger/smaller than certain variables in the Xarray test dataset. Now you've changed those variables they will be different sizes, so some will be inclined that were not previously. It's a janky setup but I wasn't sure how to do it more neatly because that's the only way kerchunk allows you to control inlining.

norlandrhagen:
Thanks! Although, I can't take credit for the idea haha. 

Late night head scratcher, I can't seem to find the right `inline_threshold` to satisfy both the lat and the time assert statements. 😕 

```
# for loadable_vars = ['lat','lon']
# lat comparison fails below 101 inline_threshold
# time comparison fails above 16  inline_threshold
```

norlandrhagen:
mypy errors seem to be unrelated: https://github.com/zarr-developers/VirtualiZarr/issues/249

TomNicholas:
Looking at this again, I think storing the datasets as file is great for roundtrip tests, but we should also strive to make other tests start from a point that doesn't require reading netCDF (and hence doesn't rely on kerchunk). That could either be in-memory kerchunk references, or on-disk kerchunk references that we then intepret using #251, or maybe it could even be something simpler in some cases.

PR #243, opened 2024-10-02
Paths as URIs

### LINKED ISSUE #242: Forbid relative paths, and use file URI scheme internally?
We currently allow manifests to contain relative local paths, e.g. `test.nc`. This is (a) more fragile than an absolute path `/test.nc`, and (b) not very consistent with cloud bucket urls, which are always absolute.

It would be more robust to ensure that paths in the `Manifest` are always absolute paths. (If the user wants to move their data around they can always use the `.rename_paths` method to explicitly adjust the paths in their manifest to point to the files' new locations.)

If all local paths are to be stored as absolute paths then we also might want to use the [file URI scheme](https://en.wikipedia.org/wiki/File_URI_scheme), so that local paths are stored as `file:///test.nc`. That way in the `Manifest` the form of the path is consistent, whether it is local or remote. However as kerchunk does not use this scheme, it will mean extra conversion steps are needed to go between the two formats.

cc @mpiannucci

TomNicholas:
This PR closes #242 at the data model level - all paths are coerced to absolute URIs (i.e. `file:///directory/test.nc` or `s3://bucket/test.nc`) as they go into the `Manifest`. 

As this forbids constructing manifests using relative paths, it requires minor changes to many tests (e.g. `test.nc`-> `/test.nc`). It also will require slightly more invasive changes to any tests that involve kerchunk references.

- [x] Closes #242
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

Sub-tasks:
- [x] Validate paths in `.rename_paths` method automatically
- [x] Add `fs_root` option internally
- [x] Add `virtual_backend_kwargs` to all backends (see https://github.com/zarr-developers/VirtualiZarr/pull/315)
- [ ] ~~Deprecation warning for `reader_options` (rename to `fsspec_kwargs`?) (also see https://github.com/zarr-developers/VirtualiZarr/pull/315)~~
- [x] Publicly expose `fs_root` for `kerchunk` and `dmrpp` readers as an option to `virtual_backend_kwargs` (requires https://github.com/zarr-developers/VirtualiZarr/pull/315)
  - [x] kerchunk
  - [ ] ~~`dmrpp` EDIT: turns out I don't think it's necessary, at least not to get tests to pass~~
- [x] Test using `fs_root`
  - [x] For `kerchunk` reader (requires https://github.com/zarr-developers/VirtualiZarr/pull/317)
  - [ ] ~~For `dmrpp` reader (requires https://github.com/zarr-developers/VirtualiZarr/pull/318)~~
- [x] Document how to use `fs_root` for `kerchunk` ~~and `dmrpp` readers~~
- [x] Use `fs_root` automatically in other readers
  - [x] hdf5
  - [x] hdf
  - [x] netcdf3
  - [x] fits
  - [x] tiff
- [x] Test this actually fixes issue of writing relative paths to icechunk (https://github.com/earth-mover/icechunk/issues/279 and https://github.com/earth-mover/icechunk/issues/402)


TomNicholas:
What should we do with the paths in kerchunk references? Are they are always meant as absolute? I guess we should assume they are absolute, unless they have `./` or `../` at the start? Would fsspec ever produce relative paths like that?

cc @martindurant

martindurant:
> Are they are always meant as absolute?

They are always meant "as interpreted by the target filesystem". The nature of that filesystem might be implied by the protocol of a path alone, but commonly additional arguments are also required. This means, that relative paths do work if the target happens to be the local filesystem (file://), but I think of the other filesystems, only ssh supports this concept at all. I would not expect this to be meaningful for basically any practical case.

Note that the dir:// filesystem adds prefixes to URLs for any filesystem, if that's useful at all.

martindurant:
(I am happy to require absolute paths even if it makes some tests slightly more verbose)

TomNicholas:
Thanks @martindurant !

> The nature of that filesystem might be implied by the protocol of a path alone, but commonly additional arguments are also required.

But is the nature of the filesystem explicitly recorded in the kerchunk references format anywhere? Obviously if the prefix is explicit (e.g. `s3://`) then you know but otherwise?

> that relative paths do work

> > I guess we should assume they are absolute, unless they have ./ or ../ at the start?

Would this approach work then?

> (I am happy to require absolute paths even if it makes some tests slightly more verbose)

This might be helpful if the above approach doesn't work.

martindurant:
> is the nature of the filesystem explicitly recorded in the kerchunk references format

No. The original intention was to have these in the "templates", but in practice, the remote_protocol, remote_options and fss arguments to ReferenceFileSystem are used (and often encoded in Intake prescriptions) in cases of ambiguity.

virtualizarr/manifests/manifest.py
```diff
@@ -24,18 +34,48 @@ class ChunkDictEntry(TypedDict):
...
+    if not any(path.startswith(prefix) for prefix in VALID_URI_PREFIXES) and path != "":
+        # assume the path is local
+        try:
+            return str(Path(path).as_uri())
+        except ValueError as e:
+            if str(e) == "relative path can't be expressed as a file URI":
+                # problem is that path is relative instead of absolute, so add context to error message that this is forbidden
+                raise ValueError(
+                    f"paths in the manifest must be absolute, but got {path}"
+                ) from e
```
TomNicholas, line 52: This often gets caught by a try-except above it, leading to a less specific error being raised.

TomNicholas:
@sharkinsspatial and I discussed this earlier and we think that we need to add some kind of optional `fs_root` kwarg to specify the location of the root of the fsspec filesystem that the relative paths were created on. (This is the same thing as specifying what to prefix the relative paths with to turn them into absolute paths.)

The user could then do

```python
cwd = str(Path.cwd().as_uri())
vds = open_virtual_dataset('refs.json', reader_options={'fs_root': cwd})
```

which would prefix any relative paths found to be relative to the current working directly on this system.

If `fs_root` was not specified and a relative path found then an exception would be raised.

> I am happy to require absolute paths

Changing kerchunk upstream actually wouldn't really help as we still want to be able to parse kerchunk references that were created with older versions of kerchunk.

virtualizarr/manifests/manifest.py
```diff
@@ -1,65 +1,118 @@
...
+
+    See https://en.wikipedia.org/wiki/File_URI_scheme
+
+    Parameters
+    ----------
+    fs_root
+        The root of the filesystem on which these references were generated.
+        Required if any (likely kerchunk-generated) paths are relative in order to turn them into absolute paths (which virtualizarr requires).
+    """
+    if not any(path.startswith(prefix) for prefix in VALID_URI_PREFIXES) and path != "":
```
jsignell, line 91: I've seen people use urlparse (https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse). Not sure if that would be easier to read or not.

TomNicholas:
> You could try updating that and let me know if that fixes it.

I would expect that to fix it - what I'm wondering about is whether or not existing DMR++ files in the wild generally contain absolute or relative paths.

ayushnag:
I have seen both. Sometimes the file contains `s3://data.nc` but more commonly if the path is `s3://data.nc`, the file just contains `<Dataset xmlns="http://xml.opendap.org/ns/DAP/4.0#" ... name="data.nc">`

TomNicholas:
> the file just contains `<Dataset xmlns="http://xml.opendap.org/ns/DAP/4.0#" ... name="data.nc">`

And this does not tell me that the true path is an S3 path does it?

virtualizarr/manifests/manifest.py
TomNicholas: Actually handling all these cases is such a PITA that maybe I should just use [`cloudpathlib.AnyPath`](https://cloudpathlib.drivendata.org/stable/anypath-polymorphism/)

ayushnag:
> > the file just contains `<Dataset xmlns="http://xml.opendap.org/ns/DAP/4.0#" ... name="data.nc">`
> 
> And this does not tell me that the true path is an S3 path does it?

No, the true path is not known with only the dmrpp file contents

TomNicholas:
> No, the true path is not known with only the dmrpp file contents

Thanks. So I was right - this is equivalent to reading kerchunk json/parquet in that sense.

virtualizarr/manifests/manifest.py
TomNicholas: Using cloudpathlib did make this very easy, at the cost of adding a new dependency. Luckily for this use case (string validation) it's only a pure-python standalone dependency.

TomNicholas:
Okay important question: **Are we trying to support manifests with http URLs in them?** Right now we have some tests which create virtual datasets containing `http://` URLs as the path, e.g:

```
virtualizarr/tests/test_backend.py::TestReadFromURL::test_read_from_url[netcdf3-https://github.com/pydata/xarray-data/raw/master/air_temperature.nc-HDF5VirtualBackend]
```

But these tests (cc @scottyhq ) doesn't actually try to read the data back as loadable xarray variables. AFAIK Icechunk could not read data from a manifest containing http URLs (cc @mpiannucci ), but fsspec presumably could?

Handling this case in the manifest validation is possible but a bit annoying as cloudpathlib doesn't support http paths (yet https://github.com/drivendataorg/cloudpathlib/pull/468), and pathlib will incorrectly conclude that the http url is a relative posix path.

TomNicholas:
> AFAIK Icechunk could not read data from a manifest containing http URLs

It's planned for Icechunk to support HTTP apparently.

> Handling this case in the manifest validation is possible

I added support for HTTP in https://github.com/zarr-developers/VirtualiZarr/pull/243/commits/fefab909540ad790db811295c38e69f5b8190565

scottyhq:
> Are we trying to support manifests with http URLs in them? I added support for HTTP in https://github.com/zarr-developers/VirtualiZarr/commit/fefab909540ad790db811295c38e69f5b8190565 

If I'm following correctly, I'd say yes! There are lots of datasets out there that are not in cloud buckets, but are on servers that support http range requests. Agreed that it would be nice if cloudpathlib handled http:// paths

TomNicholas:
This PR now builds upon #318 FYI @ayushnag

virtualizarr/tests/test_readers/test_dmrpp.py
```diff
@@ -354,3 +372,38 @@ def test_parse_chunks(
...
+            str(basic_dmrpp_temp_filepath), indexes={}, filetype="dmrpp"
+        )
+        path = vds["x"].data.manifest["0"]["path"]
+
+        # by convention, if dmrpp file path is {PATH}.nc.dmrpp, the data filepath should be {PATH}.nc
+        # and the manifest should only contain absolute file URIs
+        expected_datafile_path_uri = basic_dmrpp_temp_filepath.as_uri().removesuffix(
+            ".dmrpp"
+        )
+        assert path == expected_datafile_path_uri
```
TomNicholas, line 392: @ayushnag currently I think the comments in this test are lying about what it actually tests. I had to change relative to absolute paths in the dmrpp strings above to get all the other tests to pass, but I don't know if that's valid dmrpp, and it means that this test is not testing using a dmrpp containing a relative datafilepath :/

ayushnag: I'm not sure why modifying the XML is needed since that "name" attribute is only used if you don't provide a datapath. However the DMRPPVirtualBackend [always provides a datapath](https://github.com/zarr-developers/VirtualiZarr/blob/515262493521a3d0a0659d8290ae28e0797faf47/virtualizarr/readers/dmrpp.py#L50) which is just the filepath the user passed into `open_virtual_dataset`. Basically the input to `open_virtual_dataset` minus `.dmrpp` should be the path that ends up in the manifest.

TomNicholas:
I got the DMR++ tests passing, so everything now passes locally (🥳) but I'm not sure if the changes to the DMR++ tests are okay (https://github.com/zarr-developers/VirtualiZarr/pull/243#discussion_r1861180719)...

Also the [CI fails](https://github.com/zarr-developers/VirtualiZarr/actions/runs/12057177693/job/33621226717?pr=243#step:6:820) because the new dependency on cloudpathlib wants to install S3 dependencies to instantiate `S3Path` objects, even though I'm not using `S3Path` to actually communicate with S3, I'm just using it as a convenience wrapper class. That's annoying and I'm not sure what the simplest solution is. The crux of the issue is that in VirtualiZarr a user could in theory want to manipulate S3 URLs in manifests without ever talking to S3. Some options:
1. Remove dependency on `cloudpathlib`, which would mean more parsing code needed in this PR,
2. See if `cloudpathlib` are amenable to delaying their import checks until `S3Path` is actually used to communicate with `S3`,
3. Change tests to not use S3 paths as examples when botocore is not installed,
4. Change CI to always have S3 dependencies installed, but that would require users to install S3 dependencies when they actually aren't even trying to talk to S3.

virtualizarr/tests/test_readers/test_dmrpp.py
TomNicholas: > However the `DMRPPVirtualBackend` [always provides a datapath](https://github.com/zarr-developers/VirtualiZarr/blob/515262493521a3d0a0659d8290ae28e0797faf47/virtualizarr/readers/dmrpp.py#L50) which is just the filepath the user passed into open_virtual_dataset

Right, but most of the tests in `test_dmrpp.py` don't actually use the `DMRPPVirtualBackend` class, they go straight to the `DMRParser`, so the `datapath` doesn't get provided (because its not even known at fixture-creation time). This is what I was originally trying to point out about the structure of these tests.

martindurant:
On cloudpathlib, is upath any use? Do you really need a wrapper?

TomNicholas:
> On `cloudpathlib`, is `upath` any use? 

I think that would have the same problem as `cloudpathlib` - I don't necessarily want these paths to have any connection to an actual storage layer, I just want to strictly validate and manipulate their forms.

> Do you really need a wrapper?

Turns out I didn't need a wrapper, I can get away with just more if...else logic for each of the possible types of paths. It's slightly less neat but it avoids the dependency issues.

PR #248, opened 2024-10-06
Zarr V3 metadata fixes

LDeakin:
This fixes several problems with `zarr.json` metadata that I noticed when [implementing a chunk manifest storage transformer](https://github.com/LDeakin/zarrs/pull/79).

- Chunk key encoding should be `v2` with `.` separator to match `manifest.json`
  - the `chunk-manifest-json` storage transformer should not need to be aware of the chunk key encoding
- Fix the fill value defaulting to `NaN` for integer arrays
- Encode non-finite float fill values as strings
- Default `bytes` codec `"endian"` to `"little"`
  - This needs to be addressed properly at some point
- Fix tests using `null` fill value or `nan` fill value for integer data type

### DIFF
--- a/virtualizarr/writers/zarr.py
+++ b/virtualizarr/writers/zarr.py
@@ -92,8 +92,8 @@ def zarr_v3_array_metadata(zarray: ZArray, dim_names: list[str], attrs: dict) ->
         "configuration": {"chunk_shape": metadata.pop("chunks")},
     }
     metadata["chunk_key_encoding"] = {
-        "name": "default",
-        "configuration": {"separator": "/"},
+        "name": "v2",
+        "configuration": {"separator": "."},
     }
     metadata["codecs"] = zarray._v3_codec_pipeline()
     metadata.pop("filters")
--- a/virtualizarr/tests/test_kerchunk.py
+++ b/virtualizarr/tests/test_kerchunk.py
@@ -23,7 +23,7 @@ def test_kerchunk_roundtrip_in_memory_no_concat():
             chunks=(2, 2),
             compressor=None,
             filters=None,
-            fill_value=np.nan,
+            fill_value=0,
             order="C",
         ),
         chunkmanifest=manifest,
--- a/virtualizarr/tests/test_manifests/test_array.py
+++ b/virtualizarr/tests/test_manifests/test_array.py
@@ -37,7 +37,7 @@ class TestManifestArray:
 
     def test_create_manifestarray_from_kerchunk_refs(self):
         arr_refs = {
-            ".zarray": '{"chunks":[2,3],"compressor":null,"dtype":"<i8","fill_value":null,"filters":null,"order":"C","shape":[2,3],"zarr_format":2}',
+            ".zarray": '{"chunks":[2,3],"compressor":null,"dtype":"<i8","fill_value":0,"filters":null,"order":"C","shape":[2,3],"zarr_format":2}',
             "0.0": ["test1.nc", 6144, 48],
         }
         marr = ManifestArray._from_kerchunk_refs(arr_refs)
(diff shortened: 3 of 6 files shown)


LDeakin:
Type-checking failures seem unrelated

virtualizarr/writers/zarr.py
```diff
@@ -92,8 +92,8 @@ def zarr_v3_array_metadata(zarray: ZArray, dim_names: list[str], attrs: dict) ->
         "configuration": {"chunk_shape": metadata.pop("chunks")},
     }
     metadata["chunk_key_encoding"] = {
-        "name": "default",
-        "configuration": {"separator": "/"},
+        "name": "v2",
```
TomNicholas, line 95: This seems wrong? For writing v3 metadata?

In general if we're not planning to use this format any more (see https://github.com/zarr-developers/VirtualiZarr/issues/262#issuecomment-2423427132), how much of this PR do you want to keep @LDeakin ?

TomNicholas: Presumably all the rest of the fixes are still relevant?

LDeakin: > This seems wrong? For writing v3 metadata?

The chunk manifest example in https://github.com/zarr-developers/zarr-specs/issues/287 and `virtualizarr` produces `"0.0"` style chunk key encoding, which is `v2` with `.` separator. `default` with `/` would be `"c/0/0"`.

If the chunk key encoding of the array and the chunk manifest matches, then the `chunk-manifest-json` storage transformer does not need to concern itself with chunk key encodings, which makes sense to me.

> In general if we're not planning to use this format any more (see [#262 (comment)](https://github.com/zarr-developers/VirtualiZarr/issues/262#issuecomment-2423427132)), how much of this PR do you want to keep @LDeakin ?

Not fussed, this PR was just the minimal changes I needed to use the `chunk-manifest-json` as currently spec'd and produced by `virtualizarr`. I'd hope most of these changes would be superseded by bringing in `zarr-python` V3 as a dependency anyway.

I haven't looked thoroughly at the spec for icechunk yet, but do you see it replacing `chunk-manifest-json` entirely? Can the time travel stuff be decoupled from the chunk manifests?

TomNicholas: > The chunk manifest example in https://github.com/zarr-developers/zarr-specs/issues/287 and virtualizarr produces "0.0" style chunk key encoding, which is v2 with . separator. default with / would be "c/0/0".

My intention was to test out writing to and reading from a v3-compatible json-based chunk manifest spec. If what I actually did looks more like v2 then that's my bad for not understanding the spec properly!

> Not fussed, this PR was just the minimal changes I needed to use the chunk-manifest-json as currently spec'd and produced by virtualizarr. I'd hope most of these changes would be superseded by bringing in zarr-python V3 as a dependency anyway.

Okay thanks. Maybe we get virtualizarr working fully, then look at the updated diff, as I would expect @mpiannucci's efforts on icechunk compatibility should iron out similar concerns around fill values?

> I'd hope most of these changes would be superseded by bringing in zarr-python V3 as a dependency anyway.

👍 We're close to being able to do that now that zarr-python v3 ~alpha~ (beta today actually) is out.

> I haven't looked thoroughly at the spec for icechunk yet, but do you see it replacing chunk-manifest-json entirely? 

I think that is Earthmover's intention.

> Can the time travel stuff be decoupled from the chunk manifests?

In theory it probably could, but in practice unless there is a strong use case for using chunk manifests where you wouldn't also like to have all the other features of icechunk, I'm not really sure why you would bother separting them. All the features of icechunk are closely-related in that they all involve/require adding a new layer of indirection into the store, i.e. the manifests + snapshots (which are kind of like time-stamped consolidated metadata IIUC). This question deserves discussion on that zarr spec proposal issue though.

TomNicholas: > This question deserves discussion on that zarr spec proposal issue though.

I've asked in https://github.com/zarr-developers/zarr-specs/issues/287#issuecomment-2424334666

maxrjones:
Hey @LDeakin, thanks for opening this PR. We've had a lot of recent changes in the library to support Zarr-Python V3. Another relevant change for your PR is that the Zarr writer was removed in https://github.com/zarr-developers/VirtualiZarr/pull/426 in favor of just supporting kercunk + icechunk.

I think the changes related to "Fix tests using null fill value or nan fill value for integer data type" are still relevant. Would you be interested in continuing this PR?

LDeakin:
@maxrjones oh wow this got quite stale! I'm happy for someone else to pick up fixing that particular issue

Issue #118, opened 2024-05-16
Open on-disk kerchunk references as a virtual dataset


TomNicholas:
It might be useful to be able to open an existing kerchunk json/parquet file as a virtual dataset, e.g to make changes to it before writing it back out.

This is essentially the kerchunk version of suggestion (2) here https://github.com/TomNicholas/VirtualiZarr/issues/63#issuecomment-2025968903.

This should be really easy to implement: We already have a function for doing it ([`dataset_from_kerchunk_refs`](https://github.com/TomNicholas/VirtualiZarr/blob/2d61d1aa2382c08f746e61cd2fe7c4c5cd5c769f/virtualizarr/xarray.py#L248)), we just have to teach `open_virtual_dataset` that existing kerchunk json/parquet files are also valid filetypes to pass in.

jsignell:
I can take a crack at this one.

jsignell:
Oh! I can back off :) I do think there is an argument to be made for having different methods for `open_virtual_dataset` and `open_as_virtual_dataset` or something depending on whether the input is kerchunk-style refs vs actual data.

TomNicholas:
> I do think there is an argument to be made for having different methods for `open_virtual_dataset` and `open_as_virtual_dataset` or something depending on whether the input is kerchunk-style refs vs actual data.

Yeah this is an interesting question. The same thing will arise for Zar stores too: should there be a different function to open zarr arrays backed by chunk manifests vs zarr arrays backed by actual bytes on-disk in the store? I think in that context it would be confusing to have two functions, especially as "mixed" zarr stores are possible (and useful).

jsignell:
> should there be a different function to open zarr arrays backed by chunk manifests vs zarr arrays backed by actual bytes on-disk in the store? I think in that context it would be confusing to have two functions, especially as "mixed" zarr stores are possible (and useful).

I would say it makes sense to my brain to have separate functions for "just reading" vs "doing work" so that I can form an expectation about how long something will take to run. But I would expect the `open_as_virtual_dataset` to accept either a legacy file or a kerchunk reference file or a zarr or a ref zarr.

Issue #249, opened 2024-10-07
[Tracking] Scheduled CI is failing on main


norlandrhagen:
It seems like the scheduled CI on `main` started failing around Oct 2nd. Opening this issue to track it. 

<img width="1415" alt="image" src="https://github.com/user-attachments/assets/68d6e724-fc75-45d2-a1db-aa15968879a3">

Seems to be mypy / type hinting related:

```python 

Run mypy virtualizarr
virtualizarr/manifests/manifest.py:137: error: Incompatible types in assignment (expression has type "ndarray[Any, dtype[Any]]", variable has type "ndarray[Any, StringDType]")  [assignment]
virtualizarr/manifests/manifest.py:144: error: Type argument "StringDType" of "dtype" must be a subtype of "generic"  [type-var]
virtualizarr/manifests/manifest.py:144: note: Error code "type-var" not covered by "type: ignore" comment
virtualizarr/manifests/manifest.py:199: error: Incompatible types in assignment (expression has type "ndarray[Any, dtype[StringDType]]", variable has type "ndarray[Any, StringDType]")  [assignment]
Found 3 errors in 1 file (checked 37 source files)
Error: Process completed with exit code 1.
```

norlandrhagen:
Closed by: https://github.com/zarr-developers/VirtualiZarr/pull/252#event-14613442064

PR #251, opened 2024-10-08
Allow `open_virtual_dataset` to read existing Kerchunk references

norlandrhagen:
## Allow `open_virtual_dataset` to read existing Kerchunk references as virtual datasets

- [x] Closes https://github.com/zarr-developers/VirtualiZarr/issues/118
- [x] Tests added
- [x] Tests passing

To Do:
- [x] [JSON] trailing `\\` in variable names. Fixed. `json` was adding trailing slashes on ref write.
- [x] [JSON] coordinates not recognized. Fixed. This was related to `json` write.
- [x] [JSON] coordinates set as data variables. Fixed. This was related to `json` write.
- [x] [PARQUET] extract references from 1. LazyReferenceMapper or 2. Reconstruct reference from :sparkles:_Zarrquet_ :sparkles: format.
- [x] [FUTURE PR] Convert `inlined` vars into numpy arrays. Note: After talking with @sharkinsspatial, it seems like the HDF reader he is working on won't inline at all. Should we: 1. Raise an error if any inline data exists or 2. Add logic to convert any inlined bytes to numpy arrays to maintain more compatibility with other Kerchunk readers?

Notes:
- Mypy checks in CI are currently disabled! https://github.com/zarr-developers/VirtualiZarr/issues/249
- The utility function `_fsspec_openfile_from_filepath` forced the fsspec filesystem to open a filepath (the fault of past me). I replaced it with a simple class that has a `.open_file` method that can be used as needed. The Kerchunk `LazyReferenceMapper` required a fsspec filesystem.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,9 +46,9 @@ jobs:
           conda env list
           conda list
 
-      # - name: Type check
-      #   run: |
-      #     mypy virtualizarr
+      - name: Type check
+        run: |
+          mypy virtualizarr
 
       - name: Running Tests
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,9 +46,9 @@ jobs:
           conda env list
           conda list
 
-      - name: Type check
-        run: |
-          mypy virtualizarr
+      # - name: Type check
+      #   run: |
+      #     mypy virtualizarr
 
       - name: Running Tests
         run: |
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -421,6 +421,17 @@ Currently there are not yet any zarr v3 readers which understand the chunk manif
 This store can however be read by {py:func}`~virtualizarr.xarray.open_virtual_dataset`, by passing `filetype="zarr_v3"`.
 ```
 
+## Opening kerchunk references from disk as virtual datasets
+
+You can open kerchunk references from disk as virtual datasets. This may be useful in appending workflows or creating checkpoints for larger datasets.
+
+```python
+vds = open_virtual_dataset('combined.json', format='kerchunk_json')
+# or
+vds = open_virtual_dataset('combined.parquet', format='kerchunk_parquet')
+
+```
+
 ## Rewriting existing manifests
 
 Sometimes it can be useful to rewrite the contents of an already-generated manifest or virtual dataset.
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -423,13 +423,14 @@ This store can however be read by {py:func}`~virtualizarr.xarray.open_virtual_da
 
 ## Opening Kerchunk references from disk as virtual datasets
 
-You can open kerchunk references from disk as virtual datasets. This may be useful in appending workflows or creating checkpoints for larger datasets.
+You can open `json` or `paruqet` Kerchunk references from disk as virtual datasets. This may be useful in appending workflows or creating checkpoints for larger datasets.
 
 ```python
 
-vds = open_virtual_dataset('combined.json', format='kerchunk_json')
+vds = open_virtual_dataset('combined.json', format='kerchunk')
 # or
-vds = open_virtual_dataset('combined.parquet', format='kerchunk_parquet')
+vds = open_virtual_dataset('combined.parquet', format='kerchunk')
+
 
 ```
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -421,16 +421,15 @@ Currently there are not yet any zarr v3 readers which understand the chunk manif
 This store can however be read by {py:func}`~virtualizarr.xarray.open_virtual_dataset`, by passing `filetype="zarr_v3"`.
 ```
 
-## Opening Kerchunk references from disk as virtual datasets
+## Opening Kerchunk references as virtual datasets
 
-You can open `json` or `paruqet` Kerchunk references from disk as virtual datasets. This may be useful in appending workflows or creating checkpoints for larger datasets.
+You can open existing Kerchunk `json` or `parquet` references as Virtualizarr virtual datasets. This may be useful for converting existing Kerchunk formatted references to storage formats like [Icechunk](https://icechunk.io/).
 
 ```python
 
 vds = open_virtual_dataset('combined.json', format='kerchunk')
 # or
 vds = open_virtual_dataset('combined.parquet', format='kerchunk')
-
 
 ```
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -421,11 +421,12 @@ Currently there are not yet any zarr v3 readers which understand the chunk manif
 This store can however be read by {py:func}`~virtualizarr.xarray.open_virtual_dataset`, by passing `filetype="zarr_v3"`.
 ```
 
-## Opening kerchunk references from disk as virtual datasets
+## Opening Kerchunk references from disk as virtual datasets
 
 You can open kerchunk references from disk as virtual datasets. This may be useful in appending workflows or creating checkpoints for larger datasets.
 
 ```python
+
 vds = open_virtual_dataset('combined.json', format='kerchunk_json')
 # or
 vds = open_virtual_dataset('combined.parquet', format='kerchunk_parquet')
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,8 @@ test = [
     "pytest",
     "ruff",
     "s3fs",
-    "scipy"]
+    "scipy",
+]
 
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,8 +46,7 @@ test = [
     "pytest",
     "ruff",
     "s3fs",
-    "scipy",
-]
+    "scipy"]
 
 
 [project.urls]
@@ -85,6 +84,10 @@ ignore_missing_imports = true
 
 [[tool.mypy.overrides]]
 module = "kerchunk.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "ujson.*"
 ignore_missing_imports = true
 
 [tool.ruff]
--- a/virtualizarr/backend.py
+++ b/virtualizarr/backend.py
@@ -141,26 +141,14 @@ def open_virtual_dataset(
         filetype = FileType(filetype)
 
     if filetype == FileType.kerchunk:
-        # add parquet path
-        from fsspec.implementations.reference import LazyReferenceMapper
-
         from virtualizarr.readers.kerchunk import dataset_from_kerchunk_refs
 
         fs = _FsspecFSFromFilepath(filepath=filepath, reader_options=reader_options)
 
-        # yuck, but JSON doesn't have magic bytes
-        if fs.filepath.endswith(".json"):
-            import ujson
-
-            from virtualizarr.readers.kerchunk import dataset_from_kerchunk_refs
-
-            fpath = fs.open_file()
-
-            refs = ujson.load(fpath)
-            return dataset_from_kerchunk_refs(KerchunkStoreRefs(refs))
-
-        # also yuck, but the .parquet isn't actually a parquet it's a directory that contains named parquets for each group.
+        # The kerchunk .parquet storage format isn't actually a parquet, but a directory that contains named parquets for each group/variable.
         if fs.filepath.endswith("ref.parquet"):
+            from fsspec.implementations.reference import LazyReferenceMapper
+
             lrm = LazyReferenceMapper(filepath, fs.fs)
 
             # build reference dict from KV pairs in LazyReferenceMapper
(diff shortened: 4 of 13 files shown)


keewis:
did you see #186? That was a second attempt after #119 to implement this, and contains some documentation that you might be able to reuse?

norlandrhagen:
@keewis thanks for the rec. Having some docs already is nice.

pyproject.toml
```diff
@@ -46,8 +46,7 @@ test = [
     "pytest",
     "ruff",
     "s3fs",
-    "scipy",
-]
+    "scipy"]
```
keewis, line 49:
```suggestion
    "scipy",
]
```

virtualizarr/backend.py
```diff
@@ -133,9 +137,39 @@ def open_virtual_dataset(
...
+    if filetype == FileType.kerchunk_json:
+        import ast
+
+        from virtualizarr.readers.kerchunk import dataset_from_kerchunk_refs
+
+        fpath = _FsspecFSFromFilepath(
+            filepath=filepath, reader_options=reader_options
+        ).open_file()
+
+        refs = ast.literal_eval(fpath.read().decode("utf-8"))
```
keewis, line 169: is there a specific reason for preferring `ast.literal_eval` over `json.load` / `json.loads` (or the `ujson` version of that)?

norlandrhagen: I think the Kerchunk reference isn't valid JSON, so reading using `ast` allowed me to eval it as a dict, but very open to suggestions!

keewis: as far as I remember the various `translate` methods of the `kerchunk` objects return python dictionaries, and the [usual pattern](https://fsspec.github.io/kerchunk/tutorial.html) is to explicitly dump those to json (not sure if anything changed since I last looked).

Given that `ast.literal_eval` is supposed to be 4-5 times slower than `json.load*` (the reference is python 3.3, so the exact number may have changed), I believe we should try to avoid using `ast.literal_eval`.

norlandrhagen: Good to know about performance hit! Thanks for the feedback @keewis , I got `ujson` working and `ast` replaced.

.github/workflows/main.yml
```diff
@@ -46,9 +46,9 @@ jobs:
           conda env list
           conda list
 
-      - name: Type check
-        run: |
-          mypy virtualizarr
+      # - name: Type check
+      #   run: |
+      #     mypy virtualizarr
```
TomNicholas, line 51: Try once #252 is in

```suggestion
      - name: Type check
        run: |
          mypy virtualizarr
```

virtualizarr/backend.py
```diff
@@ -39,6 +40,8 @@ class FileType(AutoName):
     zarr = auto()
     dmrpp = auto()
     zarr_v3 = auto()
+    kerchunk_json = auto()
+    kerchunk_parquet = auto()
```
TomNicholas, line 44: Could these be amalgamated into one reader? The "kerchunk" reader for the "kerchunk" references filetype? Then the reader itself detects if it's looking at a json or parquet?

virtualizarr/backend.py
```diff
@@ -133,9 +137,38 @@ def open_virtual_dataset(
...
 
     # if filetype is user defined, convert to FileType
+
     if filetype is not None:
         filetype = FileType(filetype)
 
+    if filetype == FileType.kerchunk_parquet:
+        from fsspec.implementations.reference import LazyReferenceMapper
+
+        from virtualizarr.readers.kerchunk import dataset_from_kerchunk_refs
```
TomNicholas, line 147: This is a change for a later PR but I think we want to tidy up these `if` statements to call a consistent API of some kind that's defined by each reader.

virtualizarr/backend.py
```diff
@@ -133,9 +137,38 @@ def open_virtual_dataset(
...
+
+        fs = _FsspecFSFromFilepath(filepath=filepath, reader_options=reader_options).fs
+        lrm = LazyReferenceMapper(filepath, fs)
+
+        # build reference dict from KV pairs in LazyReferenceMapper
+        array_refs = {k: lrm[k] for k in lrm.keys()}
+
+        # where does version come from? Is it always 1?
+        full_reference = {"version": 1, "refs": array_refs}
```
TomNicholas, line 156: Isn't the `version` key present in the refs themselves?

virtualizarr/backend.py
```diff
@@ -133,9 +137,38 @@ def open_virtual_dataset(
...
+
+        return dataset_from_kerchunk_refs(KerchunkStoreRefs(full_reference))
```
TomNicholas, line 158: Following the above comment - I guess `dataset_from_kerchunk_refs` should be a common utility that every reader that produces kerchunk-formatted references calls internally before returning a virtual dataset.

virtualizarr/tests/test_backend.py
```diff
@@ -336,3 +336,35 @@ def test_open_dataset_with_scalar(self, hdf5_scalar, tmpdir):
...
+
+        ref_filepath = tmp_path / "ref.parquet"
+        refs_to_dataframe(fo=example_reference_dict, url=ref_filepath.as_posix())
+    vds = open_virtual_dataset(
+        filepath=ref_filepath.as_posix(), filetype=reference_format, indexes={}
+    )
+
+    assert list(vds) == ["air"]
+    assert set(vds.coords) == set(["lat", "lon", "time"])
+    assert set(vds.variables) == set(["lat", "air", "lon", "time"])
```
TomNicholas, line 370: Shouldn't we be doing some more stringent assertions on top of these? i.e. checking that the actual manifest contain what we expect them to?

virtualizarr/utils.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import io
-from typing import TYPE_CHECKING, Optional, Union
+from typing import TYPE_CHECKING, Dict, Optional, Union
```
TomNicholas, line 4: You can just use the built-in `dict` can't you?

TomNicholas (commented):
Great work @norlandrhagen !

> Convert inlined vars into numpy arrays. Note: After talking with @sharkinsspatial, it seems like the HDF reader he is working on won't inline at all. Should we: 
> 1. Raise an error if any inline data exists or 
> 3. Add logic to convert any inlined bytes to numpy arrays to maintain more compatibility with other Kerchunk readers?

I feel like this mixes a few issues together. We do need to be able to read back inlined kerchunk references (though it's fine to add that feature in a follow-up PR), partly because even with Sean's PR we're still going to want to use the other kerchunk readers sometimes. I think it makes sense for @sharkinsspatial 's HDF reader not to create inlined refs, so long as we have another way to create inlined refs (i.e. using the normal xarray backend for that filetype).

> Mypy checks in CI are currently disabled! 

Should be fixed by #252

> The utility function `_fsspec_openfile_from_filepath` forced the fsspec filesystem to open a filepath

Not totally sure I understand this but abstracting away fsspec details sounds good.

virtualizarr/backend.py
norlandrhagen: Good idea!

virtualizarr/backend.py
norlandrhagen: Seems like a good standardization.

virtualizarr/backend.py
norlandrhagen: Wondering if you have ideas on the auto-detection. Initially I was thinking of using the 'magic bytes' strategy that is done in `_automatically_determine_filetype`, but the Kerchunk storage format isn't really parquet, but more like a zarr mimicking directory structure containing multiple parquets. 

```
tree ref.parquet
├── air
│   └── refs.0.parq
├── lat
│   └── refs.0.parq
├── lon
│   └── refs.0.parq
└── time
    └── refs.0.parq
```

TomNicholas: I'm not sure how much we can do here. In general we need to put a bit more thought into whether or not the auto-detection can generalize for when we have a system of externally-defined readers plugging into a virtualizarr entrypoint. I feel like at best we're still going to have a detection system that is often ambiguous...

virtualizarr/backend.py
norlandrhagen: Opened up an issue to track this: https://github.com/zarr-developers/VirtualiZarr/issues/257

virtualizarr/backend.py
norlandrhagen: Good catch. From what I understand** about the Kerchunk spec. `version 1` refers to the current "JSON" structure. The parquet version doesn't have a version number. 

**Could all be very wrong

virtualizarr/backend.py
```diff
@@ -133,9 +136,41 @@ def open_virtual_dataset(
...
+
+            from virtualizarr.readers.kerchunk import dataset_from_kerchunk_refs
+
+            fpath = fs.open_file()
+
+            refs = ujson.load(fpath)
+            return dataset_from_kerchunk_refs(KerchunkStoreRefs(refs))
+
+        # also yuck, but the .parquet isn't actually a parquet it's a directory that contains named parquets for each group.
+        if fs.filepath.endswith("ref.parquet"):
```
keewis, line 163: as far as I can tell, the structure is exactly that of `zarr`, but instead of chunk files you've got parquet files. Which means that groups would be subdirectories, and variables are subdirectories with the parquet files.

To be able to reliably detect the file format, maybe we should ask `kerchunk` to put a `version.json` or `kerchunk.json` file (or similar) into the root directory?

virtualizarr/backend.py
```diff
@@ -133,9 +136,41 @@ def open_virtual_dataset(
...
+    if filetype == FileType.kerchunk:
+        # add parquet path
+        from fsspec.implementations.reference import LazyReferenceMapper
+
+        from virtualizarr.readers.kerchunk import dataset_from_kerchunk_refs
+
+        fs = _FsspecFSFromFilepath(filepath=filepath, reader_options=reader_options)
+
+        # yuck, but JSON doesn't have magic bytes
+        if fs.filepath.endswith(".json"):
```
keewis, line 152: we won't be able to work around detecting the `json` extension, but the first couple of bytes should be something like `{"version": 1, "refs":` (plus whitespace). This won't work for `.json.zstd`, but if we truly want to support that we can add that in a future PR?

TomNicholas: I agree, let's punt on the auto-detection, and just get this in.

One reason we want this feature is to help people start moving their kerchunk references into Icechunk format, so it would be good to get this in now Icechunk is out.

virtualizarr/backend.py
norlandrhagen: Yeah, I've been calling the format **_zarrquet_** 🙃 

We could check for valid parquet, but I think we would have to know the name of the groups/variables 🤷 

That could be interesting! I think the `json` kerchunk format contains a `version` in the reference, but AFAICT, the parquet version does not have a version.

docs/usage.md
```diff
@@ -421,6 +421,19 @@ Currently there are not yet any zarr v3 readers which understand the chunk manif
 This store can however be read by {py:func}`~virtualizarr.xarray.open_virtual_dataset`, by passing `filetype="zarr_v3"`.
 ```
 
+## Opening Kerchunk references from disk as virtual datasets
+
+You can open `json` or `paruqet` Kerchunk references from disk as virtual datasets. This may be useful in appending workflows or creating checkpoints for larger datasets.
```
TomNicholas, line 426: Or converting kerchunk references into different references formats.

docs/usage.md
```diff
@@ -421,6 +421,19 @@ Currently there are not yet any zarr v3 readers which understand the chunk manif
...
+
+```python
+
+vds = open_virtual_dataset('combined.json', format='kerchunk')
+# or
+vds = open_virtual_dataset('combined.parquet', format='kerchunk')
+
```
TomNicholas, line 433: Does this whitespace serve any purpose?

TomNicholas (approved):

virtualizarr/backend.py
```diff
@@ -133,9 +136,41 @@ def open_virtual_dataset(
...
+            lrm = LazyReferenceMapper(filepath, fs.fs)
+
+            # build reference dict from KV pairs in LazyReferenceMapper
+            # is there a better / more preformant way to extract this?
+            array_refs = {k: lrm[k] for k in lrm.keys()}
+
+            full_reference = {"refs": array_refs}
+
+            return dataset_from_kerchunk_refs(KerchunkStoreRefs(full_reference))
+
```
keewis, line 173: what happens if the filetype is `"kerchunk"`, but neither the json nor the parquet check evaluate to true? Should we raise? As in, should the two `if`+`return` blocks be converted to `if`+`return`/`elif`+`return`/`else`+`raise`?

docs/usage.md
norlandrhagen: I'll update and fix my spelling of ~~paruqet~~ parquet

virtualizarr/backend.py
norlandrhagen: Good call. I'll add an Exception.

TomNicholas:
Thanks @norlandrhagen - happy to merge!

Issue #83, opened 2024-04-13
Executable code examples in docs


TomNicholas:
The main usage docs revolve around code examples, but these are currently just hard-coded in. It would be good to generate these results automatically, and even better if the docs would display the lovely xarray HTML repr instead of the boring string repr.

I'm not sure what the best way to set that up with myst-markdown is. Do I need to rewrite the `usage.md` page as a jupyter notebook?

maxrjones:
If we use mkdocs (#500), we can use https://pawamoy.github.io/markdown-exec/ to execute code blocks in markdown.

maxrjones:
The usage guide includes executed codeblocks after #701

PR #252, opened 2024-10-11
Fix mypy errors around numpy functions not being strictly type hinted

TomNicholas:
Hoping this fixes some of the typing errors @norlandrhagen flagged in #249

### DIFF
--- a/virtualizarr/manifests/array_api.py
+++ b/virtualizarr/manifests/array_api.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Callable, Iterable
+from typing import TYPE_CHECKING, Any, Callable, Iterable, cast
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


TomNicholas:
No idea what's going on with the `hdf5` dependency in the docs build

norlandrhagen:
> No idea what's going on with the `hdf5` dependency in the docs build

For building the docs, do we need all the deps in the `      - -e  "..[test]" ` [line](https://github.com/zarr-developers/VirtualiZarr/blob/47a5e8702e44f71bb355bcba0ff6214fe6d09d83/ci/doc.yml#L16C21-L16C23)?

TomNicholas:
> For building the docs, do we need all the deps in the      - -e  "..[test]" [line](https://github.com/zarr-developers/VirtualiZarr/blob/47a5e8702e44f71bb355bcba0ff6214fe6d09d83/ci/doc.yml#L16C21-L16C23)?

The docs don't actually dynamically run any examples (see https://github.com/zarr-developers/VirtualiZarr/issues/83), so almost certainly we don't.

TomNicholas:
This passes the new separate test so should be good to merge?

Issue #253, opened 2024-10-11
Make mypy a separate CI job


TomNicholas:
The mypy checks are currently part of the same CI jobs that the test suite runs are. It would be better to have the `mypy` test be separate, so that it's easier to see if a PR only "failed" due to mypy rather than due to actual runtime test failures.

cc @norlandrhagen

norlandrhagen:
I can take this one if you haven't started a PR @TomNicholas.

TomNicholas:
That would be great, thanks @norlandrhagen

norlandrhagen:
https://github.com/zarr-developers/VirtualiZarr/pull/254

PR #254, opened 2024-10-11
Makes mypy a seperate CI job

norlandrhagen:
- [x] Closes (https://github.com/zarr-developers/VirtualiZarr/issues/253)

Type checking is failing, but it's the [same three](https://github.com/zarr-developers/VirtualiZarr/issues/249).

- Should we keep the python version matrix for mypy? 
- Installing all of the deps listed in `test` seems pretty overkill just for some typing library stubs. Is there a better alternative?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,10 +46,6 @@ jobs:
           conda env list
           conda list
 
-      - name: Type check
-        run: |
-          mypy virtualizarr
-
       - name: Running Tests
         run: |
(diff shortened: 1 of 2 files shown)


TomNicholas (approved):

TomNicholas:
> Should we keep the python version matrix for mypy?

Nah.

> Installing all of the deps listed in test seems pretty overkill just for some typing library stubs. Is there a better alternative?

Is there any other way? I mean if that breaks surely our actual tests will also break?

PR #256, opened 2024-10-15
Add Icechunk Support

mpiannucci:
Adds the ability to write to an Icechunk store. Co developed with @TomNicholas. 

See https://github.com/earth-mover/VirtualiZarr/pull/1 for more information (ported from that branch)

- [ ] Closes #xxxx
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


norlandrhagen:
@mpiannucci happy to help out on the icechunk integration effort. Are there bits you think that might good to split off and tackle? CI, docs, testing etc?

TomNicholas:
Docs could definitely be added in a separate PR. Also the whole problem of saving non-virtual references to icechunk.

IIUC the issues preventing this PR from going in are more to do with upstream dependencies (Zarr, kerchunk).

mpiannucci:
Correct. I am going to be working on kerchunk, but that also has some zarr upstream dependencies. Things are moving forward but will take a little time

norlandrhagen:
Totally makes sense! I'll hold off a bit until some things stabilize. Thanks again for all the work on this @mpiannucci, it'll be really cool to be able to _virtualize_ to icechunk.

TomNicholas:
@mpiannucci could you merge the changes from `main` into here please? I'm unsure how to handle some of the conflicts with `_v3_codec_pipeline`. `main` now includes a CI job that we can use to test icechunk with zarr-v3 and without kerchunk.

mpiannucci:
Im not sure if actions need to be updated or dependencies or whatever. From what i can tell, the CI fails for some type reasons, but also because Icechunk isntinstalled, and the zarr that is installed is zarr2, which makes sense.

TomNicholas:
> Icechunk isn't installed, and the zarr that is installed is zarr2

For the main tests yes, but the new CI job `upstream` should install all the [specific branches](https://github.com/zarr-developers/VirtualiZarr/blob/4b7612e5649058475223cdfbb1bf50910b70ff5b/ci/upstream.yml#L27) that Icechunk needs. Looks like those were installed correctly, and there were only a small number of [test failures](https://github.com/zarr-developers/VirtualiZarr/actions/runs/11441315284/job/31829095601?pr=256) on that branch

```python
FAILED virtualizarr/tests/test_integration.py::test_kerchunk_roundtrip_in_memory_no_concat - AssertionError: Left and right Dataset objects are not equal


Differing data variables:
L   a        (x, y) int64 64B ManifestArray<shape=(2, 4), dtype=int64, chunks...
R   a        (x, y) int64 64B ManifestArray<shape=(2, 4), dtype=int64, chunks...
FAILED virtualizarr/tests/test_readers/test_kerchunk.py::test_dataset_from_df_refs - AssertionError: assert 0 is nan
 +  where 0 = ZArray(shape=(2, 3), chunks=(2, 3), dtype=dtype('int64'), fill_value=0, order='C', compressor=None, filters=None, zarr_format=2).fill_value
 +    where ZArray(shape=(2, 3), chunks=(2, 3), dtype=dtype('int64'), fill_value=0, order='C', compressor=None, filters=None, zarr_format=2) = ManifestArray<shape=(2, 3), dtype=int64, chunks=(2, 3)>.zarray
 +      where ManifestArray<shape=(2, 3), dtype=int64, chunks=(2, 3)> = <xarray.DataArray 'a' (x: 2, y: 3)> Size: 48B\nManifestArray<shape=(2, 3), dtype=int64, chunks=(2, 3)>\nDimensions without coordinates: x, y.data
 +  and   nan = np.nan
FAILED virtualizarr/tests/test_writers/test_zarr.py::test_zarr_v3_metadata_conformance - TypeError: Object of type BytesCodec is not JSON serializable
FAILED virtualizarr/tests/test_writers/test_zarr.py::test_zarr_v3_roundtrip - TypeError: Object of type BytesCodec is not JSON serializable
FAILED virtualizarr/tests/test_writers/test_zarr.py::test_metadata_roundtrip - TypeError: Object of type BytesCodec is not JSON serializable
============= 5 failed, 51 passed, 66 skipped, 9 warnings in 1.18s =============
```

This looks fixable to me. 

We are missing Icechunk from that environment though - we should add installation via pip to that conda env in a new commit on this branch.

Then we wouldn't have our main CI passing, but we would have some CI jobs that test virtualizarr + zarr-python v2 + kerchunk, and a different CI job that tests virtualizarr + zarr-python v3 + icechunk, which is the best we can do until kerchunk supports v3.

mpiannucci:
Also Closes #193

virtualizarr/writers/icechunk.py
```diff
@@ -0,0 +1,179 @@
...
+    Write an xarray dataset whose variables wrap ManifestArrays to an Icechunk store.
+
+    Currently requires all variables to be backed by ManifestArray objects.
+
+    Parameters
+    ----------
+    ds: xr.Dataset
+    store: IcechunkStore
+    """
+    from icechunk import IcechunkStore
```
TomNicholas, line 38: I wonder if we should add a warning here that points people to the Earthmover docs describing all the dependencies that are required

virtualizarr/writers/icechunk.py
```diff
@@ -0,0 +1,179 @@
...
+        write_virtual_variable_to_icechunk(
+            store=store,
+            group=group,
+            name=name,
+            var=var,
+        )
+    else:
+        # TODO is writing loadable_variables just normal xarray ds.to_zarr?
+        # raise NotImplementedError()
+        print("skipping non-virtual variable", name)
```
TomNicholas, line 94: I think I would rather that this raised. Ideally we would raise a single message with the list of all non-virtual variables, telling the user that they can drop those from this datasets and write them into icechunk separately (even as part of the same commit).

In fact, couldn't we just make this work fairly easily using something like

```python
def write_variables_to_icechunk_group(
    variables,
    store,
    group,
):
    virtual_variables = {name: var for var in variables if isinstance(var.data, ManifestArray)}

    for name, var in virtual_variables.items():
        write_virtual_variables_to_icechunk(
            store=store,
            group=group,
            name=name,
            var=var,
        )

    loadable_variables = {name: var for var in variables if name not in virtual_variables}
    
    ds = xr.Dataset(loadable_variables)
    ds.to_zarr(icechunkstore)
```

mpiannucci: hmmmm that might indeed work. I switched to pass because raise would mean my test datasets didnt work (i never considered i could drop!! duh!).

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -0,0 +1,184 @@
...
+
+# TODO test writing grids of multiple chunks
+
+# TODO test writing to a group that isn't the root group
+
+# TODO test writing loadable variables
+
+# TODO roundtripping tests - requires icechunk compatibility with xarray
+
+# TODO test with S3 / minio
```
TomNicholas, line 184: We should still do all of these - else we're relying a bit too much on the upstream test in icechunk. But as we're pretty confident these work we can leave them to a follow-up PR.

virtualizarr/writers/icechunk.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # loop over every reference in the ChunkManifest for that array
+    # TODO inefficient: this should be replaced with something that sets all (new) references for the array at once
+    # but Icechunk need to expose a suitable API first
+    it = np.nditer(
+        [manifest._paths, manifest._offsets, manifest._lengths],  # type: ignore[arg-type]
+        flags=[
+            "refs_ok",
+            "multi_index",
+            "c_index",  # TODO is "c_index" correct? what's the convention for zarr chunk keys?
```
TomNicholas, line 155: If we add any test that this generates the correct result for any variable with more than one chunked dimension then we can remove this comment.

TomNicholas:
We should expose a `.to_icechunk` method on the accessor, but perhaps we should wait until a follow-up docs PR to do that.

mpiannucci:
Accessor and loadable variables work, although there is one little issue with xarray overwriting attributes of root group to figure out m

virtualizarr/writers/icechunk.py
```diff
@@ -0,0 +1,194 @@
+from typing import TYPE_CHECKING, cast
+
+import numpy as np
+from xarray import Dataset
+from xarray.backends.zarr import encode_zarr_attr_value
+from xarray.core.variable import Variable
+from zarr import Group  # type: ignore[import-untyped]
```
TomNicholas, line 7: Currently virtualizarr doesn't directly use zarr, it only needs it if either kerchunk or icechunk is installed. The `min-deps` CI is failing because that doesn't use either, and therefore doesn't install `zarr`. So this import should be protected.

virtualizarr/writers/icechunk.py
```diff
@@ -0,0 +1,194 @@
...
+from virtualizarr.zarr import encode_dtype
+
+if TYPE_CHECKING:
+    from icechunk import IcechunkStore  # type: ignore[import-not-found]
+
+
+VALID_URI_PREFIXES = {
+    "s3://",
+    # "gs://",
+    # "azure://",
```
TomNicholas, line 19:
```suggestion
    # "gs://",  # https://github.com/earth-mover/icechunk/issues/265
    # "azure://",  # https://github.com/earth-mover/icechunk/issues/266
```

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -0,0 +1,287 @@
...
+    array = root_group["foo"]
+
+    # check chunk references
+    # TODO we can't explicitly check that the path/offset/length is correct because icechunk doesn't yet expose any get_virtual_refs method
+
+    expected_ds = open_dataset(simple_netcdf4)
+    expected_array = expected_ds["foo"].to_numpy()
+    npt.assert_equal(array, expected_array)
+
+    # ds = open_zarr(store=icechunk_filestore, group='foo', zarr_format=3, consolidated=False)
```
TomNicholas, line 113: Shouldn't we enable this? If we already have to use the compatibility branch for xarray then we should test using xarray no?

virtualizarr/writers/icechunk.py
mpiannucci: Yup this was a mistake. An auto import from the ide. Thanks for catching

virtualizarr/accessor.py
```diff
@@ -39,6 +44,18 @@ def to_zarr(self, storepath: str) -> None:
         """
         dataset_to_zarr(self.ds, storepath)
 
+    def to_icechunk(self, store: "IcechunkStore") -> None:
+        """
+        Write an xarray dataset whose variables wrap ManifestArrays to an Icechunk store.
+
+        Currently requires all variables to be backed by ManifestArray objects.
```
TomNicholas, line 51:
```suggestion
        Write an xarray dataset to an Icechunk store.

        Any variables backed by ManifestArray objects will be be written as virtual references, any other variables will be loaded into memory before their binary chunk data is written into the store.
```

mpiannucci:
Okay thinnkkkkk were finally solid

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -0,0 +1,292 @@
...
+
+    # check chunk references
+    # TODO we can't explicitly check that the path/offset/length is correct because icechunk doesn't yet expose any get_virtual_refs method
+
+    expected_ds = open_dataset(simple_netcdf4)
+    expected_array = expected_ds["foo"].to_numpy()
+    npt.assert_equal(array, expected_array)
+
+    ds = open_zarr(store=icechunk_filestore, zarr_format=3, consolidated=False)
+    assert np.allclose(ds.foo.to_numpy(), expected_ds.foo.to_numpy())
```
TomNicholas, line 114: This would test also variable names, dimension names, indexes, and attrs
```suggestion
    import xarray.testing as xrt

    assert xrt.assert_identical(ds, expected_ds)
```

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -0,0 +1,292 @@
...
+    # check chunk references
+    # TODO we can't explicitly check that the path/offset/length is correct because icechunk doesn't yet expose any get_virtual_refs method
+
+    expected_ds = open_dataset(netcdf4_file)
+    expected_air_array = expected_ds["air"].to_numpy()
+    npt.assert_equal(scaled_air_array, expected_air_array)
+
+    # Check with xarray
+    ds = open_zarr(store=icechunk_filestore, zarr_format=3, consolidated=False)
+    assert np.allclose(ds.air.to_numpy(), expected_ds.air.to_numpy())
```
TomNicholas, line 175: `assert_identical` here too

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -0,0 +1,292 @@
...
+    assert air_array.attrs["units"] == "km"
+    assert np.allclose(air_array[:], la_v[:])
+
+    pres_array = root_group["pres"]
+    assert isinstance(pres_array, Array)
+    assert pres_array.shape == (3, 4)
+    assert pres_array.dtype == np.dtype("int32")
+    expected_ds = open_dataset(simple_netcdf4)
+    expected_array = expected_ds["foo"].to_numpy()
+    npt.assert_equal(pres_array, expected_array)
```
TomNicholas, line 285: This can also all be checked by doing `open_zarr` then using a single xarray assertion. But it might be valuable to keep these lower-level assertions too.

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+# TODO test writing to a group that isn't the root group
+
+# TODO roundtripping tests - requires icechunk compatibility with xarray
```
TomNicholas, line 290: If we have test that write a ds to icechunk, then open it with xarray and check the result is identical, then we have completed this TODO.
```suggestion
# TODO roundtripping tests - requires icechunk compatibility with xarray
```

virtualizarr/writers/icechunk.py
```diff
@@ -0,0 +1,203 @@
...
+
+    loadable_variables = {
+        name: var for name, var in variables.items() if name not in virtual_variables
+    }
+
+    # First write all the non-virtual variables
+    if len(loadable_variables) > 0:
+        # NOTE: We must set the attributes of the group before writing the dataset because the dataset
+        # will overwrite the root group's attributes with the dataset's attributes.
+        # TODO: Is this a sign that something is incorrect?
```
TomNicholas, line 87: That seems like expected behaviour to me? If we did `ds.to_zarr` even if `loadable_variables={}` then that line could double as us setting group attributes.

TomNicholas (commented):
I'm psyched about this! Just some last few commments.

TomNicholas:
Also add a note to `docs/release.rst` please! That would be a good place to mention kerchunk version weirdness too.

mpiannucci:
There is some weirdness when comparing the icechunk datasets using the testing tools that dont happen if you check the individual peices of the dataset. Leaving further refinement for another PR

PR #259, opened 2024-10-17
Skip tests that require kerchunk

### LINKED ISSUE #258: Make kerchunk dependency entirely optional
Virtualizarr doesn't need to have an explicit dependency on Kerchunk. Kerchunk is one (well it reads many formats) amongst many readers for creating virtual references, and it is now one of two options for writing virtual references (the other being Icechunk). Issue https://github.com/zarr-developers/VirtualiZarr/issues/78 points this out for the readers, but we could make the entire package only optionally depend on kerchunk.

A more urgent reason to change this is Kerchunk does not currently support zarr-python v3 (https://github.com/fsspec/kerchunk/pull/516), which is preventing us ensuring that the rest of this package works with zarr v3 (https://github.com/zarr-developers/VirtualiZarr/pull/182), which is needed for testing Icechunk compatibility (https://github.com/zarr-developers/VirtualiZarr/pull/256). 

Changing the virtualizarr business logic to make kerchunk optional is easy - the hard part is that lots of our tests are more reliant on kerchunk than they need to be. Ideally we would either rewrite those integration tests to start from in-memory references rather than files on disk, or use a non-kerchunk approach to generate references from example HDF5 files (https://github.com/zarr-developers/VirtualiZarr/pull/87). I think the former would be neater, but alternatively once the #87 is merged the latter would be a simple change.

@mpiannucci @norlandrhagen @ghidalgo3 @sharkinsspatial

TomNicholas:
All I've done here is skip all of those that currently require kerchunk, and add a new CI job to check that the others pass without kerchunk installed. I'm hoping that we can merge this PR then incrementally re-write all of the tests (apart from those in `test_readers/test_kerchunk.py` & `test_writers/test_kerchunk.py`) to run and test the same functionality overall without requiring kerchunk.

- [x] Closes #258


.github/workflows/min-deps.yml
```diff
@@ -0,0 +1,60 @@
+name: min-deps
```
TomNicholas, line 1: There was probably a better way to do this than just making an entire extra workflow file but my github-actions-fu is weak

pyproject.toml
```diff
@@ -22,20 +22,20 @@ requires-python = ">=3.10"
 dynamic = ["version"]
 dependencies = [
     "xarray>=2024.06.0",
-    "kerchunk>=0.2.5",
-    "h5netcdf",
     "numpy>=2.0.0",
-    "ujson",
     "packaging",
     "universal-pathlib",
+    "numcodecs",
```
TomNicholas, line 28: The minimum set of dependencies should look like this, and you could probably even make `universal-pathlib` optional for reading from non-remote files...

TomNicholas: Actually I re-added `ujson`, because we need it for parsing existing kerchunk-formatted references.

ci/min-deps.yml
norlandrhagen, line 1: Probably for another discussion, but could we have all deps and optional deps in pyproject.toml then the optional `h5py` in a conda env?

Or maybe it's a good reason to transition to Pixi 🥇

pyproject.toml
norlandrhagen: It might be nice to have more splits in the optional deps depending on the usecase, kind of like [how Xarray handles it](https://github.com/pydata/xarray/blob/b9780e7a32b701736ebcf33d9cb0b380e92c91d5/pyproject.toml#L52). We could have splits for different readers, writers and local/cloud usecases?

virtualizarr/tests/test_backend.py
```diff
@@ -13,9 +12,17 @@
 from virtualizarr.backend import FileType
 from virtualizarr.manifests import ManifestArray
 from virtualizarr.readers.kerchunk import _automatically_determine_filetype
-from virtualizarr.tests import has_astropy, has_tifffile, network, requires_s3fs
+from virtualizarr.tests import (
```
norlandrhagen, line 15: This is a great addition! Thanks for implementing it.

norlandrhagen (approved):
Looks great @TomNicholas!

ci/min-deps.yml
TomNicholas: I know nothing about pixi, but would be open to that.

pyproject.toml
TomNicholas: I was thinking the same, but kept this PR targeted for now.

virtualizarr/tests/test_backend.py
TomNicholas: Another good idea just stolen from xarray 😅

ci/min-deps.yml
norlandrhagen: @maxrjones or @cisaacstern might have some real experience with it, but one really cool seeming thing is that you can have both conda and pip deps all in a pyproject.toml

virtualizarr/tests/test_backend.py
norlandrhagen: The source of all good code

PR #260, opened 2024-10-17
allow creating references for empty archival datasets

keewis:
Uninitialized variables don't have chunks, but can have a fill value set:

``` python
import h5netcdf

with h5netcdf.File("test.nc", mode="w") as f:
    f.dimensions = {"x": 100, "y": 200}
    f.create_variable("var", dimensions=("x", "y"), dtype="int64", fillvalue=100, chunks=(50, 100))
```
`xarray` opens this file just fine, but `kerchunk` won't return chunks for it. `virtualizarr` uses the fill value to construct the variable instead, which doesn't work because the variable is not actually 0D. This usually indicates an issue with the files, so I chose to raise an error (it doesn't have to, though, so I'm not sure this is the best way forward).

- [x] Tests added
- [x] Tests passing
- [ ] Full type hint coverage


mdsumner:
another example fwiw is this seaice product, it started in 1978 and initially only had data every 2 days, but became daily later ... the old binary files just skipped a day but the recent-ish netcdf rewrite filled the blanks with these empty files, they only have the dims, attributes, and the dummy crs variable (which now I think about it will probably be fine for virtual-ref). 

```
https://n5eil01u.ecs.nsidc.org/PM/NSIDC-0051.002/1978.10.27/NSIDC0051_SEAICE_PS_S25km_19781027_v2.0.nc
```

that needs earthdata creds, so zipped and attached: 

[NSIDC0051_SEAICE_PS_S25km_19781027_v2.0.nc.zip](https://github.com/user-attachments/files/17415359/NSIDC0051_SEAICE_PS_S25km_19781027_v2.0.nc.zip)


But, I wanted to put it out there because these empty files do occur for various reasons.

keewis:
> I wanted to put it out there because these empty files do occur for various reasons.

thanks for confirming that these are not just broken files.

In that case I wonder how to best support these: in theory, writing `.zarray` / `.zattrs` should be enough, since `zarr` will also use the `fill_value` to fill missing chunks. In other words, how do we generate a `ManifestArray` that does not contain chunks (i.e., a size-0 `ManifestArray`, but with the shape / chunksizes from `zarray`)?

TomNicholas:
> Size-0 ManifestArray

ManifesArrays are just 3 numpy arrays in a trenchcoat, and you can have length-0 numpy arrays right? So this should be possible, but might need some special casing.

mdsumner:
also fwiw as a todo for me, the GDAL autotest suite has some metadata-only examples, and I wanted to explore how xarray treats related, as opposed to zarr python itself, in case there was some misalignment in how GDAL should behave too: 

https://github.com/OSGeo/gdal/tree/master/autotest/gdrivers/data/zarr/array_attrs.zarr

(xarray is fine with the empty netcdf, but not with the empty GDAL zarr)

keewis:
> https://github.com/OSGeo/gdal/tree/master/autotest/gdrivers/data/zarr/array_attrs.zarr

There's multiple issues with that (I think): `.zarray` is not valid json (it uses single quotes instead of double quotes for the dtype), and `!b1` is unknown to `numpy.dtype` (remove the exclamation point, maybe?).

Once those are fixed, `xarray` uses `zarr.open_group` to open files, which complains about the directory being an array, not a group. To fix that, you'd have to move `.zarray` and `.zattrs` into a subdirectory (`var`) and create a `.zgroup` file one level up. And finally, `xarray` needs the `_ARRAY_DIMENSIONS` attribute to be set and contain dimension names.

keewis:
I've repurposed this PR to instead allow reading and writing variables without chunks (detected by no chunks and `ndim > 0`).

This appears to work properly, but I need help with the typing of `ChunkManifest.empty`: why does mypy think `np.array((), dtype=np.dtypes.StringDType)` is of type `ndarray[Any, dtype[Any]]`? Do I need to use `cast` for that?

TomNicholas:
I think you might need to cast. I don't think numpy's handling of generics like this fully works yet. See also the PR I recently merged that fixed some similar typing errors.

virtualizarr/manifests/manifest.py
```diff
@@ -141,6 +141,15 @@ def __init__(self, entries: dict) -> None:
...
         self._lengths = lengths
 
+    @classmethod
+    def empty(cls) -> "ChunkManifest":
+        """Create an empty chunk manifest."""
+        return cls.from_arrays(
+            paths=np.array((), dtype=np.dtypes.StringDType),
+            offsets=np.array((), dtype="uint64"),
+            lengths=np.array((), dtype="uint64"),
+        )
```
TomNicholas: Why do we need this? Why can't we just use `ChunkManifest.__init__` with `entries={}`?

keewis: that fails because of the list unpacking in `validate_chunk_keys`. It would probably be possible to fix/work around that, but the separate class method makes the intent very clear.

TomNicholas: I think I would prefer not to have a new constructor method just for a special case of the general `__init__` method. I would rather have an if...else... inside `validate_chunk_keys`. I feel like the intent of `ChunkManifest(entries={}` is also pretty clear.

keewis: done

virtualizarr/readers/kerchunk.py
```diff
@@ -243,6 +243,9 @@ def variable_from_kerchunk_refs(
     if chunk_dict:
         manifest = ChunkManifest._from_kerchunk_chunk_dict(chunk_dict)
         varr = virtual_array_class(zarray=zarray, chunkmanifest=manifest)
+    elif len(zarray.shape) != 0:
```
TomNicholas, line 246: An explanatory comment here would be helpful

TomNicholas (commented):
So how do these get represented in the manifest? Size-0 numpy arrays? If so have you tried concatenating them or doing any other operations to make sure that `ManifestArray` doesn't break?

keewis:
> If so have you tried concatenating them or doing any other operations to make sure that `ManifestArray` doesn't break?

Good catch, I forgot to check that. It breaks, obviously, `numpy.concatenate` doesn't know how to concatenate 0d arrays. Which means I'll have to figure out how to work around that.

However, I fear this will touch almost all of the `__array_function__` functions of `ManifestArray`, so it will take me quite a while to figure out.

keewis:
after some more investigation, I believe we won't be able to use `entries={}` (nor size-0 arrays): we need to somehow pass the chunk grid (which exists, there's just no data we can refer to). One option would be to follow this comment: https://github.com/zarr-developers/VirtualiZarr/blob/e6407e01e4dc2536a64c3a418a59ead30bb353bc/virtualizarr/manifests/manifest.py#L112-L114 and explicitly pass the chunk grid shape. This would allow us to translate `entries={}` to the suggestion below.

Another option would be to construct paths / offsets / lengths with the actual shape, but the values for paths is the missing value marker (the `na_object` parameter to `np.dtypes.StringDType`) – then the concatenation would immediately work, but `manifest.dict()` would have to skip over any entry where the path is missing.

keewis:
I've done both (looks like `path == ""` already meant missing chunk), tell me what you think

virtualizarr/readers/kerchunk.py
```diff
@@ -230,6 +231,13 @@ def dataset_from_kerchunk_refs(
     return vds
 
 
+def determine_chunk_grid_shape(zarray):
+    return tuple(
+        math.ceil(length / chunksize)
+        for length, chunksize in zip(zarray.shape, zarray.chunks)
+    )
```
keewis, line 238: is that correct? This would allow partial chunks, which we usually want to avoid

TomNicholas: I already have a utility function for this

https://github.com/zarr-developers/VirtualiZarr/blob/e6407e01e4dc2536a64c3a418a59ead30bb353bc/virtualizarr/manifests/array_api.py#L295

TomNicholas: But yeah this does allow the final chunk to be smaller if that's what you mean. There should be checks for that in other places, but tbh there may be bugs lurking there.

virtualizarr/tests/test_manifests/test_array.py
```diff
@@ -203,6 +203,38 @@ def test_broadcast_any_shape(self, shape, chunks, target_shape):
...
+        marr = ManifestArray(zarray, manifest)
+
+        expanded = np.broadcast_to(marr, shape=target_shape)
+        assert expanded.shape == target_shape
+        assert len(expanded.chunks) == expanded.ndim
+        assert all(
+            len_chunk <= len_arr
+            for len_arr, len_chunk in zip(expanded.shape, expanded.chunks)
+        )
+        assert expanded.manifest.dict() == {}
```
TomNicholas, line 236: Really nice test!

TomNicholas:
So basically the concatenation works okay because under the hood the manifest still contains numpy arrays of the correct shape, they just have `path=''` for all elements?

keewis:
yes, that's it

TomNicholas:
Great. That's presumably less efficient that not storing them explicitly, but it should be robust.

keewis:
probably faster, too, because we don't need to special-case empty chunk manifests.

TomNicholas (approved):
This is great. Ready to merge?

TomNicholas:
I guess maybe a note in the `docs/releases.rst` changelog.

TomNicholas:
Thanks @keewis!

PR #261, opened 2024-10-17
Split kerchunk reader up

### LINKED ISSUE #245: Make readers pluggable via entrypoint system
We could make an entrypoint system for virtualizarr, in addition to plugging in to xarray's backend entrypoint system. That would allow this syntax for creating virtual `xr.Dataset` objects containing `ManifestArrays`, e.g.

```python
vds = vz.open_virtual_dataset(grib_paths, reader='hypergrib')
```

(as suggested in #238)

This would allow users to freely create their own virtual readers / metadata parsers (whatever we call them #239) to deal with their own gnarly file formats, without requiring upstream contributions. This model for extendability has worked very well for xarray, and it's completely analogous to the problem we have here.

If we combined this with #35, we would then have a virtualizarr xarray backend entrypoint, which calls out to specific virtualizarr readers via another entrypoint. That would allow this syntax. 

```python
vds = xr.open_dataset(grib_paths, engine='virtualizarr', reader='hypergrib')
```

I think we should make the entrypoint system in virtualizarr first before we decide if entrypoints calling entrypoints is actually a good idea... 🐶 

cc @ayushnag

### LINKED ISSUE #257: Ensure every reader uses `dataset_from_kerchunk_refs` before returning a virtual dataset
From @TomNicholas  [comment here](https://github.com/zarr-developers/VirtualiZarr/pull/251#discussion_r1796402644):  `dataset_from_kerchunk_refs` should be a common utility that every reader that produces kerchunk-formatted references calls internally before returning a virtual dataset.

TomNicholas:
Instead of trying to treat all uses of kerchunk references as one "reader", this PR instead splits them to consider each separate filetype to be one "reader", even if many of those readers call kerchunk code.

This means every "reader" is now a separate definition of a function `open_virtual_dataset`, which the top-level `open_virtual_dataset` picks between with one big `match case` statement. Generalizing that `match` to be pluggable by third-party libraries would close #245.

This also makes the structure of our dependence on kerchunk much clearer - @mpiannucci it should now be a lot easier for you to try out the kerchunk-to-icechunk use case mentioned in https://github.com/zarr-developers/VirtualiZarr/issues/258#issuecomment-2419582281.

- [x] Closes #257, addresses https://github.com/zarr-developers/VirtualiZarr/issues/258#issuecomment-2420344534, and is a big step towards #245.
- [ ] ~~Tests added~~
- [x] Tests passing
- [x] Full type hint coverage

cc @norlandrhagen @sharkinsspatial @ghidalgo3


TomNicholas:
Note to self: I should just use an ABC with an `open_virtual_dataset` method defined at this point. That would standardize the interface, which would help shut mypy up, and it brings us closer towards what we would need for an entrypoint system.

virtualizarr/readers/kerchunk.py
```diff
@@ -1,310 +1,55 @@
...
 
-from virtualizarr.backend import FileType, separate_coords
-from virtualizarr.manifests import ChunkManifest, ManifestArray
+from virtualizarr.readers.common import VirtualBackend
+from virtualizarr.translators.kerchunk import dataset_from_kerchunk_refs
 from virtualizarr.types.kerchunk import (
-    KerchunkArrRefs,
     KerchunkStoreRefs,
 )
 from virtualizarr.utils import _FsspecFSFromFilepath
```
TomNicholas, line 12: Notice the kerchunk reader never imports kerchunk

virtualizarr/readers/kerchunk.py
```diff
@@ -1,310 +1,69 @@
...
+    @staticmethod
+    def open_virtual_dataset(
+        filepath: str,
+        group: str | None = None,
+        drop_variables: Iterable[str] | None = None,
+        loadable_variables: Iterable[str] | None = None,
+        decode_times: bool | None = None,
+        indexes: Mapping[str, Index] | None = None,
+        reader_options: Optional[dict] = None,
+    ) -> Dataset:
```
TomNicholas, line 25: Another nice thing about this approach is that it makes it really clear which readers are missing which features. For example `drop_variables` did not exist for the kerchunk reader (I've now added a simple implementation of it).

TomNicholas:
@keewis you might have opinions on this PR - I'm trying to move towards a system of virtual "backend readers" like xarray has backends. My proposal (see #245) is to eventually have one actual xarray backend that calls virtualizarr, and virtualizarr calls one of a number of registered virtual backends depending on the filetype.

TomNicholas:
Note: ensure warning is raised on `decode_cf`

virtualizarr/backend.py
```diff
@@ -112,217 +142,109 @@ def open_virtual_dataset(
...
 
-            with fs.open_file() as of:
-                refs = ujson.load(of)
+            return KerchunkVirtualBackend.open_virtual_dataset(
+                filepath,
+                drop_variables=drop_variables,
+                loadable_variables=loadable_variables,
+                indexes=indexes,
+                reader_options=reader_options,
+            )
```
TomNicholas, line 175: This needs to have the same signatures as the others

TomNicholas: done in https://github.com/zarr-developers/VirtualiZarr/pull/261/commits/bf78b8489a91a3fdf865f6ebadf464f205444d98

keewis:
> you might have opinions on this PR

I'm still somewhat split on the idea to create a `virtualizarr` backend for `xr.open_dataset`: I think of `xr.open_dataset` as a way to open a dataset (i.e. metadata and data), while with `virtualizarr` we're not actually opening the dataset, we're just reading the metadata and the chunk locations. I'm aware that this distinction is somewhat blurred if you can load the actual data using the virtual dataset, but still I feel somewhat uncomfortable with that idea (no strong opposition, though).

However, I totally support the creation of a plugin system for `open_virtual_dataset`, as that means that the code for other packages does not have to live in the base package.

For an actual code review I'd have to read the changes you're proposing here, which I will try do at some point during the weekend (but the PR is huge, so might take me some time).

TomNicholas:
> I'm still somewhat split on the idea to create a virtualizarr backend for xr.open_dataset: I think of xr.open_dataset as a way to open a dataset (i.e. metadata and data), while with virtualizarr we're not actually opening the dataset, we're just reading the metadata and the chunk locations. I'm aware that this distinction is somewhat blurred if you can load the actual data using the virtual dataset, but still I feel somewhat uncomfortable with that idea (no strong opposition, though).

I agree actually. I'm still not sure that it's a good idea. Maybe `open_virtual_dataset` is sufficient.

> However, I totally support the creation of a plugin system for `open_virtual_dataset`, as that means that the code for other packages does not have to live in the base package.

:+1:

> For an actual code review I'd have to read the changes you're proposing here, which I will try do at some point during the weekend (but the PR is huge, so might take me some time).

No worries! I'm pretty sure this is fine, it can also always be refactored further.

keewis:
(the failing tests are caused by a bad merge, as the changes to `variable_from_kerchunk_refs` from #260 disappeared)

TomNicholas:
> the failing tests are caused by a bad merge

Thanks for that heads up!

> Note: ensure warning is raised on `decode_cf`

Actually the warning should be on `cf_variables`. `decode_times` should be passed through, which it now is. But I'm confused why that didn't cause failures...

virtualizarr/backend.py
```diff
@@ -1,25 +1,39 @@
...
+    "kerchunk": KerchunkVirtualBackend,
+    "zarr_v3": ZarrV3VirtualBackend,
+    "dmrpp": DMRPPVirtualBackend,
+    # all the below call one of the kerchunk backends internally (https://fsspec.github.io/kerchunk/reference.html#file-format-backends)
+    "netcdf3": NetCDF3VirtualBackend,
+    "hdf5": HDF5VirtualBackend,
+    "netcdf4": HDF5VirtualBackend,  # note this is the same as for hdf5
+    "tiff": TIFFVirtualBackend,
+    "fits": FITSVirtualBackend,
+}
```
TomNicholas, line 36: @maxrjones I think this is what you were suggesting.

TomNicholas:
I'm going to merge this now because I know @sharkinsspatial and @ayushnag would like to work off of it, but @keewis feel free to add any comments / thoughts here and I will address them in follow-up PRs.

PR #264, opened 2024-10-20
Add CI job for testing upstream versions of dependencies

### LINKED ISSUE #263: Test against Zarr v3 in CI
If we add CI workflows that use the bleeding edge versions of upstream dependencies, we can use it to test against zarr-python-v3.

If we add another env that uses upstream versions but only the minimum dependencies (i.e. no kerchunk), then we can use that to test Icechunk integration against.

We may also want to add another env to test V3 with kerchunk.

This way we can start testing every feature against each version of key dependencies, in a way that will still be a somewhat sane setup even after the whole ecosystem works with zarr-python-v3.

TomNicholas:
This environment does not import kerchunk, so this might pass 🤞 

- [x] Closes #263
- [x] Tests passing


ci/upstream.yml
```diff
@@ -0,0 +1,29 @@
...
+  - pytest-mypy
+  - pytest-cov
+  - pytest
+  - pooch
+  - fsspec
+  - pip
+  - pip:
+    - zarr==3.0.0b1
+    - git+https://github.com/pydata/xarray@zarr-v3
+    - git+https://github.com/zarr-developers/numcodecs@zarr3-codecs
```
TomNicholas, line 29:
```suggestion
    - git+https://github.com/zarr-developers/numcodecs@zarr3-codecs
    # - git+https://github.com/fsspec/kerchunk@main  # kerchunk is currently incompatible with zarr-python v3 (https://github.com/fsspec/kerchunk/pull/516)
```

ci/upstream.yml
```diff
@@ -0,0 +1,30 @@
...
```
TomNicholas, line 29:
```suggestion
    - zarr==3.0.0b1  # beta release of zarr-python v3
    - git+https://github.com/pydata/xarray@zarr-v3  # zarr-v3 compatibility branch 
    - git+https://github.com/zarr-developers/numcodecs@zarr3-codecs  # zarr-v3 compatibility branch 
```

Issue #262, opened 2024-10-18
Add Zarr Reader(s)


norlandrhagen:
Opening up a tracking issue for Zarr V2 and Zarr V3/Icechunk compatible readers. It might be useful to open up multiple existing Zarr stores and virtualizarr'ize into a single Zarr. I think @maxrjones raised this possibility a while ago, but I can't find seem to find the issue. 

From what I understand from @TomNicholas the [open_virtual_dataset_from_v3_store](https://github.com/zarr-developers/VirtualiZarr/blob/7053bc02677b31177c76904d6c90a594cff30598/virtualizarr/readers/zarr.py#L16) reader is more for chunk manifest / manifest.json style v3 stores. 

- [ ]  Zarr V2 compatible reader
- [ ]  Zarr V3 compatible reader

TomNicholas:
> the open_virtual_dataset_from_v3_store reader is more for chunk manifest / manifest.json style v3 stores.

That's correct - I think it's arguably made obsolete by the open-sourcing of Icechunk... (cc @ldeakin)

> Zarr V3/Icechunk compatible readers

Icechunk uses a different storage layout on disk, so the V3 reader won't work on it. And the Zarr store doesn't define an interface for returning virtual refs. So pulling virtual refs out of Icechunk is a separable issue, and would require Icechunk to create API to support it.

TomNicholas:
@norlandrhagen FYI we can use https://github.com/zarr-developers/zarr-python/pull/2426 instead of `fsspec.ls`!!

norlandrhagen:
Oh great find!

TomNicholas:
To support generating virtual references to sharded v3 data (which we should definitely leave for a later PR), see this https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2434419316.

PR #271, opened 2024-10-24
Zarr reader

norlandrhagen:
**WIP** PR to add a Zarr reader. Thanks to @TomNicholas for the _**how to write a reader**_ guide. 
- [x] Closes https://github.com/zarr-developers/VirtualiZarr/issues/262
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage
- [x] optimizations (e.g. using async interface to list lengths of chunks for each variable concurrently)
- [x] Read v3 Zarr

Future PR(s):
- Read v2 Zarr
- sharded v3 data

To Do:

- [x] Open the store using zarr-python v3 (behind a protected import). This should handle both v2 and v3 stores for us.
- [x] Use zarr-python to list the variables in the store, and check that all loadable_variables are present
For each virtual variable:
- [x] Use zarr-python to get the attributes and the dimension names, and coordinate names (which come from the .zmetadata or zarr.json)
- [x] Use zarr-python to also get the dtype and chunk grid info + everything else needed to create the virtualizarr.zarr.ZArray object (eventually we can skip this step and use a zarr-python array metadata class directly instead of virtualizarr.zarr.ZArray, see
https://github.com/zarr-developers/VirtualiZarr/pull/175)
- [x] Use the knowledge of the store location, variable name, and the zarr format to deduce which directory / S3 prefix the chunks must live in.
- [x] List all the chunks in that directory using fsspec.ls(detail=True), as that should also return the nbytes of each chunk. Remember that chunks are allowed to be missing.
- [x] The offset of each chunk is just 0 (ignoring sharding for now), and the length is the file size fsspec returned. The paths are just all the paths fsspec listed.
- [x] Parse the path and length information returned by fsspec into the structure that we can pass to ChunkManifest.__init__
- [x] Create a ManifestArray from our ChunkManifest and ZArray
- [x] Wrap that ManifestArray in an xarray.Variable, using the dims and attrs we read before
Get the loadable_variables by just using xr.open_zarr on the same store (should use drop_variables to avoid handling the virtual variables that we already have).
- [x] Use separate_coords to set the correct variables as coordinate variables (and avoid building indexes whilst doing it)
- [x] Merge all the variables into one xr.Dataset and return it.
- [x] All the above should be wrapped in a virtualizarr.readers.zarr.open_virtual_dataset function, which then should be called as a method from a ZarrVirtualBackend(VirtualBackend) subclass.
- [x] Finally add that ZarrVirtualBackend to the list of readers in virtualizarr.backend.py


norlandrhagen:
https://github.com/zarr-developers/VirtualiZarr/issues/273

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,372 @@
...
+                "dtype": array_metadata_dict.get("dtype", None),
+                "fill_value": array_metadata_dict.get("fill_value", None),
+                "order": array_metadata_dict.get("order", None),
+            }
+
+            # 3b.
+            # Use zarr-python to also get the dtype and chunk grid info + everything else needed to create the virtualizarr.zarr.ZArray object (eventually we can skip this step and use a zarr-python array metadata class directly instead of virtualizarr.zarr.ZArray
+            array_zarray = ZArray(
+                shape=array_metadata_dict.get("shape", None),
+                chunks=array_metadata_dict.get("chunks", None),
```
abarciauskas-bgse, line 112: seems like there should be a way to dry this up with the `array_encoding` dict above.

norlandrhagen:
Bit of an update. With the help from @sharkinsspatial @abarciauskas-bgse and @maxrjones I got a Zarr loaded as a virtual dataset. 

```python
<xarray.Dataset> Size: 3kB
Dimensions:  (time: 10, lat: 9, lon: 18)
Coordinates:
    lat      (lat) float32 36B ManifestArray<shape=(9,), dtype=float32, chunk...
    lon      (lon) float32 72B ManifestArray<shape=(18,), dtype=float32, chun...
  * time     (time) datetime64[ns] 80B 2013-01-01 ... 2013-01-03T06:00:00
Data variables:
    air      (time, lat, lon) int16 3kB ManifestArray<shape=(10, 9, 18), dtyp...
Attributes:
    Conventions:  COARDS
    title:        4x daily NMC reanalysis (1948)
    description:  Data is from NMC initialized reanalysis\n(4x/day).  These a...
    platform:     Model
    references:   http://www.esrl.noaa.gov/psd/data/gridded/data.ncep.reanaly...
 ```
 
 Next up is how to deal with `fill_values`. 
 
When I try to write it to Kerchunk JSON, I’m running into some `fill_value` dtype issues in the Zarray.

`ZArray(shape=(10,), chunks=(10,), dtype='<f4', fill_value=np.float32(nan), order='C', compressor=None, filters=None, zarr_format=2)`

Where `fill_value=np.float32(nan)` . When I try to write these to JSON via `ds.virtualize.to_kerchunk(format="dict")`, I get  `TypeError: np.float32(nan) is not JSON serializable`.

Wondering how `fill_values` like `np.float32(nan)` should be handled.

There seems to be some conversion logic in @sharkinsspatial's [HDF reader](https://github.com/zarr-developers/VirtualiZarr/blob/560829266f9036951a6f5f39a016339402422bc0/virtualizarr/readers/hdf.py#L245) for converting fill_values . It also looks like there is some fill_value handling in [zarr.py](https://github.com/zarr-developers/VirtualiZarr/blob/fffdc2d831526d4bb2926d253a7915262d0ee827/virtualizarr/zarr.py#L16).

**Comment by TomNicholas** [comment]
> I got a Zarr loaded as a virtual dataset.

Amazing!

> When I try to write it to Kerchunk JSON, I’m running into some fill_value dtype issues in the Zarray.
> 
> `ZArray(shape=(10,), chunks=(10,), dtype='<f4', fill_value=np.float32(nan), order='C', compressor=None, filters=None, zarr_format=2)`
> 
> Where `fill_value=np.float32(nan)` . When I try to write these to JSON via `ds.virtualize.to_kerchunk(format="dict")`, I get `TypeError: np.float32(nan) is not JSON serializable`.
>
> Wondering how `fill_values` like `np.float32(nan)` should be handled.

This seems like an issue that should actually be orthogonal to this PR (if it weren't for the ever-present difficulty of testing). Either the problem is in the `ZArray` class and what types it allows, or it's in the Kerchunk writer not knowing how to serialize a valid `ZArray`. Either way if `np.float32(nan)` is a valid `fill_value` for a zarr array then it's not the fault of the new zarr reader.

File: virtualizarr/readers/zarr.py [hunk 2]
```diff
@@ -0,0 +1,329 @@
...
+        # 3. For each virtual variable:
+        for var in virtual_vars:
+            # 3a.  Use zarr-python to get the attributes and the dimension names,
+            # and coordinate names (which come from the .zmetadata or zarr.json)
+            array_metadata = zg[var].metadata
+
+            array_metadata_dict = array_metadata.to_dict()
+
+            # extract _ARRAY_DIMENSIONS and remove it from attrs
+            array_dims = array_metadata_dict.get("attributes").pop("_ARRAY_DIMENSIONS")
```
TomNicholas, line 82: See https://github.com/zarr-developers/VirtualiZarr/issues/273

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,329 @@
...
+        # virtual variables are available variables minus drop variables & loadable variables
+        virtual_vars = list(
+            set(zarr_arrays) - set(loadable_variables) - set(drop_variables)
+        )
+
+        virtual_variable_mapping = {}
+        all_array_dims = []
+
+        # 3. For each virtual variable:
+        for var in virtual_vars:
```
TomNicholas, line 74: Everything inside here is for a single array so should also probably be refactored into a function

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,329 @@
...
+
+            array_chunk_sizes = {
+                val["name"].split("/")[-1]: {
+                    "path": array_path,
+                    "offset": 0,
+                    "length": val["size"],
+                }
+                for val in array_mapper.fs.ls(array_mapper.root, detail=True)
+                if not val["name"].endswith((".zarray", ".zattrs", ".zgroup"))
+            }
```
TomNicholas, line 136: Should also be a function

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,329 @@
...
+                data=array_manifest_array,
+                attrs=array_metadata_dict.get("attributes", {}),
+                encoding=array_encoding,
+            )
+
+            virtual_variable_mapping[f"{var}"] = array_variable
+
+            all_array_dims.extend([dim for dim in array_dims])
+
+        coord_names = list(set(all_array_dims))
```
TomNicholas, line 160: This seems incorrect? I also don't actually see why you need to collect `all_array_dims`

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # 4 Get the loadable_variables by just using xr.open_zarr on the same store (should use drop_variables to avoid handling the virtual variables that we already have).
+        # We want to drop 'drop_variables' but also virtual variables since we already **manifested** them.
```
TomNicholas, line 163: *groan*

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,329 @@
...
+            decode_times=decode_times,
+        )
+
+        # 6 Merge all the variables into one xr.Dataset and return it.
+        return construct_virtual_dataset(
+            virtual_vars=virtual_variable_mapping,
+            loadable_vars=loadable_vars,
+            indexes=indexes,
+            coord_names=coord_names,
+            attrs=zg.attrs.asdict(),
```
TomNicholas, line 184: you remembered the group-level attributes, nice

TomNicholas (commented):
This is a a great start! I think the main thing here is that we don't actually need kerchunk in order to test this reader.

TomNicholas:
> get chunk size with zarr-python (https://github.com/zarr-developers/zarr-python/pull/2426) instead of fsspec

I think we should just do this in this PR. We can point to Tom's PR for now in the CI, but I expect that will get merged before this does anyway. If you look at Tom's implementation it's basically what we're doing here.

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # 2a. Use zarr-python to list the variables in the store
+        zarr_arrays = [val for val in zg.keys()]
+
+        # 2b. and check that all loadable_variables are present
+        assert set(
+            loadable_variables
+        ).issubset(
+            set(zarr_arrays)
+        ), f"loadable_variables ({loadable_variables}) is not a subset of variables in existing Zarr store. This zarr contains:  {zarr_arrays}"
```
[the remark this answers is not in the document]
norlandrhagen: Oh good to know!

TomNicholas:
`Store.getsize` was just merged so we can possibly just use the same `upstream` env we are already using zarr-python https://github.com/zarr-developers/zarr-python/pull/2426

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,356 @@
...
+
+    async def get_chunk_paths() -> dict:
+        # this type hint for dict is doing a lot of work. Should this be a dataclass or typed dict?
+        chunk_paths = {}
+        # Is there a way to call `zarr_group.store.list()` per array?
+        async for item in zarr_group.store.list():
+            if not item.endswith(
+                (".zarray", ".zattrs", ".zgroup", ".zmetadata")
+            ) and item.startswith(array_name):
+                # dict key is created by splitting the value from store.list() by the array_name and trailing /....yuck..
```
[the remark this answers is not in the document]
norlandrhagen: That would be waaay better!

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,356 @@
...
+                chunk_paths[item.split(array_name + "/")[-1]] = {
+                    "path": (
+                        zarr_group.store.root / item
+                    ).as_uri(),  # should this be as_posix() or as_uri()
```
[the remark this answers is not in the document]
norlandrhagen: Great! I'll keep it and remove the comment.

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,356 @@
...
+        # v2 stores
+        # ARRAY_DIMENSIONS should be removed downstream in the icechunk writer.
+        # Should we remove them here as well?
+
+        array_dims = array_metadata_dict.get("attributes").pop("_ARRAY_DIMENSIONS")
+
+    # should these have defaults defined and shared across readers?
+    # Should these have common validation for Zarr V3 codecs & such?
+    # Note! It seems like zarr v2 and v3 don't have the same array_encoding keys..
+    array_encoding = {
```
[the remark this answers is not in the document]
norlandrhagen: Ah okay! That simplifies things.

virtualizarr/readers/zarr.py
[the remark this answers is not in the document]
norlandrhagen: https://github.com/zarr-developers/zarr-python/issues/2507

norlandrhagen:
Semi-blocked by this behavior in zarr-python: https://github.com/zarr-developers/zarr-python/issues/2554

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,444 @@
...
+
+async def get_chunk_paths(
+    zarr_group: zarr.Group, array_name: str, store_path: upath.core.UPath
+) -> dict:
+    chunk_paths = {}
+
+    # Can we call list() on an array instead of the entire store?
+    store_keys = zarr.core.sync.sync(list_store_keys(zarr_group))
+
+    for item in store_keys:
```
jhamman, line 258: The async way to do this is.

```suggestion
    async for item in zarr_group.store.list():
```

> # Can we call list() on an array instead of the entire store?

You probably want `list_dir` then!

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,444 @@
...
+
+    # use UPath for combining store path + chunk key when building chunk manifests
+    store_path = UPath(filepath)
+
+    if reader_options is None:
+        reader_options = {}
+
+    zg = zarr.open_group(
+        filepath, storage_options=reader_options.get("storage_options"), mode="r"
+    )
```
jhamman, line 176: use Zarr's async api function for this.

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,444 @@
...
+
+    # Can we call list() on an array instead of the entire store?
+    store_keys = zarr.core.sync.sync(list_store_keys(zarr_group))
+
+    for item in store_keys:
+        # should we move these filters/checks into list_store_keys?
+        if (
+            not item.endswith((".zarray", ".zattrs", ".zgroup", ".zmetadata", ".json"))
+            and item.startswith(array_name)
+            and zarr.core.sync.sync(chunk_exists(zarr_group=zarr_group, chunk_key=item))
```
jhamman, line 263: since you are in an async function, you can just `await chunk_exists(...)`

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,444 @@
...
+                    f"{zarr_group.metadata.zarr_format} not 2 or 3."
+                )
+
+            # Can we ask Zarr-python for the path and protocol?
+            # This gives path, but no protocol: zarr_group.store.path
+
+            chunk_paths[chunk_key] = {
+                "path": ((store_path / item).as_uri()),
+                "offset": 0,
+                "length": zarr.core.sync.sync(get_chunk_size(zarr_group, item)),
```
jhamman, line 287: no need to sync() here

norlandrhagen:
This V3 zarr store works:

```python

filepath = 's3://carbonplan-share/air_temp.zarr'
vds = open_virtual_dataset(path, indexes={})
vds
```

ToDo:
- [ ] Build manifests from arrays
- [ ] Zarr v2 debug

TomNicholas:
Nice @norlandrhagen !

> Build manifests from arrays

What does this mean?

norlandrhagen:
> What does this mean? EDIT: Is it referring to using [zarr-developers/zarr-python#2426](https://github.com/zarr-developers/zarr-python/pull/2426) instead of using `fs.ls()`?

Ah, I was referring to building the `ManifestArrays` with the `from_arrays` method instead of building up a dictionary first. Also, no more `fs.ls`!. Zarr-python is giving us all the chunk info!

norlandrhagen:
Awesome! I'm planning on getting back to this today, so I bet I'll have question! From what I remember from 2024, there were some questions around the `ZARR_DEFAULT_FILL_VALUE`.

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,302 @@
...
+    VirtualBackend,
+    construct_virtual_dataset,
+    maybe_open_loadable_vars_and_indexes,
+)
+from virtualizarr.utils import check_for_collisions
+from virtualizarr.zarr import ZArray
+
+if TYPE_CHECKING:
+    import zarr
+    from zarr.core.common import concurrent_map
```
jhamman, line 21: A cautionary note: `zarr.core` is meant to be private API.

norlandrhagen: So I'm guessing importing this from the Zarr private api is a big no-no?

TomNicholas: If the implementation of concurrent_map isn't too complicated you could potentially vendor it instead.

norlandrhagen:
Jan 10th ToDo / Notes:

### Issues:
- [ ] Zarr V3 normalized uris fail. It seems like this [PR](https://github.com/zarr-developers/zarr-python/pull/2533):  might fix this [issue](https://github.com/zarr-developers/zarr-python/issues/2554). I've commented out `validate_and_normalize_path_to_uri` in the Zarr reader for now. 
- [ ] The `from zarr.core.common import concurrent_map` is imported from the Zarr V3 private API. 
- [ ] The Zarr reader integration test is xfailed b/c it requires Kerchunk. Can we roundtrip using a temporary icechunk store? @TomNicholas 
- [ ] It's very possible there is some wackiness in the codecs. If you have a minute, would love some feedback on this section @sharkinsspatial 🙏 
- [ ] [Upstream tests](https://github.com/zarr-developers/VirtualiZarr/pull/375) here and on main failing with Zarr V3 updates.

TomNicholas:
> Zarr V3 normalized uris fail. It seems like this https://github.com/zarr-developers/zarr-python/pull/2533: might fix this https://github.com/zarr-developers/zarr-python/issues/2554. I've commented out `validate_and_normalize_path_to_uri` in the Zarr reader for now.

If `validate_and_normalize_path_to_uri` is being used, then I'm confused why you wouldn't run into #373?

norlandrhagen:
I removed code and docs related to https://github.com/zarr-developers/VirtualiZarr/issues/359.

TomNicholas:
Can I help here? Or is merging this blocked until we can pin zarr-python v3? (See #392)

norlandrhagen:
> Can I help here? Or is merging this blocked until we can pin zarr-python v3? (See #392)

I think switching to Zarr-python z3 would certainly simplify a few things! 
- This depends on an un-yet released version of zarr-python. 
- Writing to the newer version of icechunk is blocked as of now.

abarciauskas-bgse:
@norlandrhagen I don't have all the context so apologies but I am curious: is the `get_codec` function you are referring to the `numcodecs.get_config`? Is there an existing test or test case you are developing that uses this function?

TomNicholas:
Good question - where the hell is that function? You were showing it to me earlier but I can't find it anywhere in this PR... Maybe you have unpushed local changes @norlandrhagen ?

norlandrhagen:
Thanks for jumping in @abarciauskas-bgse! 
`get_codec` seems to be called within Zarr's [ArrayV2Metadata](https://github.com/zarr-developers/VirtualiZarr/blob/f4d9f5bb79af50f801ff8ae24fc49a77fc00a77f/virtualizarr/utils.py#L163) in utils.py of VirtualiZarr. 

The `compressor_config` the comes out of `get_codec_config` looks like: `{'name': 'zstd', 'configuration': {'level': 0, 'checksum': False}}`. 
If the structure if modified to: `{'id': 'zstd', 'level': 0, 'checksum': False}`. (`name` -> `id` and the configuration options are unnested. It works fine as an input `ArrayV2Metadata` and down the line to `get_codec`. 

Here is a more detailed traceback:

<details>

```python 
virtualizarr/tests/test_integration.py:108: in roundtrip_as_kerchunk_parquet
    vds.virtualize.to_kerchunk(f"{tmpdir}/refs.parquet", format="parquet")
virtualizarr/accessor.py:137: in to_kerchunk
    refs = dataset_to_kerchunk_refs(self.ds)
virtualizarr/writers/kerchunk.py:61: in dataset_to_kerchunk_refs
    arr_refs = variable_to_kerchunk_arr_refs(var, str(var_name))
virtualizarr/writers/kerchunk.py:107: in variable_to_kerchunk_arr_refs
    array_v2_metadata = convert_v3_to_v2_metadata(marr.metadata)
virtualizarr/utils.py:177: in convert_v3_to_v2_metadata
    v2_metadata = ArrayV2Metadata(
.venv/lib/python3.11/site-packages/zarr/core/metadata/v2.py:78: in __init__
    compressor_parsed = parse_compressor(compressor)
.venv/lib/python3.11/site-packages/zarr/core/metadata/v2.py:288: in parse_compressor
    return numcodecs.get_codec(data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

config = {'configuration': {'checksum': False, 'level': 0}, 'name': 'zstd'}

    def get_codec(config):
        """Obtain a codec for the given configuration.
    
        Parameters
        ----------
        config : dict-like
            Configuration object.
    
        Returns
        -------
        codec : Codec
    
        Examples
        --------
    
        >>> import numcodecs as codecs
        >>> codec = codecs.get_codec(dict(id='zlib', level=1))
        >>> codec
        Zlib(level=1)
    
        """
        config = dict(config)
        codec_id = config.pop('id', None)
        cls = codec_registry.get(codec_id)
        if cls is None and codec_id in entries:
            logger.debug("Auto loading codec '%s' from entrypoint", codec_id)
            cls = entries[codec_id].load()
            register_codec(cls, codec_id=codec_id)
        if cls:
            return cls.from_config(config)
>       raise UnknownCodecError(f"{codec_id!r}")
E       numcodecs.errors.UnknownCodecError: codec not available: 'None'

.venv/lib/python3.11/site-packages/numcodecs/registry.py:54: UnknownCodecError
```
</details>

norlandrhagen:
@TomNicholas I think we should be all up to date! 

Running this: `uv run pytest 'virtualizarr/tests/test_integration.py::TestRoundtrip::test_zarr_roundtrip' -s -v` (with pixi if you prefer) triggers a pytest failure. I tossed a debugger into [L162](https://github.com/zarr-developers/VirtualiZarr/blob/f4d9f5bb79af50f801ff8ae24fc49a77fc00a77f/virtualizarr/utils.py#L162) of utils.py to be able to examine the `compressor_config`

norlandrhagen:
I think I can open up a small PR to reproduce this with `tests/test_writers/test_kerchunk::testconvert_v3_to_v2_metadata`

TomNicholas:
If you can reproduce this independently of the Zarr reader then I would support opening a separate issue for that and xfailing this test in this case

TomNicholas:
(Also I think the question @abarciauskas-bgse and I had was where is `get_codec` defined exactly?)

norlandrhagen:
> (Also I think the question @abarciauskas-bgse and I had was where is `get_codec` defined exactly?)

I think it's coming from numcodecs [here](https://github.com/zarr-developers/numcodecs/blob/3c933cf19d4d84f2efc5f3a36926d8c569514a90/numcodecs/registry.py#L24). which is called from zarr.core.metadata.v2 [here](https://github.com/zarr-developers/zarr-python/blob/9e8b50ae19cc63ad573f58569c3ef5826a5c60fc/src/zarr/core/metadata/v2.py#L288) which is used by [ArrayV2Metadata](https://github.com/zarr-developers/zarr-python/blob/9e8b50ae19cc63ad573f58569c3ef5826a5c60fc/src/zarr/core/metadata/v2.py#L46C7-L46C22)

TomNicholas:
@norlandrhagen what's the move here to get this merged? Xfail that test and raise a tracking issue?

docs/faq.md
```diff
@@ -77,7 +77,7 @@ vds.virtualize.to_icechunk(icechunkstore)
 
 ### I already have some data in Zarr, do I have to resave it?
 
-No! VirtualiZarr can (well, [soon will be able to](https://github.com/zarr-developers/VirtualiZarr/issues/262)) create virtual references pointing to existing Zarr stores in the same way as for other file formats.
+No! VirtualiZarr can create virtual references pointing to existing Zarr stores in the same way as for other file formats. Note: Currently only reading Zarr V3 is supported.
```
TomNicholas, line 80: Do we have an issue to track learning to read zarr v2?

docs/faq.md
```diff
@@ -119,7 +119,7 @@ Users of Kerchunk may find the following comparison table useful, which shows wh
...
 | From a Zarr v2 store                                                     | `kerchunk.zarr.ZarrToZarr`                                                                                                          | `open_virtual_dataset(..., filetype='zarr')` (❌ Not yet implemented - see [issue #262](https://github.com/zarr-developers/VirtualiZarr/issues/262))                                                                                       |
-| From a Zarr v3 store                                                     | ❌                                                                                                          | `open_virtual_dataset(..., filetype='zarr')` (❌ Not yet implemented - see [issue #262](https://github.com/zarr-developers/VirtualiZarr/issues/262))                                                                                       |
+| From a Zarr v3 store                                                     | ❌                                                                                                          | `open_virtual_dataset(..., filetype='zarr')`                                                                                        |
```
TomNicholas, line 122: remove the big red cross?
```suggestion
| From a Zarr v3 store                                                     |                                                                                                          | `open_virtual_dataset(..., filetype='zarr')`                                                                                        |
```

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,195 @@
...
+    filepath: str,
+    *,
+    reader_options: Optional[dict] = None,
+    drop_variables: str | Iterable[str] | None = None,
+    group: str | None = None,
+):
+    reader_options = reader_options or {}
+    import ipdb
+
+    ipdb.set_trace()
```
TomNicholas, line 118:
```suggestion

```

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,195 @@
...
+    ManifestArray,
+    ManifestGroup,
+    ManifestStore,
+)
+from virtualizarr.manifests.manifest import validate_and_normalize_path_to_uri  # noqa
+from virtualizarr.readers.api import VirtualBackend
+from virtualizarr.vendor.zarr.core.common import _concurrent_map
+
+if TYPE_CHECKING:
+    pass
```
TomNicholas, line 30:
```suggestion
```

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,195 @@
...
+    np.dtype("bool").kind: False,
+    np.dtype("int").kind: 0,
+    np.dtype("float").kind: 0.0,
+    np.dtype("complex").kind: [0.0, 0.0],
+    np.dtype("datetime64").kind: 0,
+}
+
+
+if TYPE_CHECKING:
+    import zarr
```
TomNicholas, line 46:
```suggestion
import zarr
```

virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+async def get_chunk_mapping_prefix(zarr_array: zarr.AsyncArray, filepath: str) -> dict:
+    """Create a dictionary to pass into ChunkManifest __init__"""
+
+    # TODO: For when we want to support reading V2 we should parse the /c/ and "/" between chunks
+
+    prefix = zarr_array.name.lstrip("/") + "/c/"
```
TomNicholas, line 54: I think this will fail for scalars. I recently learned that zarr v3 uses just the path `c` to hold scalars (because `c/0` would imply that there is at least one dimension to have a chunk index along).

TomNicholas: ISTG it's in the zarr spec but I can't find it now

virtualizarr/xarray.py
```diff
@@ -77,10 +77,7 @@ def construct_virtual_dataset(
             return replace_virtual_with_loadable_vars(
                 fully_virtual_ds, loadable_ds, loadable_variables
             )
-
     else:
-        # TODO pre-ManifestStore codepath, remove once all readers use ManifestStore approach
-
```
TomNicholas, line 83: We still need to do this TODO 😆

TomNicholas (approved):
Thanks for your patience @norlandrhagen , and for your help @maxrjones .

I believe I spotted one small bug, but otherwise this looks great.

The implementation is very neat now!

virtualizarr/readers/zarr.py
norlandrhagen: 🤦

docs/faq.md
norlandrhagen: https://github.com/zarr-developers/VirtualiZarr/issues/565

virtualizarr/readers/zarr.py
maxrjones: It got moved from the core spec to the chunk_key_encoding extension - https://zarr-specs.readthedocs.io/en/latest/v3/chunk-key-encodings/default/index.html

norlandrhagen: I started a test to try to check this. It's xfailed for now. Can we infer a scalar `c` chunk from store.info or?

conftest.py
```diff
@@ -58,6 +58,17 @@ def zarr_store(tmpdir, request):
     return filepath
 
 
+@pytest.fixture()
+def zarr_store_scalar(tmpdir):
+    import zarr
+
+    # can/should we create a memorystore instead?
```
TomNicholas, line 65: yes, the bug should still show itself for a memorystore, because the bug is at the level of zarr keys, before the store implementation maps them to filepaths/memory locations.

virtualizarr/readers/zarr.py
norlandrhagen: Added a test for this: https://github.com/zarr-developers/VirtualiZarr/blob/ccc92796bc1f207c47a92660e8dafa3c8779ed9b/virtualizarr/tests/test_readers/test_zarr.py#L79

norlandrhagen: 1. That's what I ended up using!
2. Yup!

TomNicholas: it's a little verbose but totally clear. LGTM!

TomNicholas (approved):

PR #272, opened 2024-10-25
Append to icechunk stores

abarciauskas-bgse:
This resizes the arrays which are being appended to and, probably too naïvely, increments the append_dim index of the chunk key by an offset of the existing number of chunks along the append dimension.

Also Zarr append ref: https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/array.py#L1134-L1186

- [x] Closes https://github.com/zarr-developers/VirtualiZarr/issues/311
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


virtualizarr/writers/icechunk.py
```diff
@@ -51,7 +53,10 @@ def dataset_to_icechunk(ds: Dataset, store: "IcechunkStore") -> None:
 
     # TODO only supports writing to the root group currently
     # TODO pass zarr_format kwarg?
-    root_group = Group.from_store(store=store)
+    if store.mode.str == "a":
+        root_group = Group.open(store=store, zarr_format=3)
+    else:
```
norlandrhagen, line 58: Is `zarr_format=3` needed for `from_store` and `open` or just open?

virtualizarr/writers/icechunk.py
```diff
@@ -124,15 +134,37 @@ def write_virtual_variable_to_icechunk(
     group: "Group",
     name: str,
     var: Variable,
+    append_dim: Optional[str] = None,
```
norlandrhagen, line 137: Probably a down the road concern, but maybe we should add a validation / check that the append dim exists within the store.

virtualizarr/writers/icechunk.py
TomNicholas: If it wasn't needed before I'm not sure why it would be needed now

virtualizarr/writers/icechunk.py
```diff
@@ -96,6 +103,7 @@ def write_variables_to_icechunk_group(
             group=group,
             name=name,
             var=var,
+            append_dim=append_dim,
```
TomNicholas, line 106: This also should be added to the `ds.to_zarr` call above too, so that non-virtual variables can be appended.

TomNicholas: But I don't think we can rely on that doing validation checks for us.

virtualizarr/writers/icechunk.py
```diff
@@ -124,15 +134,37 @@ def write_virtual_variable_to_icechunk(
...
+    append_dim: Optional[str] = None,
 ) -> None:
     """Write a single virtual variable into an icechunk store"""
     ma = cast(ManifestArray, var.data)
     zarray = ma.zarray
+    shape = zarray.shape
+    mode = store.mode.str
+
+    # Aimee: resize the array if it already exists
+    # TODO: assert chunking and encoding is the same
```
TomNicholas, line 146: Should also test that it raises a clear error if you try to append with chunks of a different dtype etc. I would hope zarr-python would throw that for us.

virtualizarr/writers/icechunk.py
```diff
@@ -124,15 +134,37 @@ def write_virtual_variable_to_icechunk(
     group: "Group",
     name: str,
     var: Variable,
+    append_dim: Optional[str] = None,
 ) -> None:
     """Write a single virtual variable into an icechunk store"""
     ma = cast(ManifestArray, var.data)
     zarray = ma.zarray
+    shape = zarray.shape
```
TomNicholas, line 142: better to have a clear `if...else` then to set this and then maybe later overwrite it

virtualizarr/writers/icechunk.py
```diff
@@ -124,15 +134,37 @@ def write_virtual_variable_to_icechunk(
...
+    if name in existing_keys and mode == "a":
+        # resize
+        dims = var.dims
+        if append_dim in dims:
+            append_axis = dims.index(append_dim)
+            existing_array = group[name]
+            existing_size = existing_array.shape[append_axis]
+            existing_num_chunks = int(
+                existing_size / existing_array.chunks[append_axis]
+            )
```
TomNicholas, line 158: There's a whole beartrap here around noticing if the last chunk is smaller than the other chunks. We should throw in that case (because zarr can't support it without variable-length chunks).

virtualizarr/writers/icechunk.py
```diff
@@ -124,15 +134,37 @@ def write_virtual_variable_to_icechunk(
...
+            new_shape = list(existing_array.shape)
+            new_shape[append_axis] += var.shape[append_axis]
+            shape = tuple(new_shape)
```
TomNicholas, line 161: I wonder if some axis-handling logic from the `concat` function I wrote for `ManifestArray` could be re-used here.

virtualizarr/writers/icechunk.py
```diff
@@ -164,6 +199,8 @@ def write_manifest_virtual_refs(
     group: "Group",
     arr_name: str,
     manifest: ChunkManifest,
+    append_axis: Optional[int] = None,
+    existing_num_chunks: Optional[int] = None,
```
TomNicholas, line 203: That's very neat!

TomNicholas (commented):
> All this does at the moment is resize the arrays which are being appended to and, probably too naïvely, increments the `append_dim` index of the chunk key by an offset of the existing number of chunks along the append dimension.

I think that's great! Does xarray have any similar logic in it?

> Also this is not fully working yet, it is getting a decompression error 😭

This feature should be orthogonal to all of that, so to begin with I would concentrate on writing tests with very simple arrays, even uncompressed ones.

abarciauskas-bgse:
> I think that's great! Does xarray have any similar logic in it?

In the case of appending to a zarr store using xarray, 
- From what I can tell, resizing happens [here](https://github.com/pydata/xarray/blob/main/xarray/backends/zarr.py#L1002-L1012). (Btw if someone can explain write_region to me I would appreciate it, I couldn't find good documentation anywhere). 
- For writing actual chunks of data to keys, I believe that currently happens [here (in zarr.array._set_selection](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/array.py#L1077-L1089). I will continue to dig into how it's working in xarray and zarr to understand how it should work here.

> 
> > Also this is not fully working yet, it is getting a decompression error 😭
> 
> This feature should be orthogonal to all of that, so to begin with I would concentrate on writing tests with very simple arrays, even uncompressed ones.

Yes I think my next step will be to write some simple tests.

virtualizarr/writers/icechunk.py
abarciauskas-bgse: Looking at https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/manifests/array_api.py#L88 I think some of that could definitely be reused here.

virtualizarr/writers/icechunk.py
```diff
@@ -112,58 +124,145 @@ def write_variable_to_icechunk(
...
+            dimension_names=var.dims,
+            fill_value=zarray.fill_value,
+            # TODO fill_value?
+        )
 
-    _encoding_keys = {"_FillValue", "missing_value", "scale_factor", "add_offset"}
-    for k, v in var.encoding.items():
-        if k in _encoding_keys:
+        # TODO it would be nice if we could assign directly to the .attrs property
+        # Aimee: Can we assume that the attributes are the same for the new array?
```
TomNicholas, line 226: I'm not exactly sure what xarray's `.to_zarr` does, but normally concatenation in xarray just silently overwrites attributes and uses the attrs from the first argument. It is configurable in `xr.concat` (see the `combine_attrs` option) but doesn't look like it's configurable in `.to_zarr`, so I think just keeping the attrs the array already has is fine for this PR.

virtualizarr/writers/icechunk.py
```diff
@@ -112,58 +124,145 @@ def write_variable_to_icechunk(
...
+        for k, v in var.attrs.items():
             arr.attrs[k] = encode_zarr_attr_value(v)
+        arr.attrs["_ARRAY_DIMENSIONS"] = encode_zarr_attr_value(var.dims)
```
TomNicholas, line 229: This should be changed to match https://github.com/zarr-developers/VirtualiZarr/pull/286

virtualizarr/writers/icechunk.py
```diff
@@ -112,58 +124,145 @@ def write_variable_to_icechunk(
...
+
+
+def get_axis(
+    dims: list[str],
+    dim_name: str,
+) -> int:
+    return dims.index(dim_name)
+
+
+import zarr
```
TomNicholas, line 161: should be at the top

virtualizarr/writers/icechunk.py
```diff
@@ -112,58 +124,145 @@ def write_variable_to_icechunk(
...
-    for k, v in var.attrs.items():
-        arr.attrs[k] = encode_zarr_attr_value(v)
-    arr.attrs["_ARRAY_DIMENSIONS"] = encode_zarr_attr_value(var.dims)
+        # resize the array
+        arr = resize_array(
+            group=group,
+            name=name,
+            var=var,
+            append_axis=append_axis,
+        )
```
TomNicholas, line 211: Here you determine `existing_num_chunks`, but then don't actually use it until inside `write_manifest_virtual_refs`. I think you could move the `num_chunks` call inside `write_manifest_virtual_refs`, and eliminate the need to pass the `existing_num_chunks` arg down.

virtualizarr/writers/icechunk.py
```diff
@@ -112,58 +124,145 @@ def write_variable_to_icechunk(
...
         group=group,
         arr_name=name,
         manifest=ma.manifest,
+        append_axis=append_axis,
+        existing_num_chunks=existing_num_chunks,
     )
 
 
+def generate_chunk_key(
+    index: np.nditer.multi_index,
```
TomNicholas, line 247: I don't think we need to pass in the complicated numpy type here, because you're about to call `list` on it anyway
```suggestion
    index: tuple[int, ...],
```

virtualizarr/writers/icechunk.py
```diff
@@ -112,58 +124,145 @@ def write_variable_to_icechunk(
...
+    append_axis: Optional[int] = None,
+    existing_num_chunks: Optional[int] = None,
+) -> str:
+    if append_axis is not None:
+        list_index = list(index)
+        # Offset by the number of existing chunks on the append axis
+        list_index[append_axis] += existing_num_chunks
+        index = tuple(list_index)
+    return "/".join(str(i) for i in index)
```
TomNicholas, line 256: I believe that creating mutable lists then altering them in-place is a bit more expensive than just creating the immutable tuple you want in one go. I think instead you could just do something like

```python
return "/".join(
    str(ind + existing_num_chunks) if axis is append_axis else str(ind) for axis, ind in enumerate(index)
)
```

TomNicholas (commented):
Thanks @abarciauskas-bgse ! I have a lot of smaller comments, but generally I think this is looking really promising!

virtualizarr/writers/icechunk.py
```diff
@@ -112,58 +124,145 @@ def write_variable_to_icechunk(
...
+    append_dim: Optional[str] = None,
 ) -> None:
     """Write a single virtual variable into an icechunk store"""
     ma = cast(ManifestArray, var.data)
     zarray = ma.zarray
+    mode = store.mode.str
+
+    dims = var.dims
+    append_axis, existing_num_chunks, arr = None, None, None
+    if mode == "a" and append_dim in dims:
```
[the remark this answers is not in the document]
abarciauskas-bgse: good point, I'll add a test for this

virtualizarr/writers/icechunk.py
abarciauskas-bgse: Ah yes you're right but the challenge is, after we resize the array, than the function existing_num_chunks will not return the right size. I will think about if there is a better way to handle this, so we don't have to pass the existing_num_chunks arg around

virtualizarr/tests/test_writers/test_icechunk_append.py
```diff
@@ -0,0 +1,333 @@
...
+
+    # generate virtual dataset
+    vds = gen_virtual_dataset(file_uri=simple_netcdf4)
+
+    # create the icechunk store and commit the first virtual dataset
+    icechunk_filestore = IcechunkStore.create(storage=icechunk_storage)
+    dataset_to_icechunk(vds, icechunk_filestore)
+    icechunk_filestore.commit(
+        "test commit"
+    )  # need to commit it in order to append to it in the next lines
```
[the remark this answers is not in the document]
abarciauskas-bgse: We need to open the existing store in append mode in order to append otherwise I get the error:

```
zarr.errors.ContainsGroupError: A group exists in store <icechunk.IcechunkStore object at 0x10eaf9100> at path ''.
```

That's the error that I get just trying to use the store object from `IcechunkStore.create(`. But if I do use a store with mode='a' but _do not_ commit to the first store object, I get the following error:

```
FileNotFoundError: <icechunk.IcechunkStore object at 0x10960d490>
```

virtualizarr/writers/icechunk.py
```diff
@@ -112,58 +124,145 @@ def write_variable_to_icechunk(
...
+    manifest_api._check_same_dtypes([ma.dtype, existing_array.dtype])
+    # this is kind of gross - _v3_codec_pipeline returns a tuple
+    # Question: Does anything need to be done to apply the codecs to the new manifest array?
+    manifest_api._check_same_codecs(
+        [list(ma.zarray._v3_codec_pipeline()), existing_array.metadata.codecs]
+    )
+    manifest_api._check_same_chunk_shapes([ma.chunks, existing_array.chunks])
+    manifest_api._check_same_ndims([ma.ndim, existing_array.ndim])
+    arr_shapes = [ma.shape, existing_array.shape]
+    manifest_api._check_same_shapes_except_on_concat_axis(arr_shapes, append_axis)
```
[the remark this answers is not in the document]
abarciauskas-bgse: I did this refactor in https://github.com/zarr-developers/VirtualiZarr/pull/272/commits/28197ed004779af0931600556d0a6a8fff00e015 but let me know what you think, I'm not 100% on the nested function

abarciauskas-bgse:
@TomNicholas thanks so much for your review, I think I have addressed all your comments but I need to address the failures. I'll let you know when I would like a need a re-review.

abarciauskas-bgse:
Finally got all tests passing! But there are still a few things I want to address:
- [ ] getting _check_same_codecs to work for both ManifestArrays and Zarr Arrays of either Zarr version required a new get_codecs function that is quite complex. Not sure if that is the way to go or just use the methods from array_api that work more seamlessly with Zarr v3 arrays.
- [ ] Curious if there is a good way to type the return of `get_codecs` since, I believe, each data type may return something slightly different - such as [virtualizarr.zarr.Codec](https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/zarr.py#L33-L35) vs [zarr.abc.codec](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/abc/codec.py)
- [ ] Want to figure out if I can resolve needing dask to create valid Zarr chunk manifests from xarray tutorial air_temperature dataset

abarciauskas-bgse:
@TomNicholas curious if you have suggestions about an approach to make `get_codecs` work for both Zarr Arrays and Manifest Arrays, of both Zarr versions. See the current implementation in https://github.com/zarr-developers/VirtualiZarr/blob/icechunk-append/virtualizarr/manifests/get_codecs.py. 

I'll otherwise be looking into the range precision difference today.

TomNicholas:
> curious if you have suggestions about an approach to make `get_codecs` work for both Zarr Arrays and Manifest Arrays, of both Zarr versions.

I was going to say that 
a) I had not realised that appending could imply comparing `ManifestArrays` with `zarr.Arrays` directly in order to concatenate them in the Icechunk store,
b) Some kind of high-level function like
```python
def check_compatible(*arrs: ManifestArray | zarr.Array) -> None:
    """Raises if any codecs / anything else is incompatible so these arrays cannot be concatenated"""
```
sounds really useful,
c) The `get_codecs` stuff is likely messy right now but also will hopefully become a lot simpler once we implement https://github.com/zarr-developers/VirtualiZarr/pull/175, but that's also blocked on kerchunk supporting zarr-python v3,
d) Given (c) I suppose you might want to prioritize an approach which is forward-looking, i.e. have `get_codecs` return in the form it returns for `zarr.Array` with zarr-python v3,
e) This might imply refactorings in other places in the codebase where codecs are handled.

TomNicholas:
Also minor nit but I think your `get_codecs` logic should live in `virtualizarr.zarr.py` or even a new `virtualizarr.codecs.py` rather than inside `virtualizarr.manifests`, because it doesn't really have anything to do with the manifests themselves, only with the `ZArray` class.

abarciauskas-bgse:
@TomNicholas thanks for this - very helpful.

2 questions:

1. Regarding `check_compatible(*arrs: ManifestArray | zarr.Array)` this seems very similar to https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/manifests/array_api.py#L28-L40 which I have [modified](https://github.com/zarr-developers/VirtualiZarr/blob/icechunk-append/virtualizarr/manifests/array_api.py#L31-L45) to work for both ManifestArrays and zarr.Arrays (which really just required the get_codecs function). Did you mean you think we should create another function and, if so, how would it be different from what `check_combineable_zarr_arrays` does?

2. I can look into refactoring get_codecs to return in the form it returns for zarr.Array with zarr-python v3 but if this requires changes in a lot of other places, do you think we should address it in a different PR?

TomNicholas:
> Did you mean you think we should create another function and, if so, how would it be different from what `check_combineable_zarr_arrays` does?

I'm just saying let's make sure we have just one official function for doing this within the codebase, that gets called from inside `concatenate` as well `.to_icechunk`, and can handle all the types. Your modified version looks great, so let's just make sure we use that one everywhere.

> I can look into refactoring get_codecs to return in the form it returns for zarr.Array with zarr-python v3 but if this requires changes in a lot of other places, do you think we should address it in a different PR?

If you can get this PR to work without doing all those changes then we can punt on them, but if you need those changes I agree it would be neater to separate them into a different PR.

abarciauskas-bgse:
@TomNicholas To state what you already know, I believe we can assume if zarr-python is installed, then it would be zarr-python version 3, since zarr is a dependency only if icechunk is a dependency (as far as I can tell). 

But, if we assume we are still supporting both zarr format version 2 and zarr format version 3, then the `get_codecs` function can't always return the same type, since for zarr format version 2, there is no codecs property. So the solution I have settled on is returning either a [`virtualizarr.zarr.Codec`](https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/zarr.py#L32-L35) object, for version 2 zarr arrays, and tuple of the various codecs supported for version 3 zarr arrays.

Note: I have to do a type conversion of `array.metadata.codecs` to a tuple in codecs.py. The type of the codecs attribute in zarr-python's ArrayV3Metadata is a tuple but in actuality, a list is what is stored. I believe it's because of [these lines](https://github.com/zarr-developers/zarr-python/blob/v3.0.0-beta.2/src/zarr/core/metadata/v3.py#L257-L258). Should this be fixed in zarr-python?

TomNicholas:
> I believe we can assume if zarr-python is installed, then it would be zarr-python version 3, since zarr is a dependency only if icechunk is a dependency (as far as I can tell).

Unfortunately not, because kerchunk still uses zarr-python v2. But the only time this is going to come up is when we deal with `zarr.Array` directly, which only happens within the icechunk append functionality, and if you're using icechunk then v3 must be the one installed.

So I think you could maybe just have the `check_combinable` function that accepts `zarr.Array` raise `NotImplemented` if it encounters a `zarr.Array` with zarr-python v2 installed?

> The type of the codecs attribute in zarr-python's `ArrayV3Metadata` is a tuple but in actuality, a list is what is stored.

Yeah that looks like an upstream bug to me.

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -285,6 +286,422 @@ def test_write_loadable_variable(
...
+        dataset_to_icechunk(vds2, icechunk_filestore_append, append_dim="time")
+        new_ds = open_zarr(icechunk_filestore_append, consolidated=False, zarr_format=3)
+
+        expected_ds1, expected_ds2 = open_dataset(filepath1), open_dataset(filepath2)
+        expected_ds = concat([expected_ds1, expected_ds2], dim="time").drop_vars(
+            ["lon", "lat", "time"], errors="ignore"
+        )
+        # Because we encode attributes, attributes may differ, for example
+        # actual_range for expected_ds.air is array([185.16, 322.1 ], dtype=float32)
+        # but encoded it is [185.16000366210935, 322.1000061035156]
```
abarciauskas-bgse, line 496: @TomNicholas I cannot xrt.assert_identical when the encoded value returned from xarray.backend.zarr.encode_zarr_attr_value is different from the actual value. Let me know if you think assert_equal is acceptable or there is another route to try.

abarciauskas-bgse:
@TomNicholas I believe I have addressed all comments. I'm going to try it out with some real-world datasets and add documentation.

virtualizarr/tests/test_writers/test_icechunk.py
TomNicholas: I think that's okay? Not totally sure though. @mpiannucci do you know?

virtualizarr/writers/icechunk.py
```diff
@@ -96,65 +112,147 @@ def write_variables_to_icechunk_group(
...
+    return dims.index(dim_name)
+
+
+def _check_compatible_arrays(
+    ma: ManifestArray, existing_array: zarr.core.array.Array, append_axis: int
+):
+    manifest_api.check_combinable_zarr_arrays([ma, existing_array])
+    manifest_api.check_same_ndims([ma.ndim, existing_array.ndim])
+    arr_shapes = [ma.shape, existing_array.shape]
+    manifest_api.check_same_shapes_except_on_concat_axis(arr_shapes, append_axis)
```
TomNicholas, line 158: I know I'm being annoying but actually could we move these utility functions to another namespace, e.g. `manifests.utils`? Otherwise we have an "array API" namespace that contains the array API-compatible functions like `concat`/`stack`, but also these random internal utility functions.

virtualizarr/writers/icechunk.py
```diff
@@ -96,65 +112,147 @@ def write_variables_to_icechunk_group(
...
-        shape=zarray.shape,
-        chunk_shape=zarray.chunks,
-        dtype=encode_dtype(zarray.dtype),
-        codecs=zarray._v3_codec_pipeline(),
-        dimension_names=var.dims,
-        fill_value=zarray.fill_value,
-        # TODO fill_value?
-    )
+    dims: list[str] = cast(list[str], list(var.dims))
+    append_axis, existing_num_chunks, arr = None, None, None
```
TomNicholas, line 183: This line seems very weird - `append_axis` could be set to `None` within the `else` below, `existing_num_chunks` could maybe just be 0? and I don't see why you need to set `arr` here at all

virtualizarr/writers/icechunk.py
```diff
@@ -96,65 +112,147 @@ def write_variables_to_icechunk_group(
...
-    # TODO it would be nice if we could assign directly to the .attrs property
-    for k, v in var.attrs.items():
-        arr.attrs[k] = encode_zarr_attr_value(v)
+        # resize the array
+        arr = resize_array(
+            group=group,
+            name=name,
+            var=var,
+            append_axis=append_axis,
+        )
```
TomNicholas, line 210: This is a little weird in that the function has a return value and a side-effect, which is bad functional programming. Something like this would be a little clearer (and avoids passing in `var` only to use `var.shape`

```suggestion
        arr = group[name]

        # resize the array
        resize_array(
            arr,
            shape_to_append=var.shape,
            append_axis=append_axis,
        )
```

virtualizarr/writers/icechunk.py
```diff
@@ -96,65 +112,147 @@ def write_variables_to_icechunk_group(
...
+
+def generate_chunk_key(
+    index: tuple[int, ...],
+    append_axis: Optional[int] = None,
+    existing_num_chunks: Optional[int] = None,
+) -> str:
+    if append_axis and append_axis >= len(index):
+        raise ValueError(
+            f"append_axis {append_axis} is greater than the number of indices {len(index)}"
+        )
```
TomNicholas, line 250: There's perhaps another potential bug if someone passes a negative number for the axis...

virtualizarr/writers/icechunk.py
```diff
@@ -1,11 +1,14 @@
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Optional, cast
 
 import numpy as np
+import zarr  # type: ignore[import-untyped]
 from xarray import Dataset
 from xarray.backends.zarr import encode_zarr_attr_value
 from xarray.core.variable import Variable
+from zarr import Array
```
TomNicholas, line 8: Actually sorry this and the other zarr import should just be removed. You might need to move `from zarr import Array` inside `if TYPE_CHECKING`.

TomNicholas (commented):
Nice!!! Just some final annoying nits from me now 😁

virtualizarr/writers/icechunk.py
abarciauskas-bgse: The `dataset_to_icechunk` function accepts an append_dim by name - so I don't think the append_axis could be negative. It could not exist though which is why we have:

```
    if append_dim and append_dim not in dims:
        raise ValueError(
            f"append_dim {append_dim} not found in variable dimensions {dims}"
        )
```

virtualizarr/writers/icechunk.py
```diff
@@ -51,7 +59,14 @@ def dataset_to_icechunk(ds: Dataset, store: "IcechunkStore") -> None:
 
     # TODO only supports writing to the root group currently
     # TODO pass zarr_format kwarg?
-    root_group = Group.from_store(store=store)
+    if store.mode.str == "a":
```
mpiannucci, line 62: Just noting that `mode` has disappeared from the latest zarr beta and icechunk main branch.

It is replaced with simply `read_only` as a property on the store. The mode will still exist on the Zarr hierarchy though I believe.

abarciauskas-bgse: 🤔 I think we should leave this as-is because it doesn't look like the read_only property is working as I would expect it to in the latest released version of icechunk (`0.1.0a4`):

```python
from icechunk import IcechunkStore, StorageConfig
config = StorageConfig.filesystem("local")

store = IcechunkStore.create(storage=config, read_only=False)
print("Create: " + str(store.mode))
# AccessMode(str='w', readonly=False, overwrite=True, create=True, update=False)

store = IcechunkStore.open_existing(storage=config)
print(store.mode) 
# AccessMode(str='r', readonly=True, overwrite=False, create=False, update=False)

store = IcechunkStore.open_existing(storage=config, read_only=False) 
print(store.mode)
# AccessMode(str='r', readonly=True, overwrite=False, create=False, update=False)

store = IcechunkStore.open_existing(storage=config, mode="a")
print(store.mode)
# AccessMode(str='a', readonly=False, overwrite=False, create=True, update=True)

store = IcechunkStore.open_existing(storage=config, mode="w")
print(store.mode)
# AccessMode(str=w', readonly=False, overwrite=True, create=True, update=False)

store = IcechunkStore.open_existing(storage=config, update=True)
print(store.mode)
# AccessMode(str='r', readonly=True, overwrite=False, create=False, update=False)
```

I note from the above that the only way to open the icechunk store in append mode is still using `mode="a"`

Admittedly, I'm having trouble understanding how all the [AccessMode](https://github.com/earth-mover/icechunk/blob/icechunk-v0.1.0-alpha.4/icechunk/src/zarr.rs#L241-L246) properties are being set in icechunk.

abarciauskas-bgse: @TomNicholas what do you think about pinning the icechunk version and changing the implementation when there is a new release?

TomNicholas: > it doesn't look like the read_only property is working as I would expect

Even if that's not wrong it's certainly highly counter-intuitive - is it reported upstream?

> pinning the icechunk version and changing the implementation when there is a new release?

That sounds reasonable @abarciauskas-bgse - though hopefully by the time you add documentation to this PR this is fixed upstream 🤞

TomNicholas: > do you think it is worth reporting

I can't see any issue with the word `read_only` in it, so I just raised it https://github.com/earth-mover/icechunk/issues/404 

In general I feel raising duplicate issues is better than not flagging potentially-undiscovered bugs.

mpiannucci: I think pinning icechunk is the right thing to do. the `read_only` keyword isnt in the latest released version so its not a concern until then. Sorry for the confusion

TomNicholas:
I don't know what the issue is, but this seems like an unrelated error to the functionality that's added in this PR surely?

This PR has tests, which pass, so I would be in favour of merging it now and making the functionality public + adding documentation in a follow-up. Also some of the codecs stuff might be useful to @norlandrhagen in #271.

Also if it just sits here for a while in open status that's also fine.

mpiannucci:
Hmmmm so I have seen that error before, when I had the wrong order of codecs it caused the compression to be messed up.

mpiannucci:
Oh that's a good find! It looks like chunk is being overwritten somehow for sure. 

If I understand the notebook, the original dataset loads fine from icechunk to xarray before the append. So at least that narrows it down. 

I am happy to try and take a look (this confirms we need a way to get references out of icechunk too)

abarciauskas-bgse:
@mpiannucci thanks for the heads up, I upgraded but it did not resolve the error.

Also, I'm sure you realized this but I think with zarr a change has to be made https://github.com/mpiannucci/kerchunk/blob/v3/kerchunk/utils.py#L55 from `mode='w'` to `read_only=False` when upgrading to icechunk 0.1.0a5 (as it depends on zarr v3.0.0-beta.2).

mpiannucci:
Thanks! Ill sync those changes in kerchunk land. 

I will have time tomorrow to test this PR out for real

mpiannucci:
This is an icechunk bug :( See here: https://github.com/earth-mover/icechunk/issues/418

mpiannucci:
icehcunk alpha 6 is out now, and fixes this bug and should pave the way for this PR to get merged :) Im so exicted about this

abarciauskas-bgse:
I've modified a test to assert that it would fail with icechunk 0.1.0a5 and then succeeds with icechunk 0.1.0a6. However there seem to be some [unrelated mypy errors in the hdf module](https://github.com/zarr-developers/VirtualiZarr/actions/runs/12169854923/job/33943665882?pr=272) - should I look at resolving these?

ci/upstream.yml
```diff
@@ -28,6 +28,6 @@ dependencies:
   - fsspec
   - pip
   - pip:
-    - icechunk # Installs zarr v3 as dependency
-    # - git+https://github.com/fsspec/kerchunk@main  # kerchunk is currently incompatible with zarr-python v3 (https://github.com/fsspec/kerchunk/pull/516)
-    - imagecodecs-numcodecs==2024.6.1
+      - icechunk==0.1.0a6 # Installs zarr v3 as dependency
```
mpiannucci, line 31: We are up to a7 now, no api changes from 6 to 7

TomNicholas:
> I've modified a test to assert that it would fail with icechunk 0.1.0a5 and then succeeds with icechunk 0.1.0a6

Excellent - I was going to suggest that.

How do you feel about merging this now @abarciauskas-bgse ? With docs and/or examples in a follow-up?

abarciauskas-bgse:
Well I was feeling great about it but it seems like something else in icechunk is changed, as I'm running the [notebook](https://github.com/zarr-developers/VirtualiZarr/blob/208e83e6249ed5362f8bab1b4b857b42255e8bfc/noaa-cdr-sst.ipynb) and some functions which were previously sync are now async - note `await` is required when creating the store and also am getting an error that set_virtual_ref needs to be awaited 😞 

@mpiannucci do you know the best way to handle the async changes?

mpiannucci:
@abarciauskas-bgse  I updated your notebook here: https://github.com/zarr-developers/VirtualiZarr/pull/327 and it is working perfectly!

abarciauskas-bgse:
Thanks @mpiannucci but it's still not working for me... when I run IcechunkStore.create it returns a coroutine, not a store. I'm confounded! Not trying to be difficult, I swear! See https://github.com/zarr-developers/VirtualiZarr/blob/e38823c20134858029188260bb834669a202b13e/noaa-cdr-sst.ipynb

mpiannucci:
That is very confusing because it is definitely not an async function in a7! 

https://github.com/earth-mover/icechunk/blob/a8c33b81d329c32a8e34e7151a1a71620967067a/icechunk-python/python/icechunk/__init__.py#L133

Is there any chance your environment is crossed up?

abarciauskas-bgse:
@mpiannucci yes my environment (well at least 1 of them) was messed up. Another one is working 😅 so we should be good now. i'm going to merge your version of the notebook and move it to the examples directory after cleaning up some extra cells...

abarciauskas-bgse:
ok @TomNicholas I think this is finally g2g, I added to release docs and checked the autogenerated API documentation for to_icehunk.

As far as the example I added in examples/ ... perhaps in another PR we can add this to the virtualizarr docs or icechunk docs.

ci/upstream.yml
```diff
@@ -28,6 +28,6 @@ dependencies:
   - fsspec
   - pip
   - pip:
-    - icechunk # Installs zarr v3 as dependency
-    # - git+https://github.com/fsspec/kerchunk@main  # kerchunk is currently incompatible with zarr-python v3 (https://github.com/fsspec/kerchunk/pull/516)
-    - imagecodecs-numcodecs==2024.6.1
+      - icechunk==0.1.0a7 # Installs zarr v3 as dependency
```
TomNicholas, line 31: Better to future-proof it for when the next release of icechunk comes out?
```suggestion
      - icechunk>=0.1.0a7 # Installs zarr v3 as dependency
```

pyproject.toml
```diff
@@ -38,6 +38,9 @@ hdf_reader = [
     "imagecodecs-numcodecs==2024.6.1",
     "numcodecs"
 ]
+icechunk = [
+    "icechunk==0.1.0a7",
+]
```
TomNicholas, line 43: same here
```suggestion
icechunk = [
    "icechunk>=0.1.0a7",
]
```

virtualizarr/accessor.py
```diff
@@ -43,19 +38,24 @@ def to_zarr(self, storepath: str) -> None:
...
-    def to_icechunk(self, store: "IcechunkStore") -> None:
+    def to_icechunk(
+        self, store: "IcechunkStore", append_dim: Optional[str] = None
+    ) -> None:
         """
         Write an xarray dataset to an Icechunk store.
 
         Any variables backed by ManifestArray objects will be be written as virtual references, any other variables will be loaded into memory before their binary chunk data is written into the store.
 
+        If `append_dim` is provided, the virtual dataset will be appended on the append_dim axis to the existing IcechunkStore.
```
TomNicholas, line 49: axis -> dimension

```suggestion
        If `append_dim` is provided, the virtual dataset will be appended to the existing IcechunkStore along the `append_dim` dimension.
```

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -285,6 +287,488 @@ def test_write_loadable_variable(
...
+    storage = StorageConfig.filesystem(str(tmpdir))
+
+    # TODO instead yield store then store.close() ??
+    return storage
+
+
+def generate_chunk_manifest(
+    netcdf4_file: str,
+    shape: Tuple[int, ...],
+    chunks: Tuple[int, ...],
```
TomNicholas, line 342:
```suggestion
    shape: tuple[int, ...],
    chunks: tuple[int, ...],
```

TomNicholas (approved):
LGTM!

TomNicholas:
(@abarciauskas-bgse you should have merge rights)

PR #275, opened 2024-10-25
Update index.md

thodson-usgs:
Remove `fsspec` from `xr.open_dataset` usage example.

Did I misunderstand something? I thought we could just
```python
ds = xr.open_dataset('combined.json', ...)
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -59,13 +59,8 @@ virtual_ds.virtualize.to_kerchunk('combined.json', format='json')
 Now you can open your shiny new Zarr store instantly:
 
 ```python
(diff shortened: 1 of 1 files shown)


TomNicholas:
Oh right yes - with `engine='kerchunk'` you don't need the fsspec mapper (because the kerchunk backend engine creates the mapper for you.

Soon I want to update these top-level examples to use icechunk instead though, which would just be

```python
xr.open_zarr(icechunkstore)
```

norlandrhagen:
Merging this for now, with the goal of replacing the default examples with `icechunk`

Issue #274, opened 2024-10-25
mypy spontaneously reporting new error


sharkinsspatial:
`mypy` has begun reporting a new error without any changes to the underlying code or from what I can tell any changes in the versions of the `mypy` packages installed.

```
virtualizarr/readers/common.py:34: error: Cannot assign to a type  [misc]
virtualizarr/readers/common.py:34: error: Incompatible types in assignment (expression has type "<typing special form>", variable has type "type[DataTree]")  [assignment]
virtualizarr/readers/common.py:37: error: Definition of "__array__" in base class "ManifestArray" is incompatible with definition in base class "ExplicitlyIndexed"  [misc]
```

This error just began appearing in the [Github action](https://github.com/zarr-developers/VirtualiZarr/actions/runs/11509596589/job/32039982955) and I am unable to reproduce it locally.

TomNicholas:
Maybe same as https://github.com/pydata/xarray/issues/9624?

PR #278, opened 2024-10-28
Fix bug in RT of parquet detection

norlandrhagen:


norlandrhagen:
Typing: https://github.com/zarr-developers/VirtualiZarr/issues/274

virtualizarr/readers/kerchunk.py
```diff
@@ -38,7 +38,7 @@ def open_virtual_dataset(
         fs = _FsspecFSFromFilepath(filepath=filepath, reader_options=reader_options)
 
         # The kerchunk .parquet storage format isn't actually a parquet, but a directory that contains named parquets for each group/variable.
-        if fs.filepath.endswith("ref.parquet"):
+        if fs.filepath.endswith(".parquet"):
```
keewis, line 41: according to https://github.com/fsspec/kerchunk/issues/519#issuecomment-2429763457, you could also check for the presence of `f"{fs.filepath}/.zmetadata"`, which should avoid trying to open actual parquet files – the only downside is a additional request, but since we already check magic bytes elsewhere I don't think that's a concern:
```suggestion
        if fs.filepath.endswith(".parquet") and fs.fs.isfile(f"{fs.filepath}/.zmetadata"):
```
(or define `isfile` / `exists` on `_FsspecFSFromFilepath` and use that instead of `fs.fs.isfile`)

cc @martindurant

norlandrhagen: Thanks @keewis! 

> Also, do we want to support .parq, as well, or do we expect people to always end the filename with .parquet? If we do want to support .parq, you can use fs.filepath.endswith((".parquet", ".parq")) (not that I'd need that myself).

I don't have really strong opinions on this. Adding more file name guessing doesn't feel great to me. I could see people naming files with a `.pqt` suffix since it's really just a directory name we're looking at. Maybe we just stick with `.parquet` and mention in the docs and/or in the ValueError?

cc @TomNicholas

keewis: fine with me!

martindurant: I don't have much of an opinion. Explicit is better than guessing, but clear documentation and error message either way.

norlandrhagen: Thanks both! I added a bit to the error message.

virtualizarr/readers/kerchunk.py
```diff
@@ -61,7 +63,7 @@ def open_virtual_dataset(
 
         else:
             raise ValueError(
-                "The input Kerchunk reference did not seem to be in Kerchunk's JSON or Parquet spec: https://fsspec.github.io/kerchunk/spec.html. The Kerchunk format autodetection is quite flaky, so if your reference matches the Kerchunk spec feel free to open an issue: https://github.com/zarr-developers/VirtualiZarr/issues"
+                "The input Kerchunk reference did not seem to be in Kerchunk's JSON or Parquet spec: https://fsspec.github.io/kerchunk/spec.html. If your Kerchunk generated reference file is in parquet, make sure the file extension is `.parquet`. The Kerchunk format autodetection is quite flaky, so if your reference matches the Kerchunk spec feel free to open an issue: https://github.com/zarr-developers/VirtualiZarr/issues"
```
TomNicholas, line 66:
```suggestion
                "The input Kerchunk reference did not seem to be in Kerchunk's JSON or Parquet spec: https://fsspec.github.io/kerchunk/spec.html. If your Kerchunk generated references are saved in parquet format, make sure the file extension is `.parquet`. The Kerchunk format autodetection is quite flaky, so if your reference matches the Kerchunk spec feel free to open an issue: https://github.com/zarr-developers/VirtualiZarr/issues"
```

TomNicholas (approved):

Issue #273, opened 2024-10-25
dimension_names vs _ARRAY_DIMENSIONS


TomNicholas:
I think we're doing this wrong currently, see https://github.com/pydata/xarray/pull/9669

rabernat:
I was just about to open this exact issue. 

[Everywhere that `_ARRAY_DIMENSIONS`](https://github.com/search?q=repo%3Azarr-developers%2FVirtualiZarr%20_ARRAY_DIMENSIONS&type=code) is used will now need a different code path for V3 data.

maxrjones:
This is solved because `_ARRAY_DIMENSIONS` currently only exists in the kerchunk code, which uses Zarr V2 metadata.

PR #284, opened 2024-11-05
Bump minimum Xarray dependency to 2024.10.0

TomNicholas:
This bumps the minimum required version of xarray to the most recent release, which includes support for zarr-python v3 and for `xr.DataTree`. This shouldn't break compatibility with anything.


TomNicholas:
Okay interestingly this falls afoul of some kind of bug clearly related to https://github.com/zarr-developers/VirtualiZarr/issues/273

Issue #291, opened 2024-11-08
open_virtual_dataset fails to open tiffs


elyall:
`open_virtual_dataset` says it supports TIFF files however it currently fails and is lacking test coverage. The issue starts [here](https://github.com/zarr-developers/VirtualiZarr/blob/4ae7a19c2bb2996dc4739dfd3ebbe32b17ac1658/virtualizarr/readers/tiff.py#L48) where the kerchunk reference to a zarr array is treated as an HDF reference that has one or more groups.

<details>
<summary>Code to reproduce</summary>

```python
from pathlib import Path

import numpy as np
from PIL import Image
from virtualizarr import open_virtual_dataset
from virtualizarr.backend import FileType


def write_random_tiff(file_path):
    array = np.random.randint(0, 255, (128, 128), dtype=np.uint8)
    img = Image.fromarray(array)
    img.save(file_path)


file_path = Path.cwd() / "test.tif"
write_random_tiff(file_path)

open_virtual_dataset(str(file_path), filetype=FileType.tiff)
```

</details>

Throws:
```shell
File virtualizarr/translators/kerchunk.py:46, in extract_group(vds_refs, group)

ValueError: Multiple HDF Groups found. Must specify group= keyword to select one of []
```

---
virtualizarr = 1.1.0

TomNicholas:
Ah sorry about this @elyall !

We do have some rudimentary tests for tif but as the tiff reader is just a thin wrapper around the kerchunk reader I (naively) expected this to "just work".

Thanks for the reproducible example, I'll have a look at what might be going on now.

TomNicholas:
So I've looked into this, and I'm a bit confused by what kerchunk returns, and am ignorant of exactly what information is stored in a tiff file. The kerchunk tiff reader basically returns this information about your example tiff file:

```
{'refs': {'.zattrs': '{"_ARRAY_DIMENSIONS":["Y","X"]}', '.zarray': '{\n "chunks": [\n  128,\n  128\n ],\n "compressor": null,\n "dtype": "|u1",\n "fill_value": 0,\n "filters": null,\n "order": "C",\n "shape": [\n  128,\n  128\n ],\n "zarr_format": 2\n}', '0.0': ['/private/var/folders/zt/b84f41811p5_vygj1gjp6qx40000gn/T/pytest-of-tom/pytest-38/test_random_tiff0/rand.tiff', 122, 16384]}}
```

[...]

PR #292, opened 2024-11-09
Fix tiff reader

TomNicholas:
WIP to address problems raised in #291. Adds the example there as a test case.

- [x] Closes #xxxx
- [x] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -120,3 +120,16 @@ def simple_netcdf4(tmpdir):
     ds.to_netcdf(filepath)
 
     return filepath
+
+
+@pytest.fixture
+def random_tiff(tmpdir):
+    from PIL import Image
+
+    array = np.random.randint(0, 255, (128, 128), dtype=np.uint8)
+    img = Image.fromarray(array)
+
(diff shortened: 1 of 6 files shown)


conftest.py
```diff
@@ -120,3 +120,16 @@ def simple_netcdf4(tmpdir):
...
+
+@pytest.fixture
+def random_tiff(tmpdir):
+    from PIL import Image
+
+    array = np.random.randint(0, 255, (128, 128), dtype=np.uint8)
+    img = Image.fromarray(array)
+
+    filepath = tmpdir / "rand.tiff"
+    img.save(filepath)
```
elyall, line 133: The following will save a multi-page TIFF:
```suggestion
    array = np.random.randint(0, 255, (5, 128, 128), dtype=np.uint8)
    imgs = [Image.fromarray(frame) for frame in array]

    filepath = tmpdir / "rand.tiff"
    imgs[0].save(filepath, save_all=True, append_images=imgs[1:])
```

PR #295, opened 2024-11-14
Tiff zarr reader

### LINKED ISSUE #291: open_virtual_dataset fails to open tiffs

TomNicholas:
Alternative approach to #291, that uses `tifffile` directly instead of via kerchunk's `tiff_to_zarr`. However it doesn't work yet, not least because `tifffile` doesn't support zarr-python v3.

- [x] Closes #291
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/readers/zarr.py
+++ b/virtualizarr/readers/zarr.py
@@ -1,0 +1,50 @@
+import io
+
+from xarray import Variable
+import zarr
+
+from virtualizarr.zarr import ZArray
+from virtualizarr.manifests import ChunkManifest, ManifestArray
+
+
+def virtual_variable_from_zarr_array(za: zarr.Array) -> Variable:
+    """
+    Create a virtual xarray.Variable wrapping a ManifestArray from a single zarr.Array.
+    """
+
+    # TODO this only works with zarr-python v2 for now
+
+    attrs = dict(za.attrs)
+
+    # extract _ARRAY_DIMENSIONS and remove it from attrs
+    # TODO handle v3 DIMENSION_NAMES too
+    dims = attrs.pop("_ARRAY_DIMENSIONS")
+
+    zarray = ZArray(
+        shape=za.shape,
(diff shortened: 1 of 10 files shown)


virtualizarr/readers/zarr.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def chunkmanifest_from_zarr_array(za: zarr.Array) -> ChunkManifest:
+    import ujson
+
+    of2 = io.StringIO()
+
+    # TODO handle remote urls
+    za.store.write_fsspec(of2)# , url=url)
+    out = ujson.loads(of2.getvalue())
```
TomNicholas: This is the challenging part mentioned in https://github.com/zarr-developers/VirtualiZarr/issues/291#issuecomment-2471121132

PR #296, opened 2024-11-14
Clarify which features are currently available in FAQ

TomNicholas:
Clarifies which readers actually currently work, and removes tiff from the auto-detection to avoid false advertising like in #291.


TomNicholas:
hmm my emoji checkmarks and crosses aren't rendering https://virtualizarr--296.org.readthedocs.build/en/296/faq.html#how-do-virtualizarr-and-kerchunk-compare

PR #304, opened 2024-11-17
Update pkg install in docs contribution guide

douglatornell:
Changed the command to install the project for docs development to use `python -m pip`. This ensures that the Python interpreter from the activated environment is used when installing the package in editable mode.


TomNicholas:
Thanks you @douglatornell !

PR #306, opened 2024-11-18
Support downstream type checking by including py.typed

maxrjones:
Following https://peps.python.org/pep-0561/, a package must add a marker file named `py.typed` to their package to suppoer type checking of their code. I'd like to use VirtualiZarr's type checking in https://github.com/developmentseed/virtualize-nex-gddp-cmip6 and expect others would find this helpful, so this PR adds the marker file to the package.


TomNicholas (approved):
Oh good catch! This is worth a note in the release notes.

PR #309, opened 2024-11-19
Split optional dependencies in pyproject.toml

TomNicholas:
Takes the splitting of optional dependencies idea started in https://github.com/zarr-developers/VirtualiZarr/pull/87 to its logical conclusion.


pyproject.toml
```diff
@@ -23,30 +23,59 @@ dynamic = ["version"]
...
+    "virtualizarr[fits]",
+#    "virtualizarr[tiff]",
+#    "virtualizarr[grib]",
+#    "virtualizarr[zarr]",
+]
+# writers
+icechunk = ["icechunk"]
+all = [
+    "virtualizarr[all_readers]",
+    "virtualizarr[icechunk]",
```
TomNicholas, line 74: Right now I don't think this can possibly work, because icechunk requires zarr-python v3 but kerchunk requires <3.

maxrjones:
I could change the CI to use pip. I have been using UV a bit recently and it's been super promising for managing deps, but would be a bigger change.

cisaacstern:
without much context, obligatory [pixi](https://pixi.sh/latest/) endorsement comment

TomNicholas:
I think the CI failure is because the mypy CI job is apparently the only one that uses pip to install currently

norlandrhagen:
> Also @norlandrhagen `fastparquet` is in the CI but doesn't seem to be used anywhere?

I think it's a hidden requirement of using the Kerchunk parquet format.

TomNicholas:
> I think it's a hidden requirement of using the Kerchunk parquet format.

So does that mean we should have another writer split out in these dependencies?

pyproject.toml
maxrjones: not without truly living on the bleeding edge 😅 

https://github.com/developmentseed/virtualize-nex-gddp-cmip6/blob/f331421ee87adcfb42cb4a1904c05ec1214eb932/pyproject.toml#L24

maxrjones:

- If you need Conda-Forge in addition to PyPI distributions, also use [pixi](https://pixi.sh/latest/)

pyproject.toml
maxrjones: > Not sure we should change the public pip install to use some random branch...

sorry, I wasn't suggesting you should use that approach, but just was sharing for information about what is/isn't possible. I should've been more clear. 

> Aside: Is there a reason you are using virtualizarr main there? Would it be helpful to issue another release?

I wanted strict type hinting (https://github.com/zarr-developers/VirtualiZarr/pull/306)

norlandrhagen:
> So does that mean we should have another writer split out in these dependencies?

Probably! Also, maybe for reading existing Kerchunk parquet refs.

TomNicholas:
> Probably! Also, maybe for reading existing Kerchunk parquet refs.

Does reading existing Kerchunk parquet refs require the same additional dependencies? If so we could have `kerchunk_format = ["fastparquet"]` as another group?

norlandrhagen:
> Does reading existing Kerchunk parquet refs require the same additional dependencies? If so we could have `kerchunk_format = ["fastparquet"]` as another group?


I just checked, both reading the writing of Kerchunk Parquet refs require `fastparquet`.

TomNicholas:
In https://github.com/zarr-developers/VirtualiZarr/pull/309/commits/a633a0cadd7aca5ef5bd753f70ccd293f408b568 I embedded the `pyproject.toml` into the installation docs page so that the list of pip-installable optional dependencies always stays up-to-date. See

https://virtualizarr--309.org.readthedocs.build/en/309/installation.html

TomNicholas:
note to self: add various parallel options to the splitting here once #349 is in, because that adds optional dependencies on lithops and dask

pyproject.toml
TomNicholas: > Right now I don't think this can possibly work, because icechunk requires zarr-python v3 but kerchunk requires <3.

Now that #412 is in this can work again, so we should push this PR forward too. But I'm not sure whether we should do this one before or after your Pixi PR #407 @maxrjones ?

maxrjones:
FYI there's some redundancy between this PR and https://github.com/zarr-developers/VirtualiZarr/pull/407

TomNicholas:
> FYI there's some redundancy between this PR and https://github.com/zarr-developers/VirtualiZarr/pull/407

Yep - which one shall we try to merge first? I think this one can be done very quickly, but I don't know if it will fully work without what you're doing in #407?

maxrjones:
> > FYI there's some redundancy between this PR and #407
> 
> Yep - which one shall we try to merge first? I think this one can be done very quickly, but I don't know if it will fully work without what you're doing in #407?

I recommend merging this first. It should not depend on anything in #407.

pyproject.toml
```diff
@@ -22,41 +22,80 @@ dynamic = ["version"]
...
     "h5py",
     "hdf5plugin",
     "imagecodecs",
     "imagecodecs-numcodecs==2024.6.1",
-    "numcodecs"
+]
+# kerchunk-based readers
+hdf5 = [
+    "virtualizarr[remote]",
+    "kerchunk>=0.25.0",
```
maxrjones, line 49:
```suggestion
    "kerchunk>=0.2.5",
```
Typo?

pyproject.toml
```diff
@@ -22,41 +22,80 @@ dynamic = ["version"]
...
+# un-implemented readers
+# tiff = [
+#     "virtualizarr[remote]",
+#     "kerchunk>=0.2.8",
+#     "tifffile",
+# ]
+# grib = [
+#     "virtualizarr[remote]",
+#     "kerchunk>=0.2.8",
+# ]
```
TomNicholas, line 71: These commented-out lines are going to appear in the docs - https://virtualizarr--309.org.readthedocs.build/en/309/installation.html#optional-dependencies.

Should I (a) remove them or (b) add issue links to encourage people to contribute those readers?

maxrjones: I think we should remove them because I don't think it's a certainty that those readers will be built based on kerchunk and its dependencies. e.g., I have it on my roadmap for the next 2 months to implement a reader based on https://github.com/developmentseed/aiocogeo-rs/.

PR #315, opened 2024-11-23
Add reader_kwargs argument to open_virtual_dataset

TomNicholas:
For https://github.com/zarr-developers/VirtualiZarr/pull/243#issuecomment-2492341326 we need to pass a `fs_root` kwarg down to some but not all readers. We will have other reader-specific kwargs like this come up in future too. To be in keeping with `xarray.open_dataset`'s API these should be passed as a dict called `reader_kwargs` or `backend_kwargs`.

However there is an existing `reader_options` kwarg, which is meant to be specifically for `fsspec` options. Having both `reader_options` and `reader_kwargs` would be confusing, so should we
a) rename `reader_options` to `reader_kwargs`, having it include fsspec options too,
b) have `reader_kwargs` and `fsspec_kwargs`, deprecating `reader_options`?

So far this PR just adds a `reader_kwargs` option to all backends, but raises if it is used.

cc @norlandrhagen @sharkinsspatial 

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/backend.py
+++ b/virtualizarr/backend.py
@@ -113,6 +113,7 @@ def open_virtual_dataset(
     cftime_variables: Iterable[str] | None = None,
     indexes: Mapping[str, Index] | None = None,
     virtual_array_class=ManifestArray,
+    reader_kwargs: Optional[dict] = None,
     reader_options: Optional[dict] = None,
     backend: Optional[VirtualBackend] = None,
 ) -> Dataset:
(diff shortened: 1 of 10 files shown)


TomNicholas:
I've changed this PR to use `virtual_backend_kwargs` instead of `reader_kwargs`, which is closer to xarray's equivalent `backend_kwargs`, and also makes it less similar to `reader_options`. We should discuss this choice before implementing #245 though (as then it will become harder to change).

PR #318, opened 2024-11-24
Add dmrpp relative path integration test

ayushnag:
Adds an integration test using temp files to the dmrpp tests. cc @TomNicholas 

- [x] Tests added
- [x] Tests passing

PR #323, opened 2024-12-03
Refactor dmrpp tests to expose data file path

TomNicholas:
- [x] Supercedes #318

cc @ayushnag

### DIFF
--- a/virtualizarr/tests/test_readers/test_dmrpp.py
+++ b/virtualizarr/tests/test_readers/test_dmrpp.py
@@ -21,151 +21,150 @@ urls = [
     # TODO: later add MUR, SWOT, TEMPO and others by using kerchunk JSON to read refs (rather than reading the whole netcdf file)
 ]
 
-
-BASIC_DMRPP_XML_STR = textwrap.dedent(
-    """\
-    <?xml version="1.0" encoding="ISO-8859-1"?>
-    <Dataset xmlns="http://xml.opendap.org/ns/DAP/4.0#" xmlns:dmrpp="http://xml.opendap.org/dap/dmrpp/1.0.0#" dapVersion="4.0" dmrVersion="1.0" name="test.dmrpp">
-        <Dimension name="x" size="720"/>
-        <Dimension name="y" size="1440"/>
-        <Dimension name="z" size="3"/>
-        <Int32 name="x">
-            <Dim name="/x"/>
-            <Attribute name="long_name" type="String">
-                <Value>grid x-axis</Value>
-            </Attribute>
-            <dmrpp:chunks fillValue="-2147483647" byteOrder="LE">
-                <dmrpp:chunk offset="41268" nBytes="4"/>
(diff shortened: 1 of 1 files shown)


virtualizarr/tests/test_readers/test_dmrpp.py
```diff
@@ -20,153 +20,159 @@
...
+}
+
+
+def dmrparser(dmrpp_xml_str: str, tmp_path: Path, filename="test.nc") -> DMRParser:
+    # TODO we should actually create a dmrpp file in a temporary directory
+    # this would avoid the need to pass tmp_path separately
+
+    return DMRParser(
+        root=ET.fromstring(dmrpp_xml_str), data_filepath=tmp_path / filename
+    )
```
TomNicholas, line 175: Turns out right now there is no need for this to be a fixture - fixtures are basically callables that provide something pre-packaged whilst also managing setup and teardown. But with `tmp_path` explicitly passed in at test execution time there is nothing to setup and teardown in here, so it can just be a function instead of a fixture.

virtualizarr/tests/test_readers/test_dmrpp.py
```diff
@@ -320,36 +348,44 @@ def test_parse_filters(basic_dmrpp, var_path, dtype, expected_filters):
...
+            (1,),
             np.array([4], dtype=np.uint64),
             np.array([41276], dtype=np.uint64),
-            np.array(["test.dmrpp"], dtype=np.dtypes.StringDType),
         ),
     ],
 )
 def test_parse_chunks(
-    basic_dmrpp,
+    tmp_path,
```
TomNicholas, line 370: This test has the same kind of structure as the `open_virtual_dataset` tests I want to write in #243 - because both input test data (the parser) and the expected results need to know about the `tmp_path`.

virtualizarr/tests/test_readers/test_dmrpp.py
```diff
@@ -20,153 +20,159 @@
...
-            </dmrpp:chunks>
-        </Int32>
-        <Int32 name="b">
-            <Dim name="/b"/>
-            <dmrpp:chunks fillValue="-2147483647" byteOrder="LE">
-                <dmrpp:chunk offset="41268" nBytes="4"/>
-            </dmrpp:chunks>
-        </Int32>
-        <Group name="group1">
+DMRPP_XML_STRINGS = {
```
TomNicholas, line 23: Again these don't have setup or teardown, so each string doesn't really need its own fixture.

virtualizarr/tests/test_readers/test_dmrpp.py
ayushnag: Sounds good! I just needed them to be separate so I can use different dmrpp XMLs to test different cases. However this dictionary approach makes more sense!

ayushnag:
Looks good to me! Nice refactor of the two pytest fixtures

Issue #345, opened 2024-12-12
open_virtual_mfdataset


TomNicholas:
Should we add a top-level function `open_virtual_mfdataset`?

Though I would like to be more confident about the best way to parallelize reference generation first

https://github.com/zarr-developers/VirtualiZarr/issues/123

https://github.com/zarr-developers/VirtualiZarr/issues/7

TomNicholas:
Looking at the implementation of `xr.open_mfdataset`, I'm becoming optimistic that we can parallelize reference generation for concatenation along any number of dimensions just using `dask.delayed` or `lithops.map`. Basically we just use the same [machinery](https://github.com/pydata/xarray/blob/3bc7a883059f8a82f59bae3a7a4bafda1dd6dfa3/xarray/backends/api.py#L1551) xarray does for organising the `open_dataset` calls into a 1d list, optionally map over that in parallel, return the (small) virtual datasets back to the client over the network, then the multidimensional concatenation occurs on the client.

TomNicholas:
An implementation of `open_mfdataset` was added in #349, but I won't close this yet because there are still some todos:

- [ ] Test that the lithops / dask parallelization actually works for non-local executors (see https://github.com/zarr-developers/VirtualiZarr/pull/349#discussion_r2019902793)
- [ ] Test that this whole approach actually scales the way I'm hoping it will (see #123)
- [ ] Add function to public API
- [ ] Add documentation of how to use the function in the user guide
- [ ] Potentially fix the lithops bug upstream so we don't have to dodge it (see https://github.com/zarr-developers/VirtualiZarr/pull/349#discussion_r2013060076)
- [ ] Potentially upstream parts of the code I wrote in #349 into xarray

TomNicholas:
With #590 I think this can be closed. There are a couple of small known bugs but the function exists, is documented, and works.

PR #349, opened 2024-12-16
Add open_virtual_mfdataset

TomNicholas:
Here I have copied the code from `xr.open_mfdataset`, changed it to use `open_virtual_dataset`, and added an option to parallelize with lithops as an alternative to using `dask.delayed`.

I haven't even tried to run this yet, but I _think_ this is the right approach @tomwhite? I realised we don't need cubed's blockwise because xarray.open_mfdataset has internal logic to turn the N-dimensional concat into a 1D map already, so `lithops.map` should be fine? 

Also I think based on our conversation we should be able to use `lithops.map` instead of `lithops.map_reduce` like @thodson-usgs did in #203 because the tiny size of the virtual datasets being returned to the client means that we should be able to get away with a single reduction step on the client even at large scale? (see also https://github.com/zarr-developers/VirtualiZarr/issues/104 for justification that we only need to send back kB-sized objects).

- [x] Adds `open_virtual_mfdataset` as suggested in #345, but also sketches out how we might close both #95 and #123
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage
- [ ] ~~Make it a context manager~~


virtualizarr/backend.py
```diff
@@ -209,3 +228,211 @@ def open_virtual_dataset(
...
+        closers = [getattr_(ds, "_close") for ds in datasets]
+        if preprocess is not None:
+            datasets = [preprocess(ds) for ds in datasets]
+
+        # calling compute here will return the datasets/file_objs lists,
+        # the underlying datasets will still be stored as dask arrays
+        datasets, closers = dask.compute(datasets, closers)
+    elif parallel == "lithops":
+
+        def generate_refs(path):
```
TomNicholas, line 374: This is the equivalent of @thodson-usgs 's `map_references` function 

https://github.com/zarr-developers/VirtualiZarr/blob/1dbd119993a2f900142ee8af0ab34057f28c3831/examples/virtualizarr-with-lithops/virtualizarr-with-lithops.py#L25

virtualizarr/backend.py
```diff
@@ -209,3 +228,211 @@ def open_virtual_dataset(
...
+            # allows passing the open_virtual_dataset function to lithops without evaluating it
+            vds = open_(path, **kwargs)
+            # TODO perhaps we should just load the loadable_vars here and close before returning?
+            return vds
+
+        futures = fn_exec.map(generate_refs, paths1d)
+
+        # wait for all the serverless workers to finish, and send their resulting virtual datasets back to the client
+        completed_futures, _ = fn_exec.wait(futures, download_results=True)
+        virtual_datasets = [future.get_result() for future in completed_futures]
```
TomNicholas, line 384: IIUC this will cause every serverless worker to send a small virtual dataset back to the client process over the internet somehow

virtualizarr/backend.py
```diff
@@ -209,3 +228,211 @@ def open_virtual_dataset(
...
+                data_vars=data_vars,
+                coords=coords,
+                ids=ids,
+                join=join,
+                combine_attrs=combine_attrs,
+            )
+        elif combine == "by_coords":
+            # Redo ordering from coordinates, ignoring how they were ordered
+            # previously
+            combined = combine_by_coords(
```
TomNicholas, line 409: This is only going to work if we have used `loadable_variables` to create indexes for 1D coordinates, so it's a good reason to implement the suggestion in https://github.com/zarr-developers/VirtualiZarr/issues/335#issuecomment-2525405929

virtualizarr/backend.py
```diff
@@ -1,13 +1,23 @@
...
     Optional,
+    Sequence,
+    cast,
 )
 
-from xarray import Dataset, Index
+from xarray import DataArray, Dataset, Index, combine_by_coords
+from xarray.backends.api import _multi_file_closer
+from xarray.backends.common import _find_absolute_paths
+from xarray.core.combine import _infer_concat_order_from_positions, _nested_combine
```
TomNicholas, line 20: I don't like importing these deep xarray internals like this (though `_infer_concat_order_from_positions` and `_nested_combine` haven't changed since [I wrote them 6 years ago](https://github.com/pydata/xarray/pull/2553)), but the only alternative would be to make a general virtualizarr backend engine for xarray (see https://github.com/zarr-developers/VirtualiZarr/issues/35).

virtualizarr/backend.py
```diff
@@ -209,3 +228,211 @@ def open_virtual_dataset(
...
+            preprocess = dask.delayed(preprocess)
+    elif parallel == "lithops":
+        import lithops
+
+        # TODO use RetryingFunctionExecutor instead?
+        # TODO what's the easiest way to pass the lithops config in?
+        fn_exec = lithops.FunctionExecutor()
+
+        # lithops doesn't have a delayed primitive
+        open_ = open_virtual_dataset
```
TomNicholas, line 352: I think the code would be more straightforward if the parallel primitive we used for lithops was the same as the once we used for dask.

virtualizarr/backend.py
```diff
@@ -209,3 +228,211 @@ def open_virtual_dataset(
...
+    if parallel == "dask":
+        import dask
+
+        # wrap the open_dataset, getattr, and preprocess with delayed
+        open_ = dask.delayed(open_virtual_dataset)
+        getattr_ = dask.delayed(getattr)
+        if preprocess is not None:
+            preprocess = dask.delayed(preprocess)
+    elif parallel == "lithops":
+        import lithops
```
TomNicholas, line 345: I believe all of this could also be useful upstream in `xr.open_mfdataset`

tomwhite:
> I realised we don't need cubed's blockwise because xarray.open_mfdataset has internal logic to turn the N-dimensional concat into a 1D map already, so `lithops.map` should be fine?

Yes, that should work fine. We may want to loosen/generalize blockwise slightly in Cubed to return an arbitrary object so it can be done with Cubed - but that can be done later.

> Also I think based on our conversation we should be able to use `lithops.map` instead of `lithops.map_reduce` like @thodson-usgs did in #203 because the tiny size of the virtual datasets being returned to the client means that we should be able to get away with a single reduction step on the client even at large scale? (see also #104 for justification that we only need to send back kB-sized objects).

Agreed - it will be interesting to see this for large datasets. (It's also similar to the approach I've taken for [storing data in Icechunk](https://github.com/cubed-dev/cubed/pull/633) where the changesets are returned to the client - again, small kB-sized UUIDs.)

virtualizarr/tests/test_backend.py
```diff
@@ -438,3 +440,114 @@ def test_open_dataset_with_scalar(self, hdf5_scalar, hdf_backend):
...
+        print(combined_vds["lat"].data.zarray)
+        print(expected_vds["lat"].data.zarray)
+        print(combined_vds["lat"].data.manifest.dict())
+        print(expected_vds["lat"].data.manifest.dict())
+
+        # TODO this assertion unintentially triggers loading, see issue #354
+        # xrt.assert_identical(combined_vds.coords.variables['lat'], expected_vds.coords.variables['lat'])
+
+        # TODO I have no idea why this assertion fails for all the coords - everything about the coords looks identical
+        # xrt.assert_identical(combined_vds, expected_vds)
```
TomNicholas, line 529: I'm stuck on why this `assert_identical` on the datasets doesn't pass. It complains that all the coordinates are different, but every attribute I can think to check looks identical 😖

docs/api.rst
```diff
@@ -19,7 +19,7 @@ Reading
     :toctree: generated/
 
     open_virtual_dataset
-
+    open_virtual_mfdataset
```
TomNicholas, line 22: Note to self: the docs and especially the readme should be rewritten to put this function front and center.

virtualizarr/backend.py
TomNicholas: See https://github.com/pydata/xarray/pull/9932

virtualizarr/backend.py
TomNicholas: See https://github.com/pydata/xarray/pull/9932#issuecomment-2578523022

virtualizarr/tests/test_backend.py
TomNicholas: Found it - should be fixed by

https://github.com/zarr-developers/VirtualiZarr/pull/502

virtualizarr/backend.py
TomNicholas: This was done in #477!

virtualizarr/tests/test_backend.py
```diff
@@ -502,3 +506,66 @@ def test_open_dataset_with_scalar(self, hdf5_scalar, hdf_backend):
...
+    [
+        False,
+        ThreadPoolExecutor,
+        pytest.param(DaskDelayedExecutor, marks=requires_dask),
+        pytest.param(
+            LithopsEagerFunctionExecutor,
+            marks=[
+                requires_lithops,
+                pytest.mark.xfail(
+                    reason="Lithops bug - see https://github.com/lithops-cloud/lithops/issues/1428"
```
TomNicholas, line 525: See https://github.com/lithops-cloud/lithops/issues/1428

virtualizarr/parallel.py
```diff
@@ -0,0 +1,287 @@
...
+        # For Dask.delayed, shutdown is essentially a no-op
+        pass
+
+
+class LithopsEagerFunctionExecutor(Executor):
+    """
+    Lithops-based function executor which follows the concurrent.futures.Executor API.
+
+    Only required because lithops doesn't follow the concurrent.futures.Executor API, see https://github.com/lithops-cloud/lithops/issues/1427.
+    """
```
TomNicholas, line 192: See https://github.com/lithops-cloud/lithops/issues/1427

virtualizarr/tests/test_backend.py
TomNicholas: I was able to dodge this by not using `functools.partial`, but I'm a bit worried that my newer approach of using a closure won't work properly in a remote execution context.

pyproject.toml
```diff
@@ -119,9 +119,16 @@ dev = [
     "pytest-asyncio",
     "pytest-cov",
     "pytest-mypy",
+    "pytest",
+    "pytest-asyncio",
```
TomNicholas, line 123:
```suggestion
```

pyproject.toml
```diff
@@ -119,9 +119,16 @@ dev = [
...
     "pytest-xdist",
     "ruff",
     "s3fs",
+    "scipy",
+    "lithops",
+    "dask",
+    "virtualizarr[hdf]",
+    "obstore>=0.5.1",
```
TomNicholas, line 131:
```suggestion
```

pyproject.toml
```diff
@@ -119,9 +119,16 @@ dev = [
     "pytest-asyncio",
     "pytest-cov",
     "pytest-mypy",
+    "pytest",
+    "pytest-asyncio",
     "pytest-xdist",
     "ruff",
     "s3fs",
+    "scipy",
```
TomNicholas, line 127:
```suggestion
```

virtualizarr/backend.py
```diff
@@ -66,6 +86,7 @@ def automatically_determine_filetype(
     # TODO this should ideally handle every filetype that we have a reader for, not just kerchunk
 
     # TODO how do we handle kerchunk json / parquet here?
+    print(f"{filepath=}")
```
TomNicholas, line 89:
```suggestion
```

PR #353, opened 2024-12-17
Load 1d coord vars by default

### LINKED ISSUE #335: Add loadable_variables='coords' option
> > is there a reason one would not want to always include all coordinates in the `loadable_variables` 
> 
> The main reason though is that coordinates can be N-dimensional in general, in which case you might use a lot of storage duplicating them. But for low-dimensional coordinates I agree we should recommend people load them.
> 
> Perhaps we should even add another option: `loadable_variables='coords'` or `loadable_variables='1d_coords'`, which we could default to... The latter is pretty much what `xr.open_dataset` already does for choosing which coordinates to create indexes for.

 _Originally posted by @TomNicholas in [#330](https://github.com/zarr-developers/VirtualiZarr/issues/330#issuecomment-2524060745)_

TomNicholas:
Adds two options for `loadable_variables`: `1d_coord_dims` and `all_coords`, and changes the default to `1d_coord_dims`.

- [x] Closes #335
- [x] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/backend.py
+++ b/virtualizarr/backend.py
@@ -21,7 +21,7 @@ from virtualizarr.readers import (
     ZarrV3VirtualBackend,
 )
 from virtualizarr.readers.common import VirtualBackend
-from virtualizarr.utils import _FsspecFSFromFilepath, check_for_collisions
+from virtualizarr.utils import _FsspecFSFromFilepath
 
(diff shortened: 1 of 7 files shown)


TomNicholas:
Closing in favour of #477

PR #362, opened 2024-12-20
Configure codecov to ignore tests folder

maxrjones:
Based on Zarr-Python, I think it should work to put the configuration inside .github rather than the repository root, but let's see...

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,17 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 100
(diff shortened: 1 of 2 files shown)


maxrjones:
Looks like we're not getting codecov uploads for all the commits - https://app.codecov.io/gh/zarr-developers/VirtualiZarr/commits/All%20branches 🚀

TomNicholas (approved):
I'm not much of a CI wizard but if you think this looks good then feel free to merge it!

PR #368, opened 2025-01-02
Sync with icechunk alpha 8

mpiannucci:
- [ ] Closes #xxxx
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -421,18 +421,12 @@ By default references are placed in separate parquet file when the total number 
 We can also write these references out as an [IcechunkStore](https://icechunk.io/). `Icechunk` is a Open-source, cloud-native transactional tensor storage engine that is compatible with zarr version 3. To export our virtual dataset to an `Icechunk` Store, we simply use the {py:meth}`vds.virtualize.to_icechunk <virtualizarr.VirtualiZarrDatasetAccessor.to_icechunk>` accessor method.
 
 ```python
-# create an icechunk repository, with access to s3 for virtual data
-from icechunk import Repository, Storage, VirtualChunkContainer, local_filesystem_storage, s3_credentials, s3_store
+# create an icechunk repository, session and write the virtual dataset to the session
+from icechunk import Repository, Storage, VirtualChunkContainer, local_filesystem_storage
 storage = local_filesystem_storage(str('combined'))
-config = RepositoryConfig.default()
-container = VirtualChunkContainer("s3", "s3://", s3_store(
-    region="us-east-1",
-))
-config.set_virtual_chunk_container(container)
-credentials = containers_credentials(
-        s3=s3_credentials(anonymous=True)
-    )
-repo = Repository.create(storage=storage, config=config, credentials=credentials)
+
+# By default, local virtual references and public remote virtual references can be read wihtout extra configuration.
+repo = Repository.create(storage=storage)
 session = repo.writeable_session("main")
 
 # write the virtual dataset to the session with the IcechunkStore
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -421,14 +421,22 @@ By default references are placed in separate parquet file when the total number 
 We can also write these references out as an [IcechunkStore](https://icechunk.io/). `Icechunk` is a Open-source, cloud-native transactional tensor storage engine that is compatible with zarr version 3. To export our virtual dataset to an `Icechunk` Store, we simply use the {py:meth}`vds.virtualize.to_icechunk <virtualizarr.VirtualiZarrDatasetAccessor.to_icechunk>` accessor method.
 
 ```python
-# create an icechunk store
-from icechunk import IcechunkStore, StorageConfig, StoreConfig, VirtualRefConfig
-storage = StorageConfig.filesystem(str('combined'))
-store = IcechunkStore.create(storage=storage, mode="w", config=StoreConfig(
-    virtual_ref_config=VirtualRefConfig.s3_anonymous(region='us-east-1'),
+# create an icechunk repository, with access to s3 for virtual data
+from icechunk import Repository, Storage, VirtualChunkContainer, local_filesystem_storage, s3_credentials, s3_store
+storage = local_filesystem_storage(str('combined'))
+config = RepositoryConfig.default()
+container = VirtualChunkContainer("s3", "s3://", s3_store(
+    region="us-east-1",
 ))
+config.set_virtual_chunk_container(container)
+credentials = containers_credentials(
+        s3=s3_credentials(anonymous=True)
+    )
+repo = Repository.create(storage=storage, config=config, credentials=credentials)
+session = repo.writeable_session("main")
 
-combined_vds.virtualize.to_icechunk(store)
+# write the virtual dataset to the session with the IcechunkStore
+combined_vds.virtualize.to_icechunk(session.store)
 ```
 
 See the [Icechunk documentation](https://icechunk.io/icechunk-python/virtual/#creating-a-virtual-dataset-with-virtualizarr) for more details.
--- a/virtualizarr/accessor.py
+++ b/virtualizarr/accessor.py
@@ -52,8 +52,9 @@ class VirtualiZarrDatasetAccessor:
 
         If `append_dim` is provided, the virtual dataset will be appended to the existing IcechunkStore along the `append_dim` dimension.
 
-        If `last_updated_at` is provided, it will be used as a checksum for any virtual chunks written to the store with this operation. At read time, if any of the virtual chunks have been updated since this provided datetime, an error will be raised.
-            This protects against reading outdated virtual chunks that have been updated since the last read.
+        If `last_updated_at` is provided, it will be used as a checksum for any virtual chunks written to the store with this operation.
+        At read time, if any of the virtual chunks have been updated since this provided datetime, an error will be raised.
+        This protects against reading outdated virtual chunks that have been updated since the last read. When not provided, no check is performed.
 
         Parameters
         ----------
@@ -62,6 +63,7 @@ class VirtualiZarrDatasetAccessor:
             When provided, specifies the dimension along which to append the virtual dataset.
         last_updated_at: datetime, optional
             When provided, uses provided datetime as a checksum for any virtual chunks written to the store with this operation.
+            When not provided (default), no check is performed.
 
         """
         from virtualizarr.writers.icechunk import dataset_to_icechunk
--- a/virtualizarr/accessor.py
+++ b/virtualizarr/accessor.py
@@ -55,6 +55,7 @@ class VirtualiZarrDatasetAccessor:
                 If `last_updated_at` is provided, it will be used as a checksum for any virtual chunks written to the store with this operation.
                 At read time, if any of the virtual chunks have been updated since this provided datetime, an error will be raised.
                 This protects against reading outdated virtual chunks that have been updated since the last read. When not provided, no check is performed.
+                This value is stored in Icechunk with seconds precision, so be sure to take that into account when providing this value.
 
                 Parameters
                 ----------
--- a/virtualizarr/accessor.py
+++ b/virtualizarr/accessor.py
@@ -46,35 +46,35 @@ class VirtualiZarrDatasetAccessor:
         last_updated_at: Optional[datetime] = None,
     ) -> None:
         """
-        Write an xarray dataset to an Icechunk store.
+                Write an xarray dataset to an Icechunk store.
 
-        Any variables backed by ManifestArray objects will be be written as virtual references, any other variables will be loaded into memory before their binary chunk data is written into the store.
+                Any variables backed by ManifestArray objects will be be written as virtual references, any other variables will be loaded into memory before their binary chunk data is written into the store.
 
-        If `append_dim` is provided, the virtual dataset will be appended to the existing IcechunkStore along the `append_dim` dimension.
+                If `append_dim` is provided, the virtual dataset will be appended to the existing IcechunkStore along the `append_dim` dimension.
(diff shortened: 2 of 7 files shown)


virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -19,21 +19,16 @@
...
+    from icechunk import Repository, Storage
 
-    storage = StorageConfig.filesystem(str(tmpdir))
-
-    # TODO if icechunk exposed a synchronous version of .open then we wouldn't need to use asyncio.run here
-    # TODO is this the correct mode to use?
-    store = IcechunkStore.create(storage=storage, mode="w")
-
-    # TODO instead yield store then store.close() ??
-    return store
```
TomNicholas, line 36: Should we close the session when tearing down the fixture? Or does it not matter?

TomNicholas (commented):
Thanks for this!

Presumably the documentation should also be changed to match?

virtualizarr/tests/test_writers/test_icechunk.py
mpiannucci: Does not matter at this time

mpiannucci:
IDK what the errors are, they seem unrelated

virtualizarr/accessor.py
```diff
@@ -48,10 +52,17 @@ def to_icechunk(
...
+        If `last_updated_at` is provided, it will be used as a checksum for any virtual chunks written to the store with this operation. At read time, if any of the virtual chunks have been updated since this provided datetime, an error will be raised.
+            This protects against reading outdated virtual chunks that have been updated since the last read.
+
         Parameters
         ----------
         store: IcechunkStore
         append_dim: str, optional
+            When provided, specifies the dimension along which to append the virtual dataset.
+        last_updated_at: datetime, optional
+            When provided, uses provided datetime as a checksum for any virtual chunks written to the store with this operation.
```
TomNicholas, line 64: This is a cool feature! But without an example I'm unclear if the intended usage pattern is to pass the current time or the time at which the referenced archival data was written. In fact is there a reason the default value isn't just to use the current time?

An example either in the docstring (could add an `Examples` section) or in the usage docs would help.

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -282,9 +290,25 @@ def test_write_loadable_variable(
...
+    # the checksum_date is being used to determine if the data is valid
+    arr = np.arange(12, dtype=np.dtype("int32")).reshape(3, 4) * 2
+    var = Variable(data=arr, dims=["x", "y"])
+    ds = Dataset({"foo": var})
+    time.sleep(1)  # Make sure the checksum_date is at least one second in the future
+    ds.to_netcdf(netcdf_path)
+
+    # Now if we try to read the data back in, it should fail because the checksum_date
+    # is newer than the last_updated_at
+    with pytest.raises(IcechunkError):
```
TomNicholas, line 308: Upstream issue but `IcechunkError` seems like a unhelpfully-vague error type. I mean this assertion would pass if any other `IcechunkError` were raised for some unrelated reason...

Something more specific like `OutdatedVirtualReferenceError(IcechunkError)` seems like it might be more helpful?

virtualizarr/writers/icechunk.py
```diff
@@ -275,4 +291,5 @@ def write_manifest_virtual_refs(
             location=path.item(),
             offset=offset.item(),
             length=length.item(),
+            checksum=last_updated_at,
```
TomNicholas, line 294: Again an upstream issue, but isn't using a general `checksum` to pass in a time last modified kind of overloading two different things? Can we imagine a case where we wanted to use checksums to verify the integrity of the data but the last-modified was fine, or vice versa?

virtualizarr/accessor.py
TomNicholas: Also what is the behaviour when the default of `None` is passed?

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -282,9 +290,25 @@ def test_write_loadable_variable(
...
     assert pres_array.shape == (3, 4)
     assert pres_array.dtype == np.dtype("int32")
-    expected_ds = open_dataset(simple_netcdf4)
+    expected_ds = open_dataset(netcdf_path)
     expected_array = expected_ds["foo"].to_numpy()
     npt.assert_equal(pres_array, expected_array)
+    expected_ds.close()
+
+    # Now we can overwrite the simple_netcdf4 file with new data to make sure that
+    # the checksum_date is being used to determine if the data is valid
```
TomNicholas, line 299: This part of the test seems to be testing a separate feature - perhaps it should be split out into a separate test?

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -268,7 +273,10 @@ def test_write_loadable_variable(
     )
     vds = Dataset({"air": la_v}, {"pres": ma_v})
 
-    dataset_to_icechunk(vds, icechunk_filestore)
+    # Icechunk checksums currently store with second precision, so we need to make sure
+    # the checksum_date is at least one second in the future
+    checksum_date = datetime.now(timezone.utc) + timedelta(seconds=1)
+    dataset_to_icechunk(vds, icechunk_filestore, last_updated_at=checksum_date)
```
TomNicholas, line 279: We don't appear to have a test that defines the expected behaviour for the default value. It might also be good to raise an error if someone passes a value in the future (if icechunk doesn't do that already).

virtualizarr/tests/test_writers/test_icechunk.py
TomNicholas: We could also just add the whole `last_updated_at` feature in a follow-up PR.

TomNicholas (commented):
Thanks @mpiannucci ! A number of comments.

> IDK what the errors are, they seem unrelated

I think you were referring to the errors that should now be fixed in `main` by https://github.com/zarr-developers/VirtualiZarr/pull/370, but actually there are some real errors (import and typing) in this branch now too.

virtualizarr/accessor.py
mpiannucci: when none is passed there is no check. so thats a better default for those who dont want or care about this check

virtualizarr/tests/test_writers/test_icechunk.py
mpiannucci: This will come with time, but for now this is what we get. Typing all the errors is a lot of work whihc we have on the roadmap.

virtualizarr/writers/icechunk.py
mpiannucci: you can do both in icechunk, you can either pass in a datetime, or you can pass in an etag, which is a more effective checksum, but is a lot more work to write into this PR

virtualizarr/tests/test_writers/test_icechunk.py
mpiannucci: yeah icechunk tests this very comprehensively, but i figured it would be good to have at least one test that threads the values through

virtualizarr/tests/test_writers/test_icechunk.py
mpiannucci: This is mostly me being lazy and not wanting to have to duplicate stuff that is already in a test. I can split it outto its own test if you want

docs/usage.md
```diff
@@ -421,14 +421,22 @@ By default references are placed in separate parquet file when the total number
...
+config = RepositoryConfig.default()
+container = VirtualChunkContainer("s3", "s3://", s3_store(
+    region="us-east-1",
 ))
+config.set_virtual_chunk_container(container)
+credentials = containers_credentials(
+        s3=s3_credentials(anonymous=True)
+    )
+repo = Repository.create(storage=storage, config=config, credentials=credentials)
+session = repo.writeable_session("main")
```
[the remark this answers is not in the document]
mpiannucci: agree, up to you what is the most effective way to show how it works

virtualizarr/accessor.py
TomNicholas: Okay - let's just document in the docstring that there is no check then.

docs/usage.md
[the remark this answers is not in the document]
TomNicholas: I think maybe use a local store to reduce the number of lines as much as possible, then perhaps we have a separate section (or a link to the icechunk docs) about doing this with a cloud store. The important thing is that I want users to immediately see that actually writing a virtual dataset into icechunk is a one-liner.

virtualizarr/tests/test_writers/test_icechunk.py
TomNicholas: Okay that's fine. At least add more to the whatsnew noting this new feature though.

mpiannucci:
Lemme know what you think of the release notes and doc updates, we have to add tons of docs to icechunk itself later to match

virtualizarr/accessor.py
```diff
@@ -48,10 +52,19 @@ def to_icechunk(
...
+            When not provided (default), no check is performed.
+
```
TomNicholas, line 67: Will this work or does it need to be one second in the future?

```suggestion

Examples
--------
To ensure an error is raised if the files containing referenced virtual chunks are modified at any time from now on, pass the current time to ``last_updated_at``.

>>> from datetime import datetime
>>>
>>> vds.virtualize.to_icechunk(
...     icechunkstore,
...     last_updated_at=datetime.now(),
... )
```

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -327,13 +392,10 @@ def test_generate_chunk_key_append_axis_out_of_bounds():
 
 
 @pytest.fixture(scope="function")
-def icechunk_storage(tmpdir) -> "StorageConfig":
-    from icechunk import StorageConfig
-
-    storage = StorageConfig.filesystem(str(tmpdir))
+def icechunk_storage(tmpdir) -> "Storage":
+    from icechunk import Storage
```
TomNicholas, line 396: Presumably the `icechunk_filestore` fixture above could just use this fixture?

virtualizarr/accessor.py
TomNicholas: I guess alternatively we might imagine finding the time at which the files were last modified with `os.stat`, then passing that.

TomNicholas (commented):
Looking great - I just want to make sure I understand how you are actually supposed to use `last_updated_at`.

virtualizarr/writers/icechunk.py
```diff
@@ -23,7 +24,10 @@
 
 
 def dataset_to_icechunk(
-    ds: Dataset, store: "IcechunkStore", append_dim: Optional[str] = None
+    ds: Dataset,
+    store: "IcechunkStore",
+    append_dim: Optional[str] = None,
+    last_updated_at: Optional[datetime] = None,
```
TomNicholas, line 30: The type here is `datetime` or `None`, but the type upstream is more general. Should we also validate the input type here so that the user doesn't e.g. unwittingly pass a date as a string and be suprised when it doesn't behave the way they expected?

virtualizarr/accessor.py
mpiannucci: Yeah this works in practice, as does using the actual stamp of the file (typically) as long as your round up to the nearest second. Icechunk only store second level precision, so thats why the test adds a second because the tets are so fast it is invalid otherwise.

mpiannucci:
I think this is pretty good now, but im not sure why the read the docs build started failing

virtualizarr/accessor.py
```diff
@@ -39,19 +40,42 @@ def to_zarr(self, storepath: str) -> None:
...
+                store: IcechunkStore
+                append_dim: str, optional
+                    When provided, specifies the dimension along which to append the virtual dataset.
+                last_updated_at: datetime, optional
+                    When provided, uses provided datetime as a checksum for any virtual chunks written to the store with this operation.
+                    When not provided (default), no check is performed.
+
+        Examples
+        --------
+        To ensure an error is raised if the files containing referenced virtual chunks are modified at any time from now on, pass the current time to ``last_updated_at``.
```
TomNicholas, line 71: This incorrect indenting is probably what's causing the docs failure. The docs fail on any warning, and it's warning with

```
/home/docs/checkouts/readthedocs.org/user_builds/virtualizarr/conda/368/lib/python3.13/site-packages/virtualizarr/accessor.py:docstring of virtualizarr.accessor.VirtualiZarrDatasetAccessor.to_icechunk:13: CRITICAL: Unexpected section title.
```

mpiannucci: That was it! Thanks!

TomNicholas (approved):
Thanks @mpiannucci !

TomNicholas:
hmm somehow merging this caused a failure in the "upstream" CI

https://github.com/zarr-developers/VirtualiZarr/actions/runs/12623287580/job/35172119637

mpiannucci:
I can reproduce locally. I am pretty sure this is an issue with xarray 2025 not being zarr v3 compatible

TomNicholas:
> I can reproduce locally. I am pretty sure this is an issue with xarray 2025 not being zarr v3 compatible

ah, would you mind raising an issue (here or on xarray)?

mpiannucci:
Yeah im talking with the zarr folks to confirm. everythign works for `xarray<2025.1.0`

mpiannucci:
Logged, should hopefully get it all sorted early this week, sorry about that

TomNicholas:
Thank you!

Issue #343, opened 2024-12-12
Fill value lost with Icechunk


rabernat:
This issue is mostly likely related to the different treatment of missing data in Xarray with Zarr V3 vs. V2 (see https://github.com/pydata/xarray/issues/5475). I don't think it's specific to Icechunk, but rather is related to Zarr V3. However, it can only be reproduced with IC afaik because only IC uses V3 in VirtualiZarr.

Note: this requires `pip install git+https://github.com/mpiannucci/kerchunk@v3`

To reproduce, first create test data

```python
import xarray as xr
import numpy as np
import virtualizarr
from virtualizarr import open_virtual_dataset
import icechunk

ds = xr.DataArray([np.nan, 2, 3], name="foo").to_dataset()
ds.foo.encoding.update({"_FillValue": np.float32(-9999.9), "dtype": "f4"})

ds.to_netcdf("test.nc")

ds_nc = xr.open_dataset("test.nc").load()
print(ds_nc)
print(ds_nc.foo.encoding)
```

```
<xarray.Dataset> Size: 12B
Dimensions:  (dim_0: 3)
Dimensions without coordinates: dim_0
Data variables:
    foo      (dim_0) float32 12B nan 2.0 3.0
{'dtype': dtype('float32'), 'zlib': False, 'szip': False, 'zstd': False, 'bzip2': False, 'blosc': False, 'shuffle': False, 'complevel': 0, 'fletcher32': False, 'contiguous': True, 'chunksizes': None, 'source': '/home/jovyan/GPM_IMERG/test.nc', 'original_shape': (3,), '_FillValue': np.float32(-9999.9)}
```

Next, create an Icechunk virtual dataset

```python
dsv = open_virtual_dataset("test.nc", indexes={})
storage_config = icechunk.StorageConfig.filesystem("./icechunk-local")
store = icechunk.IcechunkStore.create(storage_config)
dsv.virtualize.to_icechunk(store)
store.commit("wrote store")
```

Now read it back

```python
ds_ic = xr.open_dataset(store, engine="zarr", zarr_format=3, consolidated=False).load()
print(ds_ic)
print(ds_ic.foo.encoding)
```

```
<xarray.Dataset> Size: 12B
Dimensions:  (dim_0: 3)
Dimensions without coordinates: dim_0
Data variables:
    foo      (dim_0) float32 12B -1e+04 2.0 3.0
[...]

PR #369, opened 2025-01-02
Keep fill value for encoding

abarciauskas-bgse:
While [trying to create a virtual icechunk store for MUR SST](https://gist.github.com/abarciauskas-bgse/daa92b4a4a90387afec654597c3509c6), using the dmrpp reader, I noticed that a mean value coming back from the icechunk store was wrong. (To be clear this issue has nothing to do with icechunk, but the data have to be stored to zarr or icechunk). After some trial and error, I found that the issue was a missing fill value in the encoding of the data variables. This PR fixes the issue by replacing `pop` with `get` for the `_FillValue` in attrs, so they can be passed along in the encoding.

I verified this works by making the same code change in the dmrpp.py file VirtualiZarr dependency of my notebook (run on hub.openveda.cloud and using the quay.io/developmentseed/veda-optimized-data-delivery-image:latest image). The mean value returned from the icechunk.

- ~Closes #xxxx~ n/a
- [x] Tests added
- [x] Tests passing
- ~Full type hint coverage~ n/a
- [x] Changes are documented in docs/releases.rst
- ~New functions/methods are listed in api.rst~ n/a
- ~New functionality has documentation~ n/a

Please review @ayushnag

### DIFF
--- a/ci/upstream.yml
+++ b/ci/upstream.yml
@@ -28,6 +28,6 @@ dependencies:
   - fsspec
   - pip
   - pip:
-      - icechunk>=0.1.0a8 # Installs zarr v3 as dependency
+      - icechunk==0.1.0a8 # Installs zarr v3 beta 3 as dependency
       # - git+https://github.com/fsspec/kerchunk@main  # kerchunk is currently incompatible with zarr-python v3 (https://github.com/fsspec/kerchunk/pull/516)
       - imagecodecs-numcodecs==2024.6.1
--- a/ci/upstream.yml
+++ b/ci/upstream.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - nodefaults
 dependencies:
-  - xarray>=2024.10.0
+  - xarray>=2024.10.0,<2025.0.0
   - h5netcdf
   - h5py
   - hdf5
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -24,6 +24,8 @@ Bug fixes
 
 - Fix bug preventing generating references for the root group of a file when a subgroup exists.
   (:issue:`336`, :pull:`337`) By `Tom Nicholas <https://github.com/TomNicholas>`_.
+- Fix bug in dmrpp reader so _FillValue is passed to variables' encodings.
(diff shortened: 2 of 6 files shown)


TomNicholas (approved):
I think this is right, thanks @abarciauskas-bgse

ayushnag:
@abarciauskas-bgse Thanks for reporting the bug! I'm still not sure if `_FillValue` should be in attributes since even when virtualizarr is called on a netcdf dataset, the `_FillValue` is not in the attributes. I was trying to track down where it is removed and I think it [happens in kerchunk here](https://github.com/fsspec/kerchunk/blob/013798ed4caa5c7dda958ee3fca348aecb7d64df/kerchunk/hdf.py#L193). I think perhaps the correct fix is to only remove `_FillValue` once it has been added to the variable encoding?

abarciauskas-bgse:
Thanks for looking @ayushnag. Are we sure that _FillValue _should_ be removed? I'm considering (a) I see `_FillValue` being listed as an [attribute in NetCDF conventions](https://docs.unidata.ucar.edu/netcdf-c/4.9.2/attribute_conventions.html) and (b) @sharkinsspatial and @rabernat have run into some issues where `_FillValue` is not present and has also caused issues (see https://github.com/zarr-developers/VirtualiZarr/issues/343 and also the `_FillValue`).

abarciauskas-bgse:
@mpiannucci @TomNicholas I had to pin `icechunk==0.1.0a8` as opposed to `icechunk>=0.1.0a8` as version 0.1.0a10 upgrades to zarr 3 which causes a lot of test failures. Similarly, I had to specify `xarray>=2024.10.0,<2025.0.0` as xarray `2025.1.1` also had breaking changes. I am planning to work on upgrading to `icechunk==0.1.0a10` (and thus the newly released zarr-python 3) so obviously these versions are short-lived 🤞🏽

abarciauskas-bgse:
I'm going to merge as tests are passing and don't want this to get stale. Will work on the upgrades suggested https://github.com/zarr-developers/VirtualiZarr/pull/375/files

PR #396, opened 2025-01-28
Fix drop_variables support in kerchunk, dmrpp

chuckwondo:
Fix reversed assignment of results from collision detection that was causing NotImplementedError to be incorrectly raised for loadable_variables when only drop_variables is supplied.

- [ ] Tests added
- [ ] Tests passing


TomNicholas (approved):
Good catch @chuckwondo !

Issue #377, opened 2025-01-13
Vendor kerchunk readers?


TomNicholas:
We still have dependency issues with kerchunk not supporting zarr-python v3, but icechunk requiring it (see https://github.com/zarr-developers/VirtualiZarr/issues/321). It unfortunately doesn't look like these will be resolved in kerchunk any time soon.

Right now we only actually use kerchunk code directly within VirtualiZarr for:
- the netCDF3 reader
- the HDF5 reader (which we now have Sean's alternative `HDF` reader for)
- the FITS reader (which is cool but fewer people use)

We really want to be able to pin `zarr-python>=3.0.0`, for so many reasons. But right now doing so would break those readers in VirtualiZarr because of the kerchunk incompatibility.

I realised yesterday that the most expedient thing to do here might be to just vendor (i.e. copy-paste) the code for the FITS and netCDF3 readers into VirtualiZarr.

The implementations for these are:
- the only part of kerchunk we need to keep right now to make our tests pass (as we don't currently have a working TIFF reader - see #291),
- they already work with zarr v3 (I think - the netCDF3 implementation doesn't even import zarr so that one really should work),
- they have barely been updated in over a year so we're not likely to miss out on regular new changes,
- they're an easily abstracted component that we can fix/replace later,
- They're not that big, ~300 lines each.

That would allow us to go full steam ahead with all the other things we need to do to be able to pin `zarr-python>=3.0.0` (e.g. #374, #182, #175).

FYI @bamford - I see you're the last person that committed to the FITS reader (https://github.com/fsspec/kerchunk/pull/531), so you will want to be aware of us potentially forking it!

maxrjones:
> We still have dependency issues with kerchunk not supporting zarr-python v3, but icechunk requiring it (see https://github.com/zarr-developers/VirtualiZarr/issues/321). It unfortunately doesn't look like these will be resolved in kerchunk any time soon.

@martindurant offered to work on kerchunk <-> zarr-python v3 compatibility over in https://github.com/fsspec/kerchunk/pull/516#issuecomment-2590042012. While I understand there may be times in which vendoring is necessary, I'm a -1 on this approach unless we first try to resolve any communication gaps about the urgency of kerchunk adopting zarr-python v3 for downstream packages.

[...]

PR #397, opened 2025-01-29
Vendor kerchunk netCDF3 reader

TomNicholas:
Passes locally with zarr v2. But I think this will break for an env with zarr v3 in it.

- [x] Closes part of #377
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/virtualizarr/vendor/kerchunk/utils.py
+++ b/virtualizarr/vendor/kerchunk/utils.py
@@ -1,0 +1,81 @@
+import base64
+
+import fsspec
+import ujson
+import zarr
+
+
+def _encode_for_JSON(store):
+    """Make store JSON encodable"""
+    for k, v in store.copy().items():
+        if isinstance(v, list):
+            continue
+        else:
+            try:
+                # minify JSON
+                v = ujson.dumps(ujson.loads(v))
+            except (ValueError, TypeError):
+                pass
+            try:
+                store[k] = v.decode() if isinstance(v, bytes) else v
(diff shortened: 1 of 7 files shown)


virtualizarr/vendor/kerchunk/utils.py
```diff
@@ -0,0 +1,81 @@
...
+                    data=thing[:],
+                    chunks=thing.shape,
+                    compression=None,
+                    overwrite=True,
+                    fill_value=thing.fill_value,
+                )
+                arr.attrs.update(original_attrs)
+
+
+def inline_array(store, threshold=1000, names=None, remote_options=None):
```
TomNicholas, line 52: We probably don't need this at all, because we never use kerchunk's inlining (we use `loadable_variables` instead).

TomNicholas:
Note that in the [kerchunk compatibility PR](https://github.com/fsspec/kerchunk/pull/516) the netCDF3 reader has now been updated to work with zarr-python v3. So we could just copy those changes here temporarily.

maxrjones:
> Note that in the [kerchunk compatibility PR](https://github.com/fsspec/kerchunk/pull/516) the netCDF3 reader has now been updated to work with zarr-python v3. So we could just copy those changes here temporarily.

I think you already suggested this elsewhere, but let's handle these updates in a separate PR

maxrjones (approved):
Looks good, I think we can ignore the code coverage gaps on the vendored readers for now

PR #399, opened 2025-01-29
Vendor kerchunk fits reader

TomNicholas:
Passes locally with zarr v2. But I think this will break for an env with zarr v3 in it.

- [x] Closes part of #377
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


TomNicholas:
Note that in the kerchunk compatibility PR the FITS reader has now been updated to work with zarr-python v3. So we could just copy those changes here temporarily.

maxrjones:
superceded by #406

PR #393, opened 2025-01-28
Adding `to_icechunk` as another type of roundtrip tests

jsignell:
- [ ] Closes #376 
- [x] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage
- [ ] Changes are documented in `docs/releases.rst`

I started working on updating the integration tests to use in-memory icechunk stores, but I can't quite get them to pass.

PR #405, opened 2025-01-30
Roundtrip tests kerchunk tests without requiring kerchunk

### LINKED ISSUE #404: Roundtrip dict and json kerchunk tests without depending on kerchunk library
It was discussed yesterday that a lot of the roundtrip test could run pretty much as-written if we just make a little helper that we can substitute in for `xr.open_dataset(..., engine="kerchunk")`. 

Making that substitution seems valuable because then a lot more of the test suite would run even in the absence of kerchunk.

This could be done as an alternative to or in addition to #376

jsignell:
It benefits this library to run as many tests as possible in as many environments as possible. So it is good to run tests that exercise reading/writing kerchunk-style references _even when kerchunk is not available_ in the testing environment.

That being said. This doesn't currently pass in the upstream env - maybe the xarray zarr 3 reader doesn't do well with fsspec?

- [x] Closes #404
- [x] Tests passing in tests env
- [ ] Tests passing in upstream env

### DIFF
--- a/virtualizarr/tests/__init__.py
+++ b/virtualizarr/tests/__init__.py
@@ -1,8 +1,10 @@
 import importlib
 import itertools
 
+import fsspec
 import numpy as np
 import pytest
+import xarray as xr
 from packaging.version import Version
 
 from virtualizarr.manifests import ChunkManifest, ManifestArray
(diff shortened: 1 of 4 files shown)


jsignell:
In #393 I added roundtripping to icechunk as just another option. I am pretty happy with how https://github.com/zarr-developers/VirtualiZarr/pull/393/commits/bd95d6048aa134c7590cec34a7662b88074ddb01 cleans these tests up and makes it easy to imagine adding more roundtrip tests.

jsignell:
Closing this. See note on issue

PR #378, opened 2025-01-15
Ab/upgrade icechunk

abarciauskas-bgse:
Just opening this as a draft PR to share the changes I implemented to get icechunk>=0.1.0a10 to work.

I believe we will need a more complete redesign and refactor of how we're handling codecs and zarr-python 3.0 integration across the codebase.

Issue #376, opened 2025-01-10
Use in-memory icechunk stores in roundtrip tests


TomNicholas:
If we used an icechunk store for all tests that involved round-tripping, we could avoid the use of kerchunk in those tests (in conjunction with #374). This would help us move towards pinning zarr-python v3.

jsignell:
I can take a crack at this one unless someone is already on it.

Issue #392, opened 2025-01-27
Planning explicit dependency on Zarr v3


sharkinsspatial:
We have a few interdependent issues which discuss the next steps for completely migrating to Zarr v3 but I thought it may be helpful to outline a rough plan of attack which consolidates all of these issues in one spot so everyone can contribute to the discussion on how to achieve this.  I'll list things sequentially, but several of these may need to be approached in a single PR due to dependency conflicts.

- [ ] https://github.com/zarr-developers/VirtualiZarr/pull/175 Replace our existing Zarray metadata representation with the new zarr-python [v3 class](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/metadata/v3.py).  This change will touch a significant portion of the codebase as our serialization/deserialization logic and readers all rely on the existing metadata structure.
- [ ] https://github.com/zarr-developers/VirtualiZarr/issues/376 Replace kerchunk in roundtrip tests with in-memory Icechunk stores.  This would likely be made simpler by first updating our Zarray representation to remove some of the v3 specific logic needed when [Icechunk support was introduced](https://github.com/zarr-developers/VirtualiZarr/pull/256/files#diff-a72af67ccc7b6159cecc8072b56d2f2f8633002e06cc4a79672cd8b2c46d4cc6).
- [x] https://github.com/zarr-developers/VirtualiZarr/issues/374 Switch our HDF5 roundtrip test to use the `HDFVirtualBackend`.  Along with the previous steps, this would remove the bulk of our dependence on kerchunk and free us to depend on Zarr v3 explicitly with only a limited loss of existing functionality.  There are several outstanding issues (and likely more) with the `HDFVirtualBackend` that were discovered and raised during the Pangeo hack day in December that will need to be addressed.  I'll link to a separate issue with the plan for tackling these.
- [x] https://github.com/zarr-developers/VirtualiZarr/issues/377 Copy remaining kerchunk readers to Virtualizarr.  To avoid losing support for netCDF3 and FITS we can temporarily copy and paste these kerchunk readers into Virtualizarr and use the `dataset->kerchunk reader->virtualizarr kerchunk reader` approach until dedicated Virtualizarr readers can be developed for these formats.

[...]

PR #406, opened 2025-01-30
Update to zarr 3 and main kerchunk

jsignell:
Pulls commits from #378, #405 and #393

UPDATE: Ideas for #376 removed in https://github.com/zarr-developers/VirtualiZarr/pull/406/commits/42414f06cd11e34f9a8730ca02430ef17b9f34f8

- [x] Tests added
- [x] Tests passing
- [ ] Full type hint coverage

The remaining failures on upstream can be broken into:

- codecs-related: 
  ```python-traceback
  FAILED virtualizarr/tests/test_codecs.py::TestCodecs::test_manifest_array_zarr_v2_normalized - AssertionError: assert (BytesCodec(e...shuffle': 1})) == (Delta(codec_...shuffle': 1}))
  FAILED virtualizarr/tests/test_writers/test_icechunk.py::test_write_loadable_variable - TypeError: Group.create_array() got an unexpected keyword argument 'exists_ok'
  FAILED virtualizarr/tests/test_writers/test_zarr.py::test_zarr_v3_metadata_conformance - AssertionError: assert (True and 1 > 1)
  ```
- probably setting up the in-memory icechunk store wrong - these are new tests that have never worked  - removed in https://github.com/zarr-developers/VirtualiZarr/pull/406/commits/42414f06cd11e34f9a8730ca02430ef17b9f34f8
  ```python-traceback
  FAILED virtualizarr/tests/test_integration.py::TestRoundtrip::test_roundtrip_no_concat[HDFVirtualBackend-roundtrip_as_in_memory_icechunk] - TypeError: Expected a BytesBytesCodec. Got <class 'numcodecs.zarr3.FixedScaleOffset'> instead.
  FAILED virtualizarr/tests/test_integration.py::TestRoundtrip::test_kerchunk_roundtrip_concat[False-time_vars0-HDFVirtualBackend-roundtrip_as_in_memory_icechunk] - TypeError: Expected a BytesBytesCodec. Got <class 'numcodecs.zarr3.FixedScaleOffset'> instead.
  FAILED virtualizarr/tests/test_integration.py::TestRoundtrip::test_kerchunk_roundtrip_concat[True-time_vars1-HDF5VirtualBackend-roundtrip_as_in_memory_icechunk] - TypeError: Group.create_array() got an unexpected keyword argument 'exists_ok'
  FAILED virtualizarr/tests/test_integration.py::TestRoundtrip::test_kerchunk_roundtrip_concat[True-time_vars1-HDFVirtualBackend-roundtrip_as_in_memory_icechunk] - TypeError: Group.create_array() got an unexpected keyword argument 'exists_ok'
  FAILED virtualizarr/tests/test_integration.py::TestRoundtrip::test_datetime64_dtype_fill_value[roundtrip_as_in_memory_icechunk] - KeyError: '<M8[ns]'
  ```


TomNicholas:
>probably setting up the in-memory icechunk store wrong - these are new tests that have never worked

I don't know that this PR can be disentangled from the changes we need to get icechunk working with zarr v3...

jsignell:
> I don't know that this PR can be disentangled from the changes we need to get icechunk working with zarr v3...

I can take the new in-memory icechunk stuff and leave the place where it could slot in. Or this can just be used as a reference point. I'm not attached 😄

TomNicholas:
So IIUC we can either:
- Merge this now, break `main` but only for `upstream`, then fix the remaining failures in follow-up PRs. That's kind of what https://github.com/zarr-developers/VirtualiZarr/issues/392#issuecomment-2620036608 implied by separating the "pin zarr" and "fix results of pinning zarr" into separate steps.
- Just try to fix the codec issues in this PR

maxrjones:
> So IIUC we can either:
> 
> * Merge this now, break `main` but only for `upstream`, then fix the remaining failures in follow-up PRs. That's kind of what [Planning explicit dependency on Zarr v3 #392 (comment)](https://github.com/zarr-developers/VirtualiZarr/issues/392#issuecomment-2620036608) implied by separating the "pin zarr" and "fix results of pinning zarr" into separate steps.
> * Just try to fix the codec issues in this PR

I'd slightly prefer if we merged this now because it seems to me like this PR really breaks the stalemate that we were in and would unblock other work. specifically for my work, I'd love to merge it into https://github.com/zarr-developers/VirtualiZarr/pull/407 to get the updates already implemented

pyproject.toml
```diff
@@ -39,7 +39,7 @@ hdf_reader = [
     "numcodecs"
 ]
 icechunk = [
-    "icechunk==0.1.0a8",
+    "icechunk>=0.1.0a12",
```
maxrjones, line 42: @jsignell what's the motivation for using a12 rather than a15?

jsignell: I saw this comment: https://github.com/zarr-developers/VirtualiZarr/pull/378#issuecomment-2608168675 and so I thought a12 would be ok. And I looked at the codebase for the latest tagged version of icechunk and I saw [this line](https://github.com/earth-mover/icechunk/blob/a13459b118e2d7eb0774f398064bdab34997b5c2/icechunk-python/pyproject.toml#L23)  which made me nervous about using newer than a12

maxrjones: nice, thanks for sharing!

TomNicholas:
> I'd slightly prefer if we merged this now

I'm happy to merge this now if @jsignell is. We want to block everyone else as little as possible.

virtualizarr/tests/test_writers/test_zarr.py
```diff
@@ -42,7 +42,7 @@ def test_zarr_v3_metadata_conformance(tmpdir, vds_with_manifest_arrays: Dataset)
     assert isinstance(metadata["fill_value"], (bool, int, float, str, list))
     assert (
         isinstance(metadata["codecs"], list)
-        and len(metadata["codecs"]) > 1
+        and len(metadata["codecs"]) == 1
```
jsignell, line 45: @abarciauskas-bgse did the expectation change so it is correct that there is one codec now?

virtualizarr/zarr.py
```diff
@@ -198,6 +191,23 @@ def _v3_codec_pipeline(self) -> Any:
 
         return codec_pipeline
 
+    def serializer(self) -> Any:
```
jsignell, line 194: This function could stand to get reviewed and tested. It looks like it was meant to be just a sketch of what a serializer could look like

virtualizarr/tests/test_writers/test_zarr.py
abarciauskas-bgse: Yes, this is because of how `serializer` was broken out of `_v3_codec_pipeline` so these are 2 separate objects now. Today I will investigate if this is how it should be, but my guess is that this will become a part of the refactor to use the zarr-python array class instead of Virtualizarr's zarr.py

jsignell: yeah I took a look at #175 and I agree that this is all going away. So it is probably not worth doing too much work around it.

jsignell:
I just double checked #410 by merging it with this branch and the tests worked! (Although I did find [76dcfb7](https://github.com/zarr-developers/VirtualiZarr/pull/406/commits/76dcfb7ee9ba9df7f132734241696546a0c59f73) 🙊 )

virtualizarr/zarr.py
TomNicholas: If this passes the tests we can perhaps merge and deal with this in the follow-up PR that refactors `ZArray` away entirely?

jsignell: Yeah exactly.

TomNicholas (approved):

jsignell:
I wasn't sure whether or not to add release notes, but feel free to push to this branch or just merge as you see fit.

TomNicholas:
We'll just add release notes in the follow-up that actually changes the kerchunk dependency to a released version

PR #407, opened 2025-01-30
Manage dependencies with pixi

### LINKED ISSUE #310: Use uv / pixi?

- If you need Conda-Forge in addition to PyPI distributions, ~~also~~ use [pixi](https://pixi.sh/latest/)

_Originally posted by @maxrjones in https://github.com/zarr-developers/VirtualiZarr/issues/309#issuecomment-2486738368_

maxrjones:
Moves from using conda environment files to using pixi to store all dependency information in pyproject.toml

Check out the updated [contributing guide](https://virtualizarr--407.org.readthedocs.build/en/407/contributing.html#contributing-code) to get started.

TODO:
- [x] Update readthedocs build setup
- [x] Add quickstart to PR description
- [x] Set up environments for minimum and upstream tests
- [x] Update contributor guide
- [x] Check that release workflow doesn't need updating
- [x] Add release notes

Closes #310


TomNicholas:
At the risk of being really annoying, should we be switching to pixi or to uv here?

maxrjones:
> At the risk of being really annoying, should we be switching to pixi or to uv here?

It doesn't matter to me as long as we find a way to install all dependencies without using conda environment files. Pip installing h5py doesn't install the netcdf4/hdf5 C libraries which will give the user a module not found error when trying to use those readers. So, in this PR I use the conda distributions of those libraries in https://github.com/zarr-developers/VirtualiZarr/blob/a2b546160e11082b8cd65d4bfe1a813abc37eaf8/pyproject.toml#L121-L124 Do you know if those wheels are available via PyPI such that we could meet this goal using uv without pixi?

maxrjones:
> There seem to be some wheels listed to download for all of these projects:

good to know, thanks. since this PR will really change the developer experience I'll just open RFC PRs for both pixi and uv and see which people prefer. Probably won't get to it until over the weekend though

TomNicholas:
Thank you @maxrjones ! Really appreciate your efforts here.

maxrjones:
FYI the one failure in this PR is due to an unreleased change in numcodecs that adds a `typesize` to the Blosc properties (https://github.com/zarr-developers/numcodecs/blob/3c933cf19d4d84f2efc5f3a36926d8c569514a90/docs/release.rst#improvements). This PR includes additional upstream checks which is why it's not caught in main. I'd prefer to either mark as xfail or leave it failing to modify the hdf Blosc configuration separately.

.gitignore
```diff
@@ -180,3 +180,11 @@ examples/
...
 
 # End of https://www.toptal.com/developers/gitignore/api/visualstudiocode
+
+# Pixi folder
+.pixi/
+
+# python virtual environment
+.venv
+venv
+pixi.lock
```
cisaacstern, line 190: @maxrjones my understanding is that lockfiles are intended to be committed into version control, for reproducibility. This is how I've used pixi in my own projects. Is excluding the pixi lock file a deliberate choice here to go against that pattern? If so I'm curious what the thought process is.

maxrjones: Mostly I just thought it'd be annoying for the versions in the upstream environment to be constantly changing. I'm familiar with committing lockfiles for research projects or ephemeral packages but not so much for software libraries, so it's interesting to know that's also a pattern.

cisaacstern: It's a good question, and in truth in my initial question I had not fully considered the distinction between how a library vs. a more ephemeral or specific application may want to handle this. Certainly, the presence of the lockfile in version control (and it's degree of churn, as you point out) could be a point of friction to those that are not familiar with it, and perhaps that is a good reason to not commit it, as you suggest.

There is possibly some risk, in not committing, of un-checked-in lockfiles in developer's local dev environments resulting in incorrect environment state, especially as regards the upstream versions (if they are attempting to test those environments locally). What I am imagining is a scenario where the upstream version has changed on the remote source (i.e. latest commit on `xarray` on github has changed), but the developer's lockfile (and associated local `.pixi/envs`) are locked against a different version. IIUC, in this scenario pixi will potentially _not_ re-pull the latest commit from github, because according to the developer's local lockfile, the environment requirements are already met (we do, in fact, have `xarray` in the environment).

I could be a little off-base here, but there is sufficient possibility of confusion here that I wanted to mention it. In the alternative where we _do_ commit the lockfile, I believe the lockfile pulled from the `VirtualiZarr` remote ref will (might?) include the specific commit to be checked out from the remote for the upstream envs.

Does that make any sense? Happy to sync briefly sometime if it does not. Generally speaking, this concern arises from my own use of `pixi.lock` to manage rapidly iterating development environments, in which I have sometimes found pixi can assume that the environment is satisfied by the lockfile, when in fact what I _want_ it do is re-pull something from an updated/latest ref.

cisaacstern: Re: library vs. ephemeral project distinction, one anecdotal data point I just found, is that pydantic commits `uv.lock` https://github.com/pydantic/pydantic/blob/main/uv.lock

maxrjones: I committed the lockfile in [`5bbbfd3` (#407)](https://github.com/zarr-developers/VirtualiZarr/pull/407/commits/5bbbfd34133724676d62913071d63497ab75b9a2). I'm curious how this would work for people on other operating systems (I use a mac) since there seem to be some wheels built for mac in the lockfile.

pyproject.toml
```diff
@@ -75,16 +76,37 @@ icechunk = [
...
 
-test = [
+# Dependency sets under dependencies-groups are NOT available via PyPI
+[dependency-groups]
+upstream = [
+    'xarray @ git+https://github.com/pydata/xarray',
+    'universal_pathlib @ git+https://github.com/fsspec/universal_pathlib',
+    'numcodecs @ git+https://github.com/zarr-developers/numcodecs',
+    'ujson @ git+https://github.com/ultrajson/ultrajson',
+    'zarr @ git+https://github.com/zarr-developers/zarr-python',
```
abarciauskas-bgse, line 92: Should zarr-python from github be an upstream dependency? I know it's also in default dependencies but just curious because zarr-python also a dependency of icechunk.

abarciauskas-bgse (commented):
LGTM! I'll let @TomNicholas provide official approval.

pyproject.toml
maxrjones: Since we're using Zarr's internal metadata classes in the `develop` branch, I think it's important to explicitly test for compatibility with the latest, unreleased changes in Zarr-Python.

.gitignore
maxrjones: I went back to my original decision to ignore it in [`382148e` (#407)](https://github.com/zarr-developers/VirtualiZarr/pull/407/commits/382148eaf61f59bd01d3fe524cb95214c67b399a) because the lock file changed with multiple upstream commits even in the short time in which I was responding to Aimee's comments. I want this to be as minimally disruptive as possible for the other developers and think ignoring it is the safer choice for that reason. We can revisit if it causes issues.

ci/min-deps.yml
TomNicholas, line 1: What defines the `min-deps` dependencies now? The dependencies in the `pyproject.toml`?

docs/contributing.md
```diff
@@ -6,51 +6,53 @@ Contributions are welcome and encouraged! We ask only that all contributors foll
...
-mamba env create -f ci/environment.yml
-mamba activate virtualizarr-tests
-pre-commit install
-# git checkout -b new-feature
-python -m pip install -e . --no-deps
-python -m pytest
+# Run all tests
+pixi run --environment test run-tests
+# Skip tests that require a network connection
+pixi run --environment test run-tests-no-network
```
TomNicholas, line 21: Why is it necessary to make running the tests a `pixi` thing, rather than just installing dependencies using pixi and then running tests using `pytest` directly?

docs/contributing.md
```diff
@@ -6,51 +6,53 @@ Contributions are welcome and encouraged! We ask only that all contributors foll
...
-mamba activate virtualizarr-docs
-python -m pip install -e .  # From project's root - needed to generate API docs
-cd docs # From project's root
-rm -rf generated
-make clean
-make html
+pixi install --environment docs
+pixi run build-docs
+```
+Pixi can also be used to serve continuously updating version of the documentation during development at [http://0.0.0.0:8000/](http://0.0.0.0:8000/).
```
TomNicholas, line 51: Oh that's cool. Will that work with #486?

TomNicholas (commented):
This looks great, I just have a couple of questions so I can understand the choices better.

docs/contributing.md
maxrjones: yeah, unfortunately they'll have some merge conflicts to resolve but it uses the same `sphinx-autobuild` extension they reference.

ci/min-deps.yml
maxrjones: yeah it's in `pyproject.toml` and defined is the combination of these dependency sets:
https://github.com/zarr-developers/VirtualiZarr/blob/efcb153973b82597460910eae44a837a759d09c8/pyproject.toml#L171

It's unfortunate the s3 (included as part of `dev` and the hdf libraries are required right now for running any tests (this was already the case) and would be good to refactor the `conftest.py` modules so this isn't the case.

I also added a line to the contributing guide showing how to list the installed environments.

TomNicholas (approved):
Amazing @maxrjones thank you so much!

Issue #311, opened 2024-11-20
Support appending to icechunk store


abarciauskas-bgse:
As a user of icechunk's virtual references I want to be able to append to an existing icechunk store.

I made this as distinct from https://github.com/zarr-developers/VirtualiZarr/issues/21 as that issue is for appending to kerchunk-backed reference stores.

TomNicholas:
Just to note that the reason this is still open is that we don't have any mention of it in the usage docs.

abarciauskas-bgse:
I will go ahead and add an example to https://virtualizarr.readthedocs.io/en/latest/usage.html#writing-to-an-icechunk-store

abarciauskas-bgse:
Closing via https://github.com/zarr-developers/VirtualiZarr/pull/272 and #409

PR #409, opened 2025-01-30
Update usage.md with append method for icechunk

abarciauskas-bgse:
- [x] Closes https://github.com/zarr-developers/VirtualiZarr/issues/311

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -431,14 +431,27 @@ We can also write these references out as an [IcechunkStore](https://icechunk.io
 ```python
 # create an icechunk repository, session and write the virtual dataset to the session
 from icechunk import Repository, Storage, VirtualChunkContainer, local_filesystem_storage
-storage = local_filesystem_storage(str('combined'))
+storage = local_filesystem_storage("./local/icechunk/store")
 
 # By default, local virtual references and public remote virtual references can be read wihtout extra configuration.
 repo = Repository.create(storage=storage)
 session = repo.writeable_session("main")
 
 # write the virtual dataset to the session with the IcechunkStore
-combined_vds.virtualize.to_icechunk(session.store)
+vds1.virtualize.to_icechunk(session.store)
+session.commit("Wrote first dataset")
+```
+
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -431,14 +431,27 @@ We can also write these references out as an [IcechunkStore](https://icechunk.io
 ```python
 # create an icechunk repository, session and write the virtual dataset to the session
 from icechunk import Repository, Storage, VirtualChunkContainer, local_filesystem_storage
-storage = local_filesystem_storage(str('combined'))
```
abarciauskas-bgse, line 434: I wasn't really sure why the argument was `str('combined')` but happy to revert this change if there is a reason

TomNicholas: I mean this can't possibly be necessary, because `str(<str obj>) == <str obj>`, even down to having the same hash. So you should be able to get rid of the enclosing `str()` call.

docs/usage.md
```diff
@@ -431,14 +431,27 @@ We can also write these references out as an [IcechunkStore](https://icechunk.io
...
 
 # write the virtual dataset to the session with the IcechunkStore
-combined_vds.virtualize.to_icechunk(session.store)
+vds1.virtualize.to_icechunk(session.store)
+session.commit("Wrote first dataset")
+```
+
+#### Append to an existing Icechunk Store
+
+You can append a virtual dataset to an existing Icechunk store using the `append_dim` argument. This is especially useful for ongoing datasets. Note that Zarr does not currently support concatenating datasets with different codecs or chunk shapes.
```
TomNicholas, line 447: I would change "ongoing" to "datasets that expand over time" or something like that.

TomNicholas (changes requested):
Thanks @abarciauskas-bgse - these two minor suggestions and then please merge!

docs/usage.md
abarciauskas-bgse: oh yes, I did already get rid of the enclosing `str` call (see change below), but I also changed the string, which I think more clearly indicates a local path

PR #417, opened 2025-02-03
Fix tests by recreating the repo every time

### LINKED ISSUE #415: Fix append failures in upstream tests
@maxrjones and @sharkinsspatial just noticed the append tests are failing in upstream, so going to take a look at fixing them.

abarciauskas-bgse:
I believe the issue here was that the repo was not getting torn down after each test, and using a fixture with the `@pytest.fixture(scope="function")` for the repo seems to fix the issue.

- [x] Closes #415 
- [x] Tests passing

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -12,15 +12,42 @@ and on conda-forge:
 conda install -c conda-forge virtualizarr
 ```
 
+## Optional dependencies
+
+VirtualiZarr has many optional dependencies, split into those for reading various file formats, and those for writing virtual references out to different formats.
+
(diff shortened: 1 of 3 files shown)


TomNicholas (approved):
Excellent!!

@mpiannucci you can tell us if there was a better way of doing this later :)

For now let's merge because this makes everything green!!!

TomNicholas:
Actually just before merging @abarciauskas-bgse can you add/change release notes to note that we actually support icechunk a12 now not a8? Then please merge

Issue #413, opened 2025-02-03
Release 1.3.0 or 2.0.0?


maxrjones:
Should we release VirtualiZarr now so that people can use it with Zarr-Python 3.0, or wait until after phase 2 (Migrate to zarr_format=3) of the ongoing [Zarr-Python 3.0 migration project](https://github.com/orgs/zarr-developers/projects/6?query=sort%3Aupdated-desc+is%3Aopen)? If we wait, would the release need to be v2.0.0 rather than 1.3.0 due to the breaking changes since the last release (https://virtualizarr.readthedocs.io/en/latest/releases.html#v1-2-1-unreleased) and the major changes to ManifestArray associated iwth https://github.com/zarr-developers/VirtualiZarr/issues/411?

TomNicholas:
Good question, but I think before we release anything we should fix compatibility with the latest version of Icechunk.

> breaking changes since the last release

I think these constitute minor changes, so would be v1.3.0.

> major changes to ManifestArray

Including these would definitely justify bumping the major version to v2.0.0.

maxrjones:
Thanks, I forgot about the upstream issue. I think it's still good to have this open now so that non-devs have awareness of our release schedule.

TomNicholas:
We now are about to have everything green! Also icechunk has now released `0.1.0`, so we're out of the `alpha` series. I suggest we:

- Merge #417 
- Update Icechunk dependency to `0.1.0` and fix anything that crops up
- Release virtualizarr `1.3.0`
- Then work on the rest of the zarr-python v3 stuff in preparation for releasing virtualizarr `2.0.0`.

TomNicholas:
Actually icechunk just released `0.1.1`

TomNicholas:
I just released `v1.3.0`, so will close this.

PR #419, opened 2025-02-04
Update Icechunk dependency to v0.1.1

TomNicholas:
- [x] Closes part of https://github.com/zarr-developers/VirtualiZarr/issues/413#issuecomment-2632338745
- [ ] ~~Tests added~~
- [ ] Tests passing

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -15,7 +15,6 @@ dependencies:
   - universal_pathlib
   - hdf5plugin
   - numcodecs
-  - icechunk>=0.1.1
   # Testing
   - codecov[toml]
   - pre-commit
@@ -38,3 +37,4 @@ dependencies:
   - pip
   - pip:
     - imagecodecs-numcodecs==2024.6.1
+    - icechunk>=0.1.1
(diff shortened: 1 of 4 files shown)


ci/upstream.yml
```diff
@@ -29,6 +29,6 @@ dependencies:
   - fsspec
   - pip
   - pip:
-      - icechunk>=0.1.0a12 # Installs zarr-python v3.0.0 as dependency
+      - git+https://github.com/earthmover/icechunk.git@main # Installs zarr-python v3.0.0 as dependency
```
TomNicholas, line 32: Note this PR now uses the `upstream` CI env to actually test against the `main` branch of icechunk

ci/environment.yml
```diff
@@ -15,6 +15,7 @@ dependencies:
   - universal_pathlib
   - hdf5plugin
   - numcodecs
+  - icechunk>=0.1.1
```
TomNicholas, line 18: Note our primary CI env also now tests against (the last released version of) icechunk, because we can do that now that icechunk and kerchunk are no longer mutually exclusive.

mpiannucci: It's not on conda just yet

TomNicholas: Oh right! Will `v0.1.1` be on conda? That would be nice.

Issue #238, opened 2024-09-13
Add hypergrib as as a grib reader


TomNicholas:
I just learned about [hypergrib](https://github.com/JackKelly/hypergrib) (via [this blog post](https://openclimatefix.org/post/lazy-loading-making-it-easier-to-access-vast-datasets-of-weather-satellite-data?fbclid=IwZXh0bgNhZW0CMTEAAR3d-3cSqpjSQhfXIMpxmxf1ehO_LxiiwCuzc4ugcgOH06Gird5I3dhP5LI_aem_lIsfwLf64bS9oKuu9G7Jow)).

I don't know much about how GRIB files are structured, but if @jackkelly makes a very performant GRIB reader that can:
1. be called from python,
2. return `ChunkManifest` objects and associated metadata (ideally using [`ChunkManifest.from_arrays()`](https://github.com/zarr-developers/VirtualiZarr/blob/53a609f75d4fee3b6b111bdfcf0df01f824480db/virtualizarr/manifests/manifest.py#L142) to avoid any intermediate representation),

then we could add it as a new reader to [`virtualizarr.readers`](https://github.com/zarr-developers/VirtualiZarr/tree/main/virtualizarr/readers).

In fact FYI my intention now is to probably make an entrypoint system for virtualizarr, in addition to plugging in to xarray's backend entrypoint system. That would allow this syntax for creating virtual `xr.Dataset` objects containing `ManifestArrays`:

```python
vds = xr.open_dataset(grib_paths, engine='virtualizarr', reader='hypergrib')
```

JackKelly:
Sounds awesome! My intention is definitely for `hypergrib` to play nicely with the existing ecosystem, including `xarray` and `virtualizarr`.

But `hypergrib` won't do anything vaguely useful for at least a few months. When I'm a bit further down the line, I'll get back in touch, so we can discuss the optimal way to plug `hypergrib` into `virtualizarr` :slightly_smiling_face:

JackKelly:
Just to flag that, after @emfdavid pointed out that some GRIB datasets contain _billions_ of GRIB messages (chunks), I've changed the design for `hypergrib` so that it no longer stores a manifest of chunks. After doing some more research, it seems likely that we'll soon see NWP GRIB datasets with _trillions_ of chunks.

A kerchunk JSON manifest of a trillion chunks would be about 50 TB!

[...]

Issue #424, opened 2025-02-04
ManifestArray should use zarr-python's ArrayV3Metadata


abarciauskas-bgse:
En route to replacing `ZArray` custom class with zarr-python's `ArrayV3Metadata`, we first need to support ManifestArrays using `ArrayV3Metadata`.

In order to achieve #411 in an incremental way, my proposal is to first support `ArrayV3Metadata` as well as `ZArray` in ManifestArray's, so that we can then incrementally refactor the readers and writers to implement or use ManifestArray's using `ArrayV3Metadata` while still having passing tests. Once all readers and writers using ManifestArrays are refactored, we can remove the custom `ZArray` class.

As suggested in #411, this will include renaming the ManifestArray `zarray` property to `metadata`.

See also #423 for a reference on how codec pipelines may be handled.

Thoughts welcome, cc @mpiannucci @TomNicholas

TomNicholas:
Splitting out refactoring `ManifestArray` and refactoring the readers is a good idea.

> my proposal is to first support `ArrayV3Metadata` as well as `ZArray` in `ManifestArray`'s,

However I don't think you should try to support both at once inside `ManifestArray` - that's going to be really hard and messy. 

Instead, you should change `._zarray: ZArray` -> `._metadata: ArrayV3Metadata` in one go, but avoid having to refactor the readers just by creating some short-lived function 
```python
def zarray_to_v3metadata(zarray: ZArray) -> ArrayV3Metadata:
    ...
```
and calling that in the readers just before they create a `ManifestArray`.

> while still having passing tests

If you wanted to avoid changing tests (that would now have to check `.metadata` instead of checking `.zarray`), you could simply do the reverse, using a function `v3metadata_to_zarray`.

abarciauskas-bgse:
That makes sense! Thanks @TomNicholas

abarciauskas-bgse:
FYI I'm working on this in [manifest-arrays-use-arrayv3metadata](https://github.com/zarr-developers/VirtualiZarr/tree/manifest-arrays-use-arrayv3metadata), not quite ready for a draft PR as I'm still working through the implementation and tests. Hoping to finish tomorrow 🤞🏽

ayushnag:
[...]

PR #427, opened 2025-02-05
Manifest/Functional Store

### LINKED ISSUE #124: Loading data from ManifestArrays without saving references to disk first
I am working on a feature in `virtualizarr` [to read dmrpp metadata files and create a virtual `xr.Dataset`](https://github.com/TomNicholas/VirtualiZarr/issues/85) containing manifest array's that can then be virtualized. This is the current workflow: 
```python
vdatasets = parser.parse(dmrs)
# vdatasets are xr.Datasets containing ManifestArray's
mds = xr.combine_nested(list(vdatasets), **xr_combine_kwargs)
mds.virtualize.to_kerchunk(filepath=outfile, format=outformat)
ds = xr.open_dataset(outfile, engine="virtualizarr", ...)
ds.time.values
```
However the chunk manifest, encoding, attrs, etc. is already in `mds` so is it possible to read data directly from this dataset? My understanding is that once the "chunk manifest" ZEP is approved and the `zarr-python` reader in `xarray` is updated this should be possible. The `xarray` reader for `kerchunk` can accept a file or the reference json object directly from `kerchunk` `SingleHdf5ToZarr ` and `MultiZarrToZarr`. So similarly can we extract the refs from `mds` and pass it to `xr.open_dataset()` directly?

There probably still needs to be a function that extracts the refs so that xarray can make a new `Dataset` object with all the indexes, cf_time handling, and `open_dataset` checks.

```python
mds = xr.combine_nested(list(vdatasets), **xr_combine_kwargs)
refs = mds.virtualize()
ds = xr.open_dataset(refs, engine="virtualizarr", ...)
```

Even reading directly from the ManifestArray dataset is possible but not sure how the new dataset object with numpy arrays and indexes would be separate from the original dataset
```python
mds = xr.combine_nested(list(vdatasets), **xr_combine_kwargs)
mds.time.values
```

ayushnag:
- [X] Closes #124
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

Also discussed in depth here: https://github.com/zarr-developers/VirtualiZarr/issues/238#issuecomment-2585549691


TomNicholas:
Sweet! Thanks @ayushnag.

Note this requires #424 because of https://github.com/zarr-developers/VirtualiZarr/issues/424#issuecomment-2637984652.

virtualizarr/manifests/manifest.py
```diff
@@ -54,6 +65,110 @@ def with_validation(
         return ChunkEntry(path=path, offset=offset, length=length)
 
 
+class ManifestStore(Store):
```
TomNicholas, line 68: I think this class should be defined in it's own file, and potentially even made public (developer) API.

virtualizarr/manifests/manifest.py
```diff
@@ -54,6 +65,110 @@ def with_validation(
...
+    supports_partial_writes: bool = False
+    supports_listing: bool = True
+
+    fs: AsyncFileSystem
+    vds: xr.Dataset
+
+    def __init__(self, fs, vds, read_only=True):
+        super().__init__(read_only=read_only)
+        self.fs = fs
+        self.vds = vds
```
TomNicholas, line 80: What happens if some of the variables in the dataset are not virtual? Do we have any need to support that case?

virtualizarr/manifests/manifest.py
```diff
@@ -54,6 +65,110 @@ def with_validation(
...
+
+    async def clear(self) -> None:
+        self.fs = None
+        self.vds = None
+
+    def __str__(self) -> str:
+        return f"manifest://{id(self.vds)}"
```
TomNicholas, line 87: We might want to build a custom repr using the `xarray.Dataset` repr here. Low priority suggestion though.

virtualizarr/manifests/manifest.py
```diff
@@ -54,6 +65,110 @@ def with_validation(
...
 
 
+class ManifestStore(Store):
+    supports_writes: bool = False
+    supports_deletes: bool = False
+    supports_partial_writes: bool = False
+    supports_listing: bool = True
+
+    fs: AsyncFileSystem
+    vds: xr.Dataset
```
TomNicholas, line 75: Technically this could also be a `DataTree`, or maybe a dict of `Datasets`. But we might never need that.

virtualizarr/manifests/manifest.py
```diff
@@ -54,6 +65,110 @@ def with_validation(
...
+        prototype: BufferPrototype,
+        byte_range: None = None,  # could this optionally accept a RangeByteRequest?
+    ) -> Buffer | None:
+        if not self._is_open:
+            await self._open()
+        print("key: ", key)
+        print("key split: ", key.split("/"))
+        array_name, _, chunk_key = key.split("/")
+        # TODO: is this the best way?
+        url, offset, length = self.vds[array_name].data.manifest.dict()[chunk_key]
```
TomNicholas, line 111: I think this is a reason to improve the API of `ChunkManifest` to make it more dict-like (separately from this PR). Although I guess we already have the entire numpy array containing all references already in memory.

virtualizarr/manifests/manifest.py
```diff
@@ -54,6 +65,110 @@ def with_validation(
...
+            paths.append(url)
+            starts.append(offset)
+            stops.append(offset + length)
+        res = await self.fs._cat_ranges(paths, starts, stops, on_error="return")
+        return [prototype.buffer.from_bytes(r) for r in res]
+
+    async def exists(self, key: str) -> bool:
+        array_name, _, chunk_key = key.split("/")
+        url, _, _ = self.vds[array_name].data.manifest.dict()[chunk_key]
+        return await self.fs._exists(url)
```
TomNicholas, line 145: I think this should instead look at the `._paths` array in the manifest and return `True` if the path string is non-empty.

ayushnag: But I think we need to check the path for that specific chunk. The `_paths` array cannot retrieve the specific path for that key

TomNicholas: What I mean is that there is a convention in the `ChunkManifest` class that a zarr store with missing chunks has those chunks represented using an empty path string in the manifest. (The numpy array implementation prevented me from just omitting those keys entirely.)

It's the readers' job to find out if a chunk doesn't actually exist, so by the time we get here we should already know which chunks exist in storage and which don't without having to reach out to the storage to find out.

TomNicholas:
FYI you could rebase this onto @abarciauskas-bgse 's new `zarr-python-3.0` branch if you want to push it forward by building atop #424.

Issue #414, opened 2025-02-03
Updating fill_value logic for HDFVirtualBackend


sharkinsspatial:
I'm working through trying to update the "fill value" (quoted to denote the massive ambiguity around this loaded term as described by @rabernat in https://github.com/pydata/xarray/issues/5475#issuecomment-2381477207) for the `HDFVirtualBackend` but it seems this will be a multi-level issue.  My hope is our logic can follow the interpretation that @rabernat set out in https://github.com/pydata/xarray/issues/5475#issuecomment-2383046633 so that the reader will use the HDF dataset `fillvalue` property for the the Zarr `fill_value` and if the HDF dataset has a `_FillValue` `attr` that will be preserved as an array `attr` for use by xarray's decoding logic to represent masked values.  This is a transition from the the old `kerchunk` logic that was discussed and updated in https://github.com/fsspec/kerchunk/issues/177.

This presents a bit of a chicken and egg problem for our transition to Zarr V3 https://github.com/zarr-developers/VirtualiZarr/issues/392 and specifically our transition to `ArrayV3Metadata` https://github.com/zarr-developers/VirtualiZarr/issues/411 as an internal representation.  As outlined by @rabernat, Zarr v2 uses different semantics for `fill_value` and the spec is less rigorous about its [encoding](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html#fill-value-encoding).  The Zarr V3 specification has adopted @rabernat's new semantics and has more detailed guidelines for [fill_value encoding](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#fill-value).

IIUC as we transition the `HDFVirtualBackend` to use the new Zarr v3 `fill_value` semantics and encoding our `kerchunk` based roundtrip tests will be broken unless we add specific logic to our `to_kerchunk` serialization that transform v3 ->  v2 semantics.  I have not searched the new kerchunk v3 compatibility PR https://github.com/fsspec/kerchunk/pull/516#event-16126071360 exhaustively but it appears that the semantic fill_value logic used by the HDF reader has not changed and is still using the v2 logic so even serializing to v3 compatible kerchunk dictionaries would fail equivalence assertions.

I'll be working through a PR to update the HDFVirtualBackend semantics in our current codebase and I'll assess the extent of test failures and how they can be addressed.

[...]

PR #420, opened 2025-02-04
Align HDF reader CF _FillValue with Zarr v3 semantics

sharkinsspatial:
- [x] Closes #356
- [x] Tests added
- [x] Tests passing

PR #429, opened 2025-02-06
Manifest arrays use arrayv3metadata

### LINKED ISSUE #424: ManifestArray should use zarr-python's ArrayV3Metadata

abarciauskas-bgse:
This is still very much a WIP - many tests and implementations still need to be fixed.

A few notes:
- It was suggested we remove ZArray completely as a part of this work, as opposed to using a conversion function for ZArrays to ArrayV3Metadata. So we should be able to remove ZArray as a part of this pr.
- It was suggested _not_ to use zarr's `_parse_chunk_encoding_v3` function since it is a private function and may change, which is why some of that logic is replicated in `convert_to_codec_pipeline`

## Checklist

- [x] Closes #424 
- [x] Manifest tests passing
- [x] Library (codecs, etc) tests passing
- [x] Reader tests passing
- [x] test_integration tests passing
- [x] test_xarray tests passing
- [x] Writer tests passing
- [x] Cleanup dead code
- [x] Consider reorganizing codecs and zarr modules
- [x] Full type hint coverage
- [x] Tests added for new functions
- ~New functions/methods are listed in `api.rst`~
- ~New functionality has documentation~


virtualizarr/manifests/array.py
```diff
@@ -55,32 +55,43 @@ def __init__(
...
+    def metadata(self) -> ArrayV3Metadata:
+        return self._metadata
 
     @property
     def chunks(self) -> tuple[int, ...]:
-        return tuple(self.zarray.chunks)
+        """
+        Individual chunk size by number of elements.
+        """
+        if isinstance(self._metadata.chunk_grid, RegularChunkGrid):
```
TomNicholas, line 77: I think this should be checked at construction time, because the concatenation methods assume a regular chunk grid too.

virtualizarr/manifests/array.py
```diff
@@ -55,32 +55,43 @@ def __init__(
...
+        dtype_str = self.metadata.data_type
+        return dtype_str.to_numpy()
 
     @property
     def shape(self) -> tuple[int, ...]:
-        return tuple(int(length) for length in list(self.zarray.shape))
+        """
+        Array shape by number of elements along each dimension.
+        """
+        return tuple(int(length) for length in list(self.metadata.shape))
```
TomNicholas, line 94: Can you not just return `self.metadata.shape`?

virtualizarr/manifests/array_api.py
```diff
@@ -100,12 +102,12 @@ def concatenate(
         lengths=concatenated_lengths,
     )
 
-    # chunk shape has not changed, there are just now more chunks along the concatenation axis
-    new_zarray = first_arr.zarray.replace(
-        shape=tuple(new_shape),
-    )
+    metadata_copy = first_arr.metadata.to_dict().copy()
```
TomNicholas, line 105: This seems overly verbose, and something we will want in multiple places. Let's make a `replace` constructor function, before adding a method upstream.

virtualizarr/writers/kerchunk.py
```diff
@@ -87,8 +87,7 @@ def variable_to_kerchunk_arr_refs(var: Variable, var_name: str) -> KerchunkArrRe
             for chunk_key, entry in marr.manifest.dict().items()
         }
 
-        zarray = marr.zarray.replace(zarr_format=2)
-
+        array_v2_metadata = convert_v3_to_v2_metadata(marr.metadata)
```
abarciauskas-bgse, line 90: I don't think we want to go with this approach, perhaps we should allow for manifest arrays to have ArrayV2Metadata for writing to kerchunk.

TomNicholas: >  don't think we want to go with this approach

Why not? 

> perhaps we should allow for manifest arrays to have `ArrayV2Metadata` for writing to kerchunk.

I really don't see how that would make anything easier. Inside `variable_to_kerchunk_arr_refs` is the only place you should need to deal with v2 metadata, so why try to generalize it in the rest of the package if you know you won't need it anywhere else?

abarciauskas-bgse: that makes sense to me, I think I was just concerned about bugs trying to go between ArrayV3Metadata and ArrayV2Metadata. But having a single interface for metadata in memory is probably safer over all.

abarciauskas-bgse: Oh, I recalled, I am not sure yet about the best way to go from ArrayV3Codecs to V2 compressors and filters. I don't see anything yet in zarr-python for how to do that. I welcome any suggestions.

abarciauskas-bgse: Looking closer at https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/array.py#L4290-L4334 it looks like we can parse the codecs in the ArrayV3Metadata:

```python
filters, serializer, codecs = array_v3_metadata.codecs
```

virtualizarr/zarr.py
```diff
@@ -239,3 +212,139 @@ def _num_codec_config_to_configurable(num_codec: dict) -> dict:
...
+        dtype=dtype,
+        serializer=serializer,
+    )
+    return cast(tuple[ZarrCodec, ...], (*codecs[0], codecs[1], *codecs[2]))
+
+
+from zarr.abc.codec import ArrayArrayCodec, ArrayBytesCodec, BytesBytesCodec
+
+
+def identify_codec(codec):
```
abarciauskas-bgse, line 279: @d-v-b we have to extract compressors and filters from ArrayV3Metadata for use in writing V2 metadata to kerchunk and also for use in the icechunk writer where we call zarr-python's `group.require_array` - is this method the best way to do so?

d-v-b: yes, that function reflects the mapping between words and class names we use in zarr-python. But this conceptualization is specific to zarr v3: in zarr v2, any function could be a compressor, and any function could be a filter. It wasn't strictly typed at all. So there is no guarantee that something which was a filter in zarr v2 is a valid filter in zarr v3.

virtualizarr/tests/test_writers/test_kerchunk.py
```diff
@@ -31,7 +29,7 @@ def test_accessor_to_kerchunk_dict(self):
             "refs": {
                 ".zgroup": '{"zarr_format":2}',
                 ".zattrs": "{}",
-                "a/.zarray": '{"shape":[2,3],"chunks":[2,3],"dtype":"<i8","fill_value":null,"order":"C","compressor":null,"filters":null,"zarr_format":2}',
+                "a/.zarray": '{"shape":[2,3],"chunks":[2,3],"fill_value":0,"order":"C","filters":[],"dimension_separator":".","compressor":null,"attributes":{},"zarr_format":2,"dtype":"<i8"}',
```
abarciauskas-bgse, line 32: [If the fill_value passed to ArrayV3Metadata is `None`, ArrayV3Metadata will automatically coerce the fill_value to the default of the data type](https://github.com/zarr-developers/zarr-python/blob/v3.0.2/src/zarr/core/metadata/v3.py#L271-L272). For example, fill_value will be `0` if `None` is passed. This means that writing back to v2 metadata will generate a fill_value of the default for the data type, rather than set it as `null`, which is valid for Zarr V2. See also https://github.com/pydata/xarray/issues/5475#issuecomment-2383046633

virtualizarr/tests/test_integration.py
```diff
@@ -230,24 +221,24 @@ def test_non_dimension_coordinates(self, tmpdir, roundtrip_func, hdf_backend):
...
-            # fill_value=0.0,
-            filters=None,
-            order="C",
+        metadata = array_v3_metadata(
             shape=shape,
-            zarr_format=2,
+            chunks=chunks,
+            codecs=[{"name": "numcodecs.zlib", "configuration": {"level": 1}}],
+            # Cannot do in Zarr v3
+            data_type=np.dtype("M8[ns]"),
```
abarciauskas-bgse, line 239: This throws an error `KeyError: '<M8[ns]'` from https://github.com/zarr-developers/zarr-python/blob/v3.0.2/src/zarr/core/metadata/v3.py#L704. Datetime dtypes are not yet supported in zarr-python 3 (see https://zarr.readthedocs.io/en/stable/user-guide/v3_migration.html:

>The following features that were supported by Zarr-Python 2 have not been ported to Zarr-Python 3 yet:
>- Datetime and timedelta dtypes ([#2616](https://github.com/zarr-developers/zarr-python/issues/2616))

virtualizarr/tests/test_readers/test_netcdf3.py
```diff
@@ -1,15 +1,15 @@
...
 
 from virtualizarr import open_virtual_dataset
 from virtualizarr.manifests import ChunkManifest, ManifestArray
 from virtualizarr.tests import requires_scipy
-from virtualizarr.zarr import ZArray
 
 
 @requires_scipy
-def test_read_netcdf3(netcdf3_file):
+@pytest.mark.xfail(reason="zarr-python 3.0 does not support big endian")
```
abarciauskas-bgse, line 11: I don't think the netcdf3 reader will work with ArrayV3Metadata because the data type is big endian, which is not yet supported by zarr-python 3. See [v3 migration guide](https://zarr.readthedocs.io/en/stable/user-guide/v3_migration.html):

>The following features that were supported by Zarr-Python 2 have not been ported to Zarr-Python 3 yet:
>- Big Endian dtypes ([#2324](https://github.com/zarr-developers/zarr-python/issues/2324))

abarciauskas-bgse:
There is still some clean up and organization to work on in this PR, so I am leaving it in draft, but there are a few outstanding issues and questions to address, so I would appreciate input from @TomNicholas @sharkinsspatial @d-v-b when time allows.

## Outstanding issues/questions

- [ ] Do we want to pass around zarr codec objects (`{"name": "zlib", "configuration": {"level": 1}}`) or numcodecs (`{"id": "zlib", "level": 1}`)? I think we have to convert to zarr codec objects to use zarr's get_codec_class anyways. Will this be obsolete once zarr codecs are registered in numcodecs?

The following relate to not-yet-implemented data types and fill value behavior of zarr-python 3.0, so when we use ArrayV3Metadata to write to V2, we get different or unsupported behavior:
- [ ] `fill_value=None` coercion: https://github.com/zarr-developers/VirtualiZarr/pull/429#discussion_r1948386048
- [ ] datetime (and timedelta) data types not supported: https://github.com/zarr-developers/VirtualiZarr/pull/429#discussion_r1948387659
- [ ] big endian not supported: https://github.com/zarr-developers/VirtualiZarr/pull/429#discussion_r1948388834

virtualizarr/codecs.py
```diff
@@ -12,29 +10,39 @@
...
-    Parameters:
-        array (Union[ManifestArray, ZarrArray]): The input array, either ManifestArray or Zarr Array.
 
-    Returns:
-        List[Optional[Codec]]: A list of codecs or an empty list if no codecs are found.
-
-    Raises:
-        ImportError: If `zarr` is required but not installed.
-        ValueError: If the array type is unsupported.
+def get_codecs(array: Union["ManifestArray", "Array"]) -> CodecPipeline:
```
TomNicholas, line 18: nit to make types clearer, see above comment on zarr being a required dependency

```suggestion
import zarr

def get_codecs(array: Union["ManifestArray", "zarr.Array"]) -> CodecPipeline:
```

virtualizarr/codecs.py
```diff
@@ -12,29 +10,39 @@
...
+        A tuple of zarr v3 codecs representing the codec pipeline.
+
+    Raises
+    ------
+    ImportError
+        If `zarr` is required but not installed.
+    ValueError
+        If the array type is unsupported or the array's metadata is not in zarr v3 format.
+    NotImplementedError
+        If zarr-python v3 is not installed.
```
TomNicholas, line 39: Why both `ImportError` and `NotImplementedError`?

virtualizarr/tests/test_integration.py
TomNicholas: This restriction must apply to loadable variables too? In which case this is potentially a blocker?

TomNicholas (commented):
> Do we want to pass around zarr codec objects (`{"name": "zlib", "configuration": {"level": 1}}`) or numcodecs (`{"id": "zlib", "level": 1}`)

Presumably we just want to do it however zarr-python v3 does it internally?

> not-yet-implemented data types and fill value behavior of zarr-python 3.0

So do we:
- Wait to fix those upstream before merging this PR,
- Merge this PR with xfails and wait to fix those upstream before releasing next
- Merge this PR and release VirtualiZarr even if some features become broken until fixed upstream.

I think I would prefer option 2, but it depends how long it's likely to take to fix these things upstream.

virtualizarr/codecs.py
```diff
@@ -1,6 +1,4 @@
-from typing import TYPE_CHECKING, Union
-
-from virtualizarr.zarr import Codec
+from typing import TYPE_CHECKING, Tuple, Union
```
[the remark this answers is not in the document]
abarciauskas-bgse: yes I think zarr will be a required dependency now

virtualizarr/zarr.py
```diff
@@ -238,4 +212,121 @@ def _num_codec_config_to_configurable(num_codec: dict) -> dict:
...
+    codecs: list[dict] | None = [],
+) -> tuple[ZarrCodec, ...]:
+    """
+    Convert compressor, filters, serializer, and dtype to a pipeline of ZarrCodecs.
+
+    Parameters
+    ----------
+    dtype : Any
+        The data type.
+    codecs: list[ZarrCodec] | None
```
abarciauskas-bgse, line 245: fix this parameter description

TomNicholas:
> Mark those data types (datetime, timedelta and big endian data types) as not supported in the next release, and still release this after merging (and testing).

This seems basically the same as my option 3?

> Create a not-for-release feature branch for zarr-python-3

Given that both @ayushnag 's `ManifestStore` and @raphael's zarr reader are blocked on this, it would be nice to have a branch that they can merge from. So maybe that suggestion could work? We would just need to keep close eye on the sync'ing as you say.

Does anyone have strong feelings about that (@jsignell @maxrjones ?)

norlandrhagen:
> Yes my first suggestion was similar to your third suggestion.
> 
> It sounds like we are leaning towards a feature branch (zarr-python-3.0?) but I'll wait for others to chime in. cc @norlandrhagen @ayushnag

That option seems great to me. Thanks for moving this along @abarciauskas-bgse

virtualizarr/manifests/array_api.py
abarciauskas-bgse: https://github.com/zarr-developers/VirtualiZarr/pull/429/commits/aad511f86ff1e89dd4cf6b37252bf6adf3d3d289

virtualizarr/manifests/array.py
abarciauskas-bgse: https://github.com/zarr-developers/VirtualiZarr/pull/429/commits/a0f72b2517accd63e7228a146e9a7fdaf87e1c8a

abarciauskas-bgse:
I'm going to continue to review this tomorrow but the tests are passing and I've done an initial reorganization of the code that was in zarr.py. So if any of @TomNicholas @norlandrhagen @ayushnag @sharkinsspatial @jsignell @mpiannucci want to start to review please go ahead 🙏🏽

virtualizarr/manifests/array_api.py
```diff
@@ -236,12 +253,11 @@ def broadcast_to(x: "ManifestArray", /, shape: tuple[int, ...]) -> "ManifestArra
         lengths=broadcasted_lengths,
     )
 
-    new_zarray = x.zarray.replace(
-        chunks=new_chunk_shape,
-        shape=new_shape,
+    new_metadata = _update_metadata(
+        x, new_shape=list(new_shape), new_chunks=list(new_chunk_shape)
```
TomNicholas, line 257:
```suggestion
    new_metadata = _update_metadata(
        x, new_shape=new_shape, new_chunks=new_chunk_shape
```

virtualizarr/readers/dmrpp.py
```diff
@@ -406,24 +408,33 @@ def _parse_variable(self, var_tag: ET.Element) -> Variable:
...
+                "configuration": {"chunk_shape": chunks_shape},
+            },
+            chunk_key_encoding={"name": "default"},
+            fill_value=fill_value,
+            codecs=convert_to_codec_pipeline(
+                codecs=codecs,
+                dtype=dtype,
+            ),
+            attributes=attrs,
+            dimension_names=None,
```
TomNicholas, line 434: Do we want to specify attrs here? I feel like we want neither `dimension_names` nor `attrs`, because in virtualizarr those both are stored in the xarray objects instead.

virtualizarr/readers/dmrpp.py
```diff
@@ -406,24 +408,33 @@ def _parse_variable(self, var_tag: ET.Element) -> Variable:
...
         fill_value = attrs.pop("_FillValue", None)
-        # create ManifestArray and ZArray
-        zarray = ZArray(
-            chunks=chunks_shape,
-            dtype=dtype,
-            fill_value=fill_value,
-            filters=filters,
-            order="C",
+        # create ManifestArray
+        metadata = ArrayV3Metadata(
```
TomNicholas, line 420: This call seems to have quite a lot of boilerplate that authors of virtualizarr readers will never need, so as it's generally useful perhaps we should provide our own constructor function in `.utils`?

virtualizarr/tests/test_backend.py
```diff
@@ -230,9 +229,10 @@ class TestReadFromURL:
                 "grib",
                 "https://github.com/pydata/xarray-data/raw/master/era5-2mt-2019-03-uk.grib",
             ),
-            (
+            pytest.param(
                 "netcdf3",
                 "https://github.com/pydata/xarray-data/raw/master/air_temperature.nc",
+                marks=pytest.mark.xfail(reason="Not supported by zarr-python 3.0"),
```
TomNicholas, line 235: Link to issue here else it's not clear that this is an endian problem

virtualizarr/tests/test_codecs.py
```diff
@@ -1,161 +1,141 @@
...
-from virtualizarr.tests import (
-    requires_zarr_python,
-    requires_zarr_python_v3,
-)
-from virtualizarr.zarr import Codec
+from virtualizarr.codecs import convert_to_codec_pipeline, get_codecs
 
+arrayarray_codec = {"name": "numcodecs.delta", "configuration": {"dtype": "<i8"}}
+arraybytes_codec = {"name": "bytes", "configuration": {"endian": "little"}}
+bytesbytes_codec = {
```
TomNicholas, line 11: Following convention for global variables

```suggestion
ARRAYARRAY_CODEC = {"name": "numcodecs.delta", "configuration": {"dtype": "<i8"}}
ARRAYBYTES_CODEC = {"name": "bytes", "configuration": {"endian": "little"}}
BYTESBYTES_CODEC = {
```

virtualizarr/tests/test_codecs.py
```diff
@@ -1,161 +1,141 @@
...
-        expected_codecs = Codec(
-            compressor=Blosc(
-                cname="zstd", clevel=5, shuffle=Blosc.SHUFFLE, blocksize=0
-            ),
-            filters=[
-                Delta(dtype="<i4"),
-            ],
+    def test_manifest_array_zarr_v3_with_codecs(self, create_manifestarray):
+        """Test get_codecs with ManifestArray using multiple v3 codecs."""
+        test_codecs = [
```
TomNicholas, line 52: This could also be a global variable `TEST_CODECS`, or even a fixture providing multiple sets of test codecs. See also @dcherian's `codecs` hypothesis strategy that we could use in future (https://github.com/zarr-developers/zarr-python/pull/2822).

virtualizarr/tests/test_manifests/test_array.py
```diff
@@ -212,115 +220,83 @@ def test_broadcast_empty(self, shape, chunks, grid_shape, target_shape):
...
-        zarray = ZArray(
-            chunks=(5, 1, 10),
-            compressor={"id": "zlib", "level": 1},
-            dtype=np.dtype("int32"),
-            fill_value=0.0,
-            filters=None,
-            order="C",
-            shape=(5, 1, 20),
-            zarr_format=2,
+    # FAILING: TypeError: no implementation found for 'numpy.concatenate' on types that implement __array_function__: [<class 'virtualizarr.manifests.array.ManifestArray'>, <class 'numpy.ndarray'>]
```
TomNicholas, line 245: This is a bad bug and shouldn't be forgotten - it implies that somehow inside the manifest / `ManifestArray` concatenation implementation some array is the wrong type.

(This can't be being thrown from the top-level `np.concatenate` you can see in the test, so it must be being thrown by the lower-level `np.concatenate` call that is actually used to merge the manifests internally in `.array_api`.)

virtualizarr/translators/kerchunk.py
```diff
@@ -1,17 +1,104 @@
...
+
+    Parameters:
+    ----------
+    decoded_arr_refs_zarray : dict
+        A dictionary containing the decoded Zarr array reference information.
+        Expected keys include "dtype", "fill_value", "zarr_format", "filters",
+        "compressor", "chunks", and "shape".
+    Returns:
+    -------
+    ArrayV3Metadata
```
TomNicholas, line 59:
```suggestion
    ArrayV3Metadata
    
```

virtualizarr/writers/icechunk.py
```diff
@@ -242,21 +243,17 @@ def write_virtual_variable_to_icechunk(
         )
     else:
         append_axis = None
-
-        # Get the codecs and convert them to zarr v3 format
-        codecs = zarray._v3_codecs()
-
-        # create array if it doesn't already exist
+        # TODO: Should codecs be an argument to zarr's AsyncrGroup.create_array?
```
TomNicholas, line 246: You're asking for an upstream change here right?

virtualizarr/writers/kerchunk.py
```diff
@@ -67,13 +90,65 @@ def remove_file_uri_prefix(path: str):
         return path
 
 
+def convert_v3_to_v2_metadata(
```
TomNicholas, line 93: Should this (a) not live in the kerchunk-specific writer module, (b) actually live in zarr-python upstream? Or is there no use for it upstream?

TomNicholas (commented):
This is absolutely great @abarciauskas-bgse ! Comments are really just minor.

virtualizarr/tests/test_writers/test_kerchunk.py
sharkinsspatial: IIUC based on our discussion in https://github.com/zarr-developers/VirtualiZarr/issues/414 and https://github.com/pydata/xarray/issues/5475#issuecomment-2383046633.  This is the behavior we want and expect.  The v3 semantics are that `fill_value` denotes uninitialized regions of the array so it seems that returning a value with a consistent dtype would probably be what we want.

virtualizarr/tests/test_integration.py
sharkinsspatial: @TomNicholas I might be misunderstanding here but `loadable_variables` will never reach the `ArrayV3Metadata`.  In the interim until there is upstream support for Datetime `dtypes` should our readers introspect variables and force Datetime variables as `loadable_variables`?  I know this seems a bit heavy handed but it might give users a better experience in the short term.

TomNicholas: > `loadable_variables` will never reach the ArrayV3Metadata

No, but we're still saving the `loadable_variables` into zarr v3 via icechunk as native chunks eventually. So if zarr doesn't support this dtype don't we still have a problem? Or is there some encoding step that dodges the issue?

> force Datetime variables as loadable_variables?

I would be fine with that, and actually after #353 that would mostly happen by default for realistic datasets.

docs/usage.md
```diff
@@ -134,20 +139,49 @@ A Zarr array is defined not just by the location of its constituent chunk data,
...
 ```
+
 ```
-ZArray(shape=(2920, 25, 53), chunks=(2920, 25, 53), dtype=int16, compressor=None, filters=None, fill_value=None)
+ArrayV3Metadata(shape=(2920, 25, 53),
+                data_type=<DataType.float64: 'float64'>,
+                chunk_grid=RegularChunkGrid(chunk_shape=(2920, 25, 53)),
+                chunk_key_encoding=DefaultChunkKeyEncoding(name='default',
+                                                           separator='/'),
+                fill_value=np.float64(-327.67),
```
abarciauskas-bgse, line 165: Seeing the full representation I looked into why the fill value was `np.float64(-327.67)` as that fill value was not in the xarray encoding but was the fill value of the `h5py.Dataset`, is it worth digging into why that float value wasn't in the xarray encoding? @sharkinsspatial

docs/usage.md
```diff
@@ -134,20 +139,49 @@ A Zarr array is defined not just by the location of its constituent chunk data,
...
 ```
+
 ```python
-marr.zarray
+marr.metadata
 ```
+
 ```
-ZArray(shape=(2920, 25, 53), chunks=(2920, 25, 53), dtype=int16, compressor=None, filters=None, fill_value=None)
+ArrayV3Metadata(shape=(2920, 25, 53),
```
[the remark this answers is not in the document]
abarciauskas-bgse: I like the idea of replacing the markdown with rendered output of notebooks. Is there anything you would like me to change here for now?

conftest.py
```diff
@@ -150,3 +158,222 @@ def simple_netcdf4(tmp_path: Path) -> str:
...
+
+        chunkmanifest = ChunkManifest(entries=d)
+
+        return ManifestArray(chunkmanifest=chunkmanifest, metadata=metadata)
+
+    return _create_manifestarray
+
+
+@pytest.fixture
+def array_v3_metadata():
```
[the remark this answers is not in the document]
d-v-b: have a look at the signature of this function, which has a lot of sane defaults, and which works for v2 and v3 metadata: https://github.com/zarr-developers/zarr-python/blob/99621ecf0b81400e323828111363fe21cf0c7592/src/zarr/core/array.py#L4008-L4030. I think we could consider adding an `ArrayV3Metadata.build` method that has a signature like this, which should make creating metadata documents a lot easier.

docs/usage.md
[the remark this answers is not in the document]
TomNicholas: No, no action needed in this PR, just pointing out an improvement for someone to do later.

virtualizarr/readers/dmrpp.py
abarciauskas-bgse: `dimension_names` is a required argument (even though it can be `None`, I'm not sure why `None` is not set as a default in the [__init__ method](https://github.com/zarr-developers/zarr-python/blob/v3.0.2/src/zarr/core/metadata/v3.py#L259-L260). I have removed setting attributes from this reader in https://github.com/zarr-developers/VirtualiZarr/pull/429/commits/86d1de5af91501fbdd0b0d8e2f8545645993a7d1

Should we request upstream for a default of None to be allowed in the __init__ method of ArrayV3Metadata?

virtualizarr/readers/dmrpp.py
abarciauskas-bgse: Updated in https://github.com/zarr-developers/VirtualiZarr/pull/429/commits/86d1de5af91501fbdd0b0d8e2f8545645993a7d1

virtualizarr/tests/test_integration.py
sharkinsspatial: @TomNicholas As discussed in https://github.com/earth-mover/icechunk/issues/119 if we are using `xarray` CF datetime encoding the raw array dtype for serialization will be `int` and `xarray` will use the `units` `attr` for decoding.

TomNicholas: Oh great, thanks for the explanation/reminder @sharkinsspatial .

virtualizarr/tests/test_codecs.py
abarciauskas-bgse: I've implemented the codecs as global vars, and implemented a multiple scenario test in test_codecs with various codec scenarios in https://github.com/zarr-developers/VirtualiZarr/pull/429/commits/b5a1dc69b29da135e1021c3c994b34fe48071f48 - do you think this is a good solution in lieu of using hypothesis strategy (I need to read up more on `hypothesis`)

virtualizarr/tests/test_codecs.py
abarciauskas-bgse: https://github.com/zarr-developers/VirtualiZarr/pull/429/commits/97bc7cd26136e0b0cafb6fc8f120cd42f793465e

virtualizarr/tests/test_manifests/test_array.py
abarciauskas-bgse: Is there an issue for this so it is not forgotten? This seems unrelated to this PR, right?

virtualizarr/codecs.py
```diff
@@ -1,109 +1,133 @@
...
+
+    num_codec_copy = num_codec.copy()
+    name = "numcodecs." + num_codec_copy.pop("id")
+    return {"name": name, "configuration": num_codec_copy}
+
+
+def extract_codecs(
+    codecs: CodecPipeline,
+) -> tuple[
+    tuple[ArrayArrayCodec, ...], ArrayBytesCodec | None, tuple[BytesBytesCodec, ...]
```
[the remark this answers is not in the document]
abarciauskas-bgse: I think it is correct, I have added a type alias, but essentially we just need a way of extracting out the various codec types so we can insert an ArrayBytesCodec (when there is none, which is always in this codebase) in-between ArrayArrayCodecs and BytesBytesCodecs

conftest.py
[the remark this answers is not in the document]
abarciauskas-bgse: I revisited this upon also building a utils method for `create_v3_array_metadata` that readers can use. Now the fixture just calls that function and I removed the array_v3_metadata_dict altogether (and call to_dict in tests where I want to test the construction from dict functionality).

virtualizarr/tests/test_manifests/test_array.py
TomNicholas: There is now! #433.

virtualizarr/tests/test_codecs.py
TomNicholas: That's great. The codecs are now parametrizable, which is what we need to generalize those parameters later to a whole hypothesis strategy.

virtualizarr/tests/test_manifests/test_array.py
TomNicholas: @abarciauskas-bgse this test passes though doesn't it? It's not xfailed... So is there actually a bug here to fix?

docs/usage.md
abarciauskas-bgse: Discussing this in a separate thread, will follow up once resolved...

virtualizarr/tests/test_backend.py
abarciauskas-bgse: https://github.com/zarr-developers/VirtualiZarr/pull/429/commits/c245b0a5c2d1c6c4eba22557b22c497473284853

virtualizarr/tests/test_manifests/test_array.py
abarciauskas-bgse: My apologies @TomNicholas for the confusion: I didn't recall writing this comment so I thought it was unrelated to this PR. I think we can close #433 as this must have been a failure that occurred in the process of this refactor.

conftest.py
```diff
@@ -1,13 +1,25 @@
+"""Pytest configuration and fixtures for virtualizarr tests."""
```
abarciauskas-bgse, line 1: I reorganized this file as it includes fixtures, variables and helper functions of various purpose.

docs/usage.md
abarciauskas-bgse: The outcome of this side conversation was that we have not yet resolved https://github.com/zarr-developers/VirtualiZarr/issues/414 via https://github.com/zarr-developers/VirtualiZarr/pull/420 and that the way to do this is by including the HDF5 attribute `_FillValue` in the Xarray Variable's encoding. And then when writing the dataset to icechunk, it will be included as an attr (see https://github.com/zarr-developers/VirtualiZarr/blob/c245b0a5c2d1c6c4eba22557b22c497473284853/virtualizarr/writers/icechunk.py#L263-L266)

virtualizarr/readers/hdf/hdf.py
```diff
@@ -283,9 +284,6 @@ def _dataset_to_variable(
         list: xarray.Variable
             A list of xarray variables.
         """
-        # This chunk determination logic mirrors zarr-python's create
-        # https://github.com/zarr-developers/zarr-python/blob/main/zarr/creation.py#L62-L66
```
abarciauskas-bgse, line 287: I removed this comment because I think the reference is from a previous version of zarr-python - @sharkinsspatial do you know if we can and should include an updated link?

virtualizarr/codecs.py
```diff
@@ -1,109 +1,136 @@
...
+    """
+    if num_codec["id"].startswith("numcodecs."):
+        return num_codec
+
+    num_codec_copy = num_codec.copy()
+    name = "numcodecs." + num_codec_copy.pop("id")
+    return {"name": name, "configuration": num_codec_copy}
+
+
+def extract_codecs(
```
abarciauskas-bgse, line 34: 🚧 needs a test

virtualizarr/codecs.py
```diff
@@ -1,109 +1,136 @@
...
-    """Check if the array is an instance of Zarr Array."""
-    try:
-        from zarr import Array
+    return codec_pipeline
 
-        return isinstance(array, Array)
-    except ImportError:
-        return False
 
+def get_codec_config(codec: ZarrCodec) -> dict[str, Any]:
```
abarciauskas-bgse, line 93: 🚧 needs a test

virtualizarr/translators/kerchunk.py
```diff
@@ -1,17 +1,96 @@
...
+            for codec in v2_metadata.filters
+            if codec is not None
+        ]  # type: ignore[assignment]
+    if v2_metadata.compressor:
+        zarray_dict["compressor"] = v2_metadata.compressor.get_config()  # type: ignore[assignment]
+
+    return json.dumps(zarray_dict, separators=(",", ":"), cls=NumpyEncoder)
+
+
+def from_kerchunk_refs(decoded_arr_refs_zarray) -> "ArrayV3Metadata":
```
abarciauskas-bgse, line 45: To compare with https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/zarr.py#L99-L122

virtualizarr/translators/kerchunk.py
```diff
@@ -1,17 +1,96 @@
...
 from virtualizarr.readers.common import separate_coords
 from virtualizarr.types.kerchunk import (
     KerchunkArrRefs,
     KerchunkStoreRefs,
 )
-from virtualizarr.zarr import ZArray, ZAttrs, determine_chunk_grid_shape
+from virtualizarr.utils import determine_chunk_grid_shape
+
+
+def to_kerchunk_json(v2_metadata: ArrayV2Metadata) -> str:
```
abarciauskas-bgse, line 25: to compare with https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/zarr.py#L129-L135

virtualizarr/manifests/utils.py
```diff
@@ -116,3 +167,23 @@ def check_compatible_arrays(
     check_same_ndims([ma.ndim, existing_array.ndim])
     arr_shapes = [ma.shape, existing_array.shape]
     check_same_shapes_except_on_concat_axis(arr_shapes, append_axis)
+
+
+def copy_and_replace_metadata(
```
abarciauskas-bgse, line 172: 🚧 needs a test

virtualizarr/translators/kerchunk.py
abarciauskas-bgse: This function is mostly the same except we coerce filters and codecs to valid codec configurations for zarr-python 3.0 so we can then construct `ArrayV3Metadata`

virtualizarr/translators/kerchunk.py
abarciauskas-bgse: This function is a bit more complex than it was before because we are going from zarr.core.metadata.v2.ArrayV2Metadata to json (rather than from ZArray.dict() to json). So we have to convert the filters and compressor to serializable objects using `get_config`.

virtualizarr/tests/test_integration.py
abarciauskas-bgse: Should we include any test or documentation about the xarray method of datetime encoding @sharkinsspatial @TomNicholas ? Perhaps as a separate issue since it's not directly related to this PR, apart from being an approach to get around the current zarr-python datetime/timedelta data type limitation.

virtualizarr/writers/icechunk.py
abarciauskas-bgse: I think so, should I create an issue for that?

virtualizarr/writers/kerchunk.py
abarciauskas-bgse: >(a) not live in the kerchunk-specific writer module

This is the only module that requires it, but I'm open to moving it, should I put it in `utils.py`?

>(b) actually live in zarr-python upstream? Or is there no use for it upstream?

I'm not sure if there is a use for it, but I could create another issue if you think it is warranted to ask.

abarciauskas-bgse:
@TomNicholas this PR is ready for re-review and 🤞🏽 hopefully good to merge (to `zarr-python-3.0`). There are a few outstanding comments that are mostly related to creating new issues:

* https://github.com/zarr-developers/VirtualiZarr/pull/429/files#r1948387659
* https://github.com/zarr-developers/VirtualiZarr/pull/429/files#r1953785807
* https://github.com/zarr-developers/VirtualiZarr/pull/429/files#r1953787670

And then I might also create a ticket to test this branch against all (or most) current examples.

virtualizarr/writers/kerchunk.py
TomNicholas: Let's just put it in `utils.py` for now, as it's not specifically related to kerchunk.

@d-v-b is there anything like this / any need for this function upstream?

virtualizarr/tests/test_integration.py
TomNicholas: Raised #437 to track that.

virtualizarr/writers/icechunk.py
TomNicholas: If you think it would be useful then go for it. Shouldn't stop us merging this though.

TomNicholas (approved):
Thank you @abarciauskas-bgse ! Let's just get this merged into a stable branch so that others can build on it.

virtualizarr/writers/icechunk.py
abarciauskas-bgse: https://github.com/zarr-developers/zarr-python/issues/2848

norlandrhagen:
Awesome work @abarciauskas-bgse!

Issue #442, opened 2025-02-20
Use Icechunk's new `.set_virtual_chunks` method


TomNicholas:
In https://github.com/earth-mover/icechunk/issues/734 Seba added a more efficient method for setting virtual chunks - setting many in one method call instead of one per method call. It still requires iterating over every chunk in python but does seem to be much faster than what we have right now.

We should switch to using the new method internally, which should be a very simple change.

norlandrhagen:
Happy to take a stab at this.

TomNicholas:
Go for it!

TomNicholas:
If you want to check the performance you could use the same trick I used in the notebook here: https://github.com/earth-mover/icechunk/issues/401

TomNicholas:
Closed by #443

PR #443, opened 2025-02-20
Update `store.set_virtual_ref` -> `store.set_virtual_refs`

norlandrhagen:
- [x] Closes [#442](https://github.com/zarr-developers/VirtualiZarr/issues/442)
- [x] Tests passing

### DIFF
--- a/virtualizarr/writers/icechunk.py
+++ b/virtualizarr/writers/icechunk.py
@@ -306,6 +306,7 @@ def write_manifest_virtual_refs(
     last_updated_at: Optional[datetime] = None,
 ) -> None:
     """Write all the virtual references for one array manifest at once."""
+    # from icechunk import VirtualChunkSpec
 
     key_prefix = f"{group.name}/{arr_name}"
 
--- a/virtualizarr/writers/icechunk.py
+++ b/virtualizarr/writers/icechunk.py
@@ -325,22 +325,15 @@ def write_manifest_virtual_refs(
         op_flags=[["readonly"]] * 3,  # type: ignore
     )
 
-    virtual_chunk_spec_list = []
-
-    for path, offset, length in it:
-        # it.multi_index will be an iterator of the chunk shape
-        index = it.multi_index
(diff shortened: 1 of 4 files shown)


norlandrhagen:
It seems like bumping Icechunk to `0.2.0` caused a few appending test failures.

cc @abarciauskas-bgse @TomNicholas 
```python 
FAILED virtualizarr/tests/test_writers/test_icechunk.py::TestAppend::test_append_virtual_ref_without_encoding - KeyError: '///foo'
FAILED virtualizarr/tests/test_writers/test_icechunk.py::TestAppend::test_append_virtual_ref_with_encoding - KeyError: '///air'
FAILED virtualizarr/tests/test_writers/test_icechunk.py::TestAppend::test_append_with_compression_succeeds[2] - KeyError: '///air'
FAILED virtualizarr/tests/test_writers/test_icechunk.py::TestAppend::test_append_with_compression_succeeds[3] - KeyError: '///air'
```

TomNicholas:
Aren't you about to change the part of the write code that causes this failure in this PR anyway? I suggest you push on with this PR and see if it fixes it.

virtualizarr/writers/icechunk.py
```diff
@@ -323,17 +325,22 @@ def write_manifest_virtual_refs(
         op_flags=[["readonly"]] * 3,  # type: ignore
     )
 
+    virtual_chunk_spec_list = []
```
norlandrhagen, line 328: This builds up a list of `VirtualChunkSpec` objects then does a single `set_virtual_refs`. Should this be batched or handled differently?

TomNicholas: I think this is what we want to do in this PR at least. We will gain the benefit of not going back and forth between python and rust on every chunk, only on each array.

Also you could use a list comprehension instead of creating an empty list and then appending to it

norlandrhagen: Good call!

abarciauskas-bgse:
LGTM!

TomNicholas (approved):

PR #453, opened 2025-02-27
Require zarr-python>=3.0

TomNicholas:
This PR just merges the other development branch, and so is meant as a way to check the testing status of trying to pin zarr-python>=3.0 by merging @abarciauskas-bgse 's work in #429.

- [x] Tests passing


abarciauskas-bgse:
@TomNicholas thanks for opening this - I just looked at the test failures and it appears the failure is because in the kerchunk writer, we are expecting `.zarray["filters"]` to be `[]` but the value for `.zarray["filters"]` is actually `null`. Would you like me to investigate where this change is coming from or do you have an idea?

TomNicholas:
Without looking any further - I have no idea. So you're just as likely to be able to fix that as I am 😀

abarciauskas-bgse:
@TomNicholas working with @sharkinsspatial we found this change in zarr-python 3.0.4 release caused the change in filters from `[]` to `null`, causing the test failure: https://github.com/zarr-developers/zarr-python/pull/2847. I have updated the test expectation so now tests are passing.

TomNicholas:
Woo! Does this mean we can merged this?? Or am I forgetting other things that would break (endianess?)

TomNicholas:
(Also @abarciauskas-bgse I'm not sure how commit credit really works in GH but if you wanted the repo history to definitely reflect all the work you did here I would be happy to close this PR in favour of you opening and merging one that does the same thing.)

abarciauskas-bgse:
I can reopen it, thanks @TomNicholas, but it does still include xfailing tests for NetCDF3, FITS (which use big endian) and datetime data type. It sounded like @d-v-b could be working on adding more data type support?

abarciauskas-bgse:
(I'll leave this PR open until we determine if we're ok with allowing those data types to go unsupported for now)

TomNicholas:
FITS is unimportant IMO, I doubt even a single person is using it.

The datetime data type matters, but will become less important once I change the default to load all 1D coordinates into memory, as I assume that most occurrences of the datetime type are in small coordinate variables that will or can be loaded.

netCDF3 is more annoying, as there is no workaround for that. But I'm also not sure it's worth holding this up, as merging this will enable lots of good stuff.

PR #454, opened 2025-02-28
Mark hanging test as xfail

maxrjones:
cc @TomNicholas and @sharkinsspatial 

The dependencies are all the same between the passing and hanging workflows, so I think this was actually cause by a change in https://github.com/zarr-developers/VirtualiZarr/pull/420. We should probably revert the PR if it might take a while to fix the issue. 

This is the only test that uses the `netcdf4_inlined_ref` fixture, so I think it's probably something going on there but I haven't looked at all.

### DIFF
--- a/virtualizarr/tests/test_readers/test_kerchunk.py
+++ b/virtualizarr/tests/test_readers/test_kerchunk.py
@@ -233,6 +233,9 @@ def test_open_virtual_dataset_existing_kerchunk_refs(
 
 
 @requires_kerchunk
+@pytest.mark.xfail(
+    reason="Test hangs after https://github.com/zarr-developers/VirtualiZarr/pull/420"
+)
 def test_notimplemented_read_inline_refs(tmp_path, netcdf4_inlined_ref):
     # For now, we raise a NotImplementedError if we read existing references that have inlined data
     # https://github.com/zarr-developers/VirtualiZarr/pull/251#pullrequestreview-2361916932
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,6 +29,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.11", "3.12"]
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/min-deps.yml
+++ b/.github/workflows/min-deps.yml
@@ -29,6 +29,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.12"]
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -29,6 +29,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.12"]
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
 


maxrjones:
Pretty weird one - I can reproduce using [act](https://github.com/nektos/act) with the Python 3.11 environment, but not locally. I wonder if it's an os-specific issue since we only use ubuntu for the tests. I'm not sure how to debug within the docker container. @chuckwondo do you have experience with that?

TomNicholas:
Thanks for looking into this @maxrjones!

I clearly shouldn't have merged the PR, feel free to revert it.

chuckwondo:
> Pretty weird one - I can reproduce using [act](https://github.com/nektos/act) with the Python 3.11 environment, but not locally. I wonder if it's an os-specific issue since we only use ubuntu for the tests. I'm not sure how to debug within the docker container. @chuckwondo do you have experience with that?

Sorry, I'm not following. I see that this build failed, but it shows as canceled, and I cannot see why: https://github.com/zarr-developers/VirtualiZarr/actions/runs/13580434648/job/37998176988

Where are you seeing something hanging? What are you doing to reproduce via act?

chuckwondo:
@maxrjones, I pull the latest from main (containing #420 merged) and ran `act -j test --matrix python-version:3.11` and it succeeded. Am I missing something, or was there perhaps a transient network issue in the network tests that were perhaps causing whatever "hanging" you're referring to?

maxrjones:
I think the issue is an issue with file locking due to [this parameterized fixture](https://github.com/zarr-developers/VirtualiZarr/blob/c461b3310b9ff9b4e1bb7fadd49ef12de7d13832/virtualizarr/tests/test_readers/conftest.py#L374-L395) writing to the same file. I may have a fix within in the next half hour, otherwise I'll reach out for help again.

Sorry for causing confusion @chuckwondo - I cancelled the workflow manually after a half hour since it should only take a couple minutes. 

I am convinced this is not strictly a network issue based on having also seen the test hang (i.e., run for more than 500% the usual time) using `act -j test`. It may be necessary to run multiple matrix options in order for the file locking to occur, and even then it might be intermittent.

virtualizarr/tests/test_readers/conftest.py
```diff
@@ -372,18 +372,22 @@ def scalar_fill_value_hdf5_file(tmpdir):
...
+    },
+    {"label": "NaN", "fill_value": "NaN", "data": np.array(["three"], dtype="S10")},
+    {"label": "compound", "fill_value": compound_fill, "data": compound_data},
 ]
 
 
 @pytest.fixture(params=fill_values)
 def cf_fill_value_hdf5_file(tmpdir, request):
-    filepath = f"{tmpdir}/cf_fill_value.nc"
+    filepath = f"{tmpdir}/cf_fill_value_{request.param['label']}.nc"
```
chuckwondo, line 390: How about changing `tmpdir` to `tmp_path` (as pytest has replaced the former with the latter), and using it like so:

```suggestion
    filepath = tmp_path / f"cf_fill_value_{request.param['label']}.nc"
```

I'd even suggest annotating the function parameter (`tmp_path: Path`)

chuckwondo (approved):
A couple of ignorable suggestions, so I've approved in case you choose to ignore them.

chuckwondo:
> I think this is an issue with file locking due to [this parameterized fixture](https://github.com/zarr-developers/VirtualiZarr/blob/c461b3310b9ff9b4e1bb7fadd49ef12de7d13832/virtualizarr/tests/test_readers/conftest.py#L374-L395) writing to the same file. I may have a fix within in the next half hour, otherwise I'll reach out for help again.
> 
> Sorry for causing confusion @chuckwondo - I cancelled the workflow manually after a half hour since it should only take a couple minutes.
> 
> I am convinced this is not strictly a network issue based on having also seen the test hang (i.e., run for more than 500% the usual time) using `act -j test`. It may be necessary to run multiple matrix options in order for the file locking to occur, and even then it might be intermittent.

Thanks for clarifying!

virtualizarr/tests/test_readers/test_kerchunk.py
```diff
@@ -201,7 +201,7 @@ def test_open_virtual_dataset_existing_kerchunk_refs(
         # Test valid json and parquet reference formats
 
         if reference_format == "json":
-            ref_filepath = tmp_path / "ref.json"
+            ref_filepath = tmp_path / "ref1.json"
```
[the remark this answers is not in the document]
maxrjones: thanks, I'll change this back.

virtualizarr/tests/test_readers/conftest.py
```diff
@@ -387,7 +387,7 @@ def scalar_fill_value_hdf5_file(tmpdir):
 
 @pytest.fixture(params=fill_values)
 def cf_fill_value_hdf5_file(tmpdir, request):
-    filepath = tmp_path / f"cf_fill_value_{request.param['label']}.nc"
+    filepath = tmpdir / f"cf_fill_value_{request.param['label']}.nc"
```
chuckwondo, line 390: You can keep `tmpdir`, but pytest has replaced it with `tmp_path`, so it's a typo only if you don't also change the name of the function parameter :-)

maxrjones: I'll leave this for later, just hoping to get the tests all green before heading out for the weekend :)

maxrjones:
> @maxrjones, I don't think this is the issue.
> 
> I see the current build hanging at this point:
> 
> ```
> virtualizarr/tests/test_readers/test_kerchunk.py::test_open_virtual_dataset_existing_kerchunk_refs[json] PASSED [ 81%]
> virtualizarr/tests/test_readers/test_kerchunk.py::test_open_virtual_dataset_existing_kerchunk_refs[invalid] PASSED [ 82%]
> ```
> 
> The other build hung similarly:
> 
> ```
> virtualizarr/tests/test_readers/test_kerchunk.py::test_open_virtual_dataset_existing_kerchunk_refs[json] PASSED [ 82%]
> virtualizarr/tests/test_readers/test_kerchunk.py::test_open_virtual_dataset_existing_kerchunk_refs[invalid] PASSED [ 82%]
> virtualizarr/tests/test_readers/test_kerchunk.py::test_open_virtual_dataset_existing_kerchunk_refs[parquet] PASSED [ 82%]
> ```
> 
> These tests don't even use the fixture in question.

Yes, I shouldn't have been using such a shot in the dark approach. I think the changes in this PR are good, but not directed towards the problem.

I opened https://github.com/zarr-developers/VirtualiZarr/pull/455 as a more targeted change, but there are so many warnings about closing the CachingFileManager on an open file that I don't expect it to actually help 😕. I think it might be necessary to do a full evaluation of where files are opened and closed to figure out if it's truly a problem.

It's not impossible that it's a network issue with the xarray tutorial dataset as well. Maybe it's worth saving a copy locally to use and see if the test still hangs. As hooked as I am on the problem, I'm a bit out of time to work on it more now.

sharkinsspatial:
@maxrjones Thanks for investigating this.  Apologies, I wasn't able to reproduce this locally and I hadn't considered using `act` (nice concept @maxrjones 👍 ) in a few attempts so I thought it might be an ephemeral CI issue but apparently not.  Regardless I think improving our file handle hygiene is a good idea going forward.

TomNicholas:
> I think it might be necessary to do a full evaluation of where files are opened and closed to figure out if it's truly a problem.

I think once we have the `ManifestStore` we will be able to reason about this much more clearly. We can then basically change `open_virtual_dataset` to 

1) open the file
2) create `ManifestArrays` for all variables by reading metadata from the file
3) close the file
4) optionally later try to load some of the `ManifestArrays` using the `ManifestStore`, but at that point the reading of data will be implemented using the same internals as a read-only `FsspecStore`/`ObjectStore` internally, which presumably/hopefully manages file handles responsibly.

maxrjones:
Superseded by #460

Issue #390, opened 2025-01-27
RuntimeWarning: deallocating CachingFileManager during main test execution


chuckwondo:
Similar to #385, there are dozens (over 60 as of this writing!) of the following warnings emitted during execution of the main tests:

> RuntimeWarning: deallocating CachingFileManager(...), but file is not already closed. This may indicate a bug.

This is again due to failing to properly close open files in tests/fixtures, so does not appear to be indicative of any virtualizarr bug, but the volume of messages is quite disconcerting to anybody looking at the job logs. Further, these _might_ lead to testing problems at some point in time.

Cleaning up these warnings should more readily make potentially more concerning problems more visible to those who keep an eye on such logs.

sharkinsspatial:
@chuckwondo We may want to consider following the xarray approach of disallowing internal warnings outlined in this issue https://github.com/pydata/xarray/issues/3266 (where the xarray test suite was also issuing the same warning from the netcdf4 library).

chuckwondo:
> [@chuckwondo](https://github.com/chuckwondo) We may want to consider following the xarray approach of disallowing internal warnings outlined in this issue [pydata/xarray#3266](https://github.com/pydata/xarray/issues/3266) (where the xarray test suite was also issuing the same warning from the netcdf4 library).

If I'm reading that correctly, I believe that's a separate issue.

However, it may also be applicable to this repo, but not for the warnings I noted above. The ones above are due to failing to close file pointers. For this particular issue, we want to be properly using context managers in our tests.

Other warnings not due to failing to close file pointers we may certainly want to silence appropriately, to also avoid extraneous logging output that might make it hard to see actual problems.

chuckwondo:
@sharkinsspatial, for reference, see what I did here: https://github.com/zarr-developers/VirtualiZarr/pull/386

TomNicholas:
Was this closed by #386 @chuckwondo ?

chuckwondo:
> Was this closed by [#386](https://github.com/zarr-developers/VirtualiZarr/pull/386) [@chuckwondo](https://github.com/chuckwondo) ?

No, that was a separate set of warnings. The 2 sets of warnings appear for different github workflows. The ones in #386 appear in the upstream tests, whereas the ones here appear in the main tests.

PR #455, opened 2025-03-01
Close open file handle if there aren't loadable_vars

maxrjones:
There's a ton of warnings like the following when running the tests:
```
/Users/max/miniforge3/envs/virtualizarr-tests/lib/python3.11/json/encoder.py:357: RuntimeWarning: deallocating CachingFileManager(<class 'h5netcdf.core.File'>, '/private/var/folders/70/hc_nynms54d8lp67z4rsfctc0000gp/T/pytest-of-max/pytest-22/test_cf_fill_value_roundtrip_c5/cf_fill_value.nc', mode='r', kwargs={'invalid_netcdf': None, 'decode_vlen_strings': True, 'driver': None}, manager_id='9996df4a-48c7-4871-b282-9d4619d8f32e'), but file is not already closed. This may indicate a bug.
```

I think we need to be more careful about open file handles in the readers, this handles one case in which I think it can be closed.


chuckwondo:
@maxrjones, see #390

maxrjones:
> @maxrjones, see #390

You called it 😅

TomNicholas (approved):
Oh nice catch. Yeah the whole `loadable_vars` implementation is a bit hacky atm, but it will be significantly cleaner after #353 and #427.

maxrjones:
> Oh nice catch. Yeah the whole `loadable_vars` implementation is a bit hacky atm, but it will be significantly cleaner after #353 and #427.

Unfortunately this didn't actually fix the issue (test is still hanging). Thanks for the pointers to those PRs - good to know that they exist to not spend much time cleaning up any file handling in the existing code.

PR #457, opened 2025-03-02
Revert "Align HDF reader CF _FillValue with Zarr v3 semantics (#420)"

TomNicholas:
This reverts commit c461b3310b9ff9b4e1bb7fadd49ef12de7d13832, i.e. PR #420, based on @maxrjones' comment in #454.

FYI  @sharkinsspatial @chuckwondo

PR #460, opened 2025-03-03
Reapply HDF reader fillvalue

chuckwondo:
Reapply reverted changes from #420, but with tests modified to use context managers to cleanup resources in an attempt to  prevent the build from hanging during unit tests.

### DIFF
--- a/virtualizarr/readers/hdf/hdf.py
+++ b/virtualizarr/readers/hdf/hdf.py
@@ -8,11 +8,13 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Tuple,
     Union,
 )
 
 import numpy as np
 import xarray as xr
+from xarray.backends.zarr import FillValueCoder
 
 from virtualizarr.manifests import (
     ChunkEntry,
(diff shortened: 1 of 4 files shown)


chuckwondo:
@TomNicholas, @maxrjones, @sharkinsspatial, I believe this fixes the issue with the hanging unit tests that started when #420 was merged. I have reverted Tom's revert in #457 (thus restoring the origin changes from #420) and have updated related unit tests to use context managers so that file resources are cleaned up, and this appears to have fixed the problem.

Using the origin merge of #420, I was able to periodically (not consistently; perhaps every 3rd attempt) reproduce the hanging test issue locally via `act -j test`. However, with the changes in this PR, I was unable to reproduce the problem after many attempts. Given that the PR also avoided hanging on the unit tests, I'm fairly confident that making use of context managers was indeed the trick, and that #390 should perhaps be addressed sooner rather than later.

sharkinsspatial:
Thanks @chuckwondo. 🙇

Issue #468, opened 2025-03-03
File handle resource leak


chuckwondo:
In `virtualizarr/readers/common.py`, the function `maybe_open_loadable_vars_and_indexes` has a file resource leak when a non-empty list of `loadable_variables` is supplied. In this case, [an xarray Dataset is opened](https://github.com/zarr-developers/VirtualiZarr/blob/v1.3.2/virtualizarr/readers/common.py#L53-L58), but never closed (when `loadable_variables` is non-empty), and the dataset is _not_ returned by the function, so the caller has no way of closing it either.

This is visible in numerous warnings emitted during unit testing, of the following form: `RuntimeWarning: deallocating CachingFileManager`

Issue #390 points out these warnings as well, but many of them cannot be addressed without refactoring the `maybe_open_loadable_vars_and_indexes` function to enable proper closing of the currently hidden dataset.

TomNicholas:
> the dataset is not returned by the function, so the caller has no way of closing it either.

This is the problem - I wonder if we can set the `._close` attribute on the `Dataset` that is returned by `open_virtual_dataset` such that it can be closed by the caller.

chuckwondo:
> > the dataset is not returned by the function, so the caller has no way of closing it either.
> 
> This is the problem - I wonder if we can set the `._close` attribute on the `Dataset` that is returned by `open_virtual_dataset` such that it can be closed by the caller.

I see that `maybe_open_loadable_vars_and_indexes` is called from the various implementations of `open_virtual_dataset`, and that each one then calls `construct_virtual_dataset` and returns that value.

Are you saying that we would modify `maybe_open_loadable_vars_and_indexes` to return the dataset it opens, and then set `._close` on the dataset produced by `construct_virtual_dataset` to the dataset returned by `maybe_open_loadable_vars_and_indexes` before returning that dataset produced by `construct_virtual_dataset`?

Something along these lines?

```
...
ds, loadable_vars, indexes = maybe_open_loadable_vars_and_indexes(...)
vds = construct_virtual_dataset(...)

if ds is not None:
    vds._close = ds.close

return vds
```

Or perhaps we could also modify `construct_virtual_dataset` to accept the (possibly `None`) dataset returned by `maybe_open_loadable_vars_and_indexes` and set `._close` there, which would avoid a bit of code duplication.

chuckwondo:
Ah, I see the following comment inside `construct_virtual_dataset`:

```python
[...]

PR #469, opened 2025-03-03
Use context managers in test functions

### LINKED ISSUE #390: RuntimeWarning: deallocating CachingFileManager during main test execution

chuckwondo:
Elimiinate 32 of 70 of the following warnings emitted while running unit tests: `RuntimeWarning: deallocating CachingFileManager`

The remaining 38 appear to all be related to the HDF5VirtualBackend, so they are not addressed in this PR, as they will require some code refactoring, as identified in #468.

Fixes #390

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -61,7 +61,7 @@ def netcdf4_file_with_data_in_multiple_groups(tmp_path: Path) -> str:
 
 
 @pytest.fixture
-def netcdf4_files_factory(tmp_path: Path) -> Callable:
+def netcdf4_files_factory(tmp_path: Path) -> Callable[[], tuple[str, str]]:
     def create_netcdf4_files(
         encoding: Optional[Mapping[str, Mapping[str, Any]]] = None,
     ) -> tuple[str, str]:
@@ -96,7 +96,8 @@ def netcdf4_file_with_2d_coords(tmp_path: Path) -> str:
 def netcdf4_virtual_dataset(netcdf4_file):
     from virtualizarr import open_virtual_dataset
 
-    return open_virtual_dataset(netcdf4_file, indexes={})
+    with open_virtual_dataset(netcdf4_file, indexes={}) as ds:
+        yield ds
 
 
 @pytest.fixture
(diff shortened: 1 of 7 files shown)


TomNicholas:
This looks great @chuckwondo , but I'm confused as to how it actually prevents the warnings. We do not yet have any implementation of the `__exit__` method that xarray's backends use, so how is this working? Is it effectively piggybacking off of xarray's `__exit__` method?

https://github.com/pydata/xarray/blob/0184702f16c3f744fc9096c7dac690626dcc6922/xarray/backends/common.py#L323

chuckwondo:
> This looks great @chuckwondo , but I'm confused as to how it actually prevents the warnings. We do not yet have any implementation of the `__exit__` method that xarray's backends use, so how is this working? Is it effectively piggybacking off of xarray's `__exit__` method?
> 
> https://github.com/pydata/xarray/blob/0184702f16c3f744fc9096c7dac690626dcc6922/xarray/backends/common.py#L323

Right, and that's why my changes do not eliminate all of the warnings. Some of them are eliminated because they introduce context managers where xarray datasets (and other things that deal directly with files) are directly used. The rest remain precisely because of what you mention (hence, why I also opened #468), but I added the context managers in these places anyway, for consistency as well as for automatically taking advantage of the necessary changes. Once the other related issues are addressed, the remaining warning should automatically go away as well.

If you like, I can attempt to pare down the changes to _only_ the places that eliminate warnings _now_, rather than including context managers that will only eliminate warnings later, when the other related changes land.

TomNicholas:
Thank you @chuckwondo , that makes sense. I just wanted to check I wasn't going to merge this without understanding it :)

TomNicholas (approved):

PR #475, opened 2025-03-05
Add mursst lithops example

abarciauskas-bgse:


TomNicholas:
This is cool! If you're happy with it we should merge it as an example. However, IIUC, this approach should quickly be superceded by the `open_virtual_mfdataset` approach in #349. The difference is that you are using lithops to wrap xarray calls, where I am using xarray (/VirtualiZarr) calls to wrap lithops.

abarciauskas-bgse:
@TomNicholas I'm happy with it but would also be happy to try out `open_virtual_mfdataset` next time around.

TomNicholas (approved):

Issue #473, opened 2025-03-04
Re-implement `loadable_variables` using ManifestStore


TomNicholas:
https://github.com/zarr-developers/VirtualiZarr/issues/124 describes @ayushnag's idea to allow loading data variables directly from the in-memory `ManifestArrays`, rather than having to write to kerchunk/icechunk then reading from that. PR #458 will add a zarr-compliant in-memory virtual `ManifestStore` that wraps a virtual dataset and would allow loading data from it via 

```python
virtual_ds = vz.open_virtual_dataset(filepath)
manifeststore = vz.ManifestStore(virtual_ds)
lazy_ds = xr.open_zarr(manifeststore)
loaded_ds = ds.load()
```

This issue is to track the idea that once #458 is implemented we should refactor the implementation of `loadable_variables` to use this `ManifestStore` + `xr.open_zarr` approach internally, for all backends. Currently this is instead done by each virtual backend calling out to a different xarray backend, depending on the filetype.

There are multiple reasons to re-implement this:
1. We would no longer need every virtual backend to have a corresponding xarray backend,
2. We would be able to guarantee (and create property-based tests - see #394) that loading data via `loadable_variables` will give the same result as creating a virtual dataset, writing to icechunk, then loading,
3. Make it easier entralize file handle management, so we can close file handles in the way xarray can (see https://github.com/zarr-developers/VirtualiZarr/issues/468).

FYI @chuckwondo

maxrjones:
Completed by #601

PR #477, opened 2025-03-06
Change default loadable_variables (and indexes) to match xarray's behaviour

### LINKED ISSUE #335: Add loadable_variables='coords' option

TomNicholas:
This is a second attempt at addressing #335, being more brutal about removing options that aren't used. It is also intended to make implementing #473 easier.

The idea is that no-one really cares about all the complexity of distinguishing between 1D coordinate variables with and without indexes. Instead we should just default to the same index-creation behaviour as xarray uses, and the easiest way to do that is just to use `xr.open_dataset` and drop variables the user didn't actually want to load.

This will be inefficient right now (in the same way that the current implementation is inefficient) because we fully scan over the whole file twice. But this sets up for #473, which will avoid scanning over the file more than once.

- [x] Closes #335
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


TomNicholas:
I just removed `indexes={}` everywhere in the tests, but maybe that should actually be done in a follow-up PR...

TomNicholas:
FYI @maxrjones @sharkinsspatial this PR has got to the point where I think the only failing tests are those which use a kerchunk-based reader, as I haven't ported the kerchunk translation code yet. So you could maybe build off this branch already...

TomNicholas:
I'm very close now - when I run this locally the only failing tests left is for the old kerchunk HDF5 reader

TomNicholas:
To summarize what's actually been changed (from the release note):

```
  Which variables are loadable by default has changed. The behaviour is now to make loadable by default the
  same variables which `xarray.open_dataset` would create indexes for: i.e. one-dimensional coordinate variables whose
  name matches the name of their only dimension (also known as "dimension coordinates").
  Pandas indexes will also now be created by default for these loadable variables.
  This is intended to provide a more friendly default, as often you will want these small variables to be loaded
  (or "inlined", for efficiency of storage in icechunk/kerchunk), and you will also want to have in-memory indexes for these variables
  (to allow `xarray.combine_by_coords` to sort using them).
  The old behaviour is equivalent to passing ``loadable_variables=[]`` and ``indexes={}``.
```

TomNicholas:
Oh balls I made this PR against `main` not against `develop`

maxrjones:
> Oh balls I made this PR against `main` not against `develop`

I just resolved the merge conflicts with develop to take a look at the Kerchunk stuff you mentioned. Mind if I push it to your branch?

TomNicholas:
go for it!

maxrjones:
@TomNicholas regarding https://github.com/zarr-developers/VirtualiZarr/pull/477#issuecomment-2743822384, looking at https://github.com/TomNicholas/VirtualiZarr/blob/e2c3a79f5c127957d65420bfa66dd9b869f19d03/virtualizarr/writers/kerchunk.py#L154-L159, the Zarr V2 array metadata is in both `arr_refs[.zattrs]` and `arr_refs[.zarray]` where they should only be in `arr_refs[.zarray]`. I think you want to pop any Zarr array spec related metadata from the variable in `replace_virtual_with_loadable_vars()`.

maxrjones:
> @TomNicholas regarding [#477 (comment)](https://github.com/zarr-developers/VirtualiZarr/pull/477#issuecomment-2743822384), looking at [TomNicholas/VirtualiZarr@`e2c3a79`/virtualizarr/writers/kerchunk.py#L154-L159](https://github.com/TomNicholas/VirtualiZarr/blob/e2c3a79f5c127957d65420bfa66dd9b869f19d03/virtualizarr/writers/kerchunk.py#L154-L159), the Zarr V2 array metadata is in both `arr_refs[.zattrs]` and `arr_refs[.zarray]` where they should only be in `arr_refs[.zarray]`. I think you want to pop any Zarr array spec related metadata from the variable's .attrs in `replace_virtual_with_loadable_vars()`.

Actually, sorry nevermind this doesn't have anything to do with `replace_virtual_with_loadable_vars()`, the redundancy is due to adding `var.encoding` to the `arr_refs[".zattrs"]`

TomNicholas:
One thing I didn't really think about before changing this behaviour is that now that the default is to potentially load _some_ `loadable_variables`, readers which read metadata-only formats (i.e. dmrpp and kerchunk) will (after ) attempt to load data from the original source be default.

This is almost certainly not want the user wants to do, but either we:
1. Have all readers behave consistently wrt the `loadable_variables` kwarg, but force users of the kerchunk and dmrpp readers to always say `loadable_variables=[]` explicitly,
2. Have the dmrpp & kerchunk readers silently ignore the `loadable_variables` kwarg's intended default behaviour and do their own thing (i.e. not load anything),
3. Have the dmrpp & kerchunk readers ignore the `loadable_variables` kwarg's intended default behaviour but raise a warning when doing so.

I think I prefer 3.

cc @ayushnag @betolink

conftest.py
```diff
@@ -195,7 +195,7 @@ def netcdf4_virtual_dataset(netcdf4_file):
     """Create a virtual dataset from a NetCDF4 file."""
     from virtualizarr import open_virtual_dataset
 
-    with open_virtual_dataset(netcdf4_file, indexes={}) as ds:
+    with open_virtual_dataset(netcdf4_file, loadable_variables=[]) as ds:
```
TomNicholas, line 198: Required otherwise we get inlined variables in the kerchunk file which we don't know how to read (#489)

docs/usage.md
```diff
@@ -28,27 +28,27 @@ vds = open_virtual_dataset('air.nc')
 
 (Notice we did not have to explicitly indicate the file format, as {py:func}`open_virtual_dataset <virtualizarr.open_virtual_dataset>` will attempt to automatically infer it.)
 
-Printing this "virtual dataset" shows that although it is an instance of `xarray.Dataset`, unlike a typical xarray dataset, it does not contain numpy or dask arrays, but instead it wraps {py:class}`ManifestArray <virtualizarr.manifests.ManifestArray>` objects.
+Printing this "virtual dataset" shows that although it is an instance of `xarray.Dataset`, unlike a typical xarray dataset, instead of some numpy or dask arrays, instead it wraps {py:class}`ManifestArray <virtualizarr.manifests.ManifestArray>` objects.
```
TomNicholas, line 31:
```suggestion
Printing this "virtual dataset" shows that although it is an instance of `xarray.Dataset`, unlike a typical xarray dataset, in addition to a few in-memory numpy arrays, it also wraps {py:class}`ManifestArray <virtualizarr.manifests.ManifestArray>` objects.
```

docs/usage.md
```diff
@@ -230,7 +230,9 @@ But before we combine our data, we might want to consider loading some variables
 
 ## Loading variables
 
-Whilst the values of virtual variables (i.e. those backed by `ManifestArray` objects) cannot be loaded into memory, you do have the option of opening specific variables from the file as loadable lazy numpy/dask arrays, just like `xr.open_dataset` normally returns. These variables are specified using the `loadable_variables` argument:
+Whilst the values of virtual variables (i.e. those backed by `ManifestArray` objects) cannot be loaded into memory, you do have the option of opening specific variables from the file as loadable lazy numpy/dask arrays, just like `xr.open_dataset` normally returns.
```
TomNicholas, line 233:
```suggestion
Whilst the values of virtual variables (i.e. those backed by `ManifestArray` objects) cannot be loaded into memory, you do have the option of opening specific variables from the file as loadable lazy numpy arrays, just like `xr.open_dataset` normally returns.
```

docs/usage.md
```diff
@@ -261,18 +263,21 @@ Loading variables can be useful in a few scenarios:
...
+Whilst this does mean the original data will be duplicated in your new virtual zarr store, by loading your coordinates into memory they can be inlined in the reference file for fast reads from the virtual store.
+
+However, you should not do this for higher-dimensional variables, as then you might use a lot of storage duplicating them, defeating the point of the virtual zarr approach. Also, anything duplicated could become out of sync with the referenced origial files, especially if not using a transactional storage engine like `Icechunk`.
+```
 
 ### CF-encoded time variables
 
-To correctly decode time variables according to the CF conventions, you need to pass `time` to `loadable_variables` and ensure the `decode_times` argument of `open_virtual_dataset` is set to True (`decode_times` defaults to None).
+To decode time variables according to the CF conventions, you need ensure `time` is one of the `loadable_variables` and the `decode_times` argument of `open_virtual_dataset` is set to `True` (`decode_times` defaults to None).
```
TomNicholas, line 280:
```suggestion
To decode time variables according to the CF conventions, you must ensure `time` is one of the `loadable_variables` and the `decode_times` argument of `open_virtual_dataset` is set to `True` (`decode_times` defaults to None).
```

docs/usage.md
```diff
@@ -362,32 +367,30 @@ combined_vds['air'].data.manifest.dict()
 ```
 
 ```{note}
-The keyword arguments `coords='minimal', compat='override'` are currently necessary because the default behaviour of xarray will attempt to load coordinates in order to check their compatibility with one another. In future this [default will be changed](https://github.com/pydata/xarray/issues/8778), such that passing these two arguments explicitly will become unnecessary.
+If you have any virtual coordinate variables, you will likely need to specify the keyword arguments ``coords='minimal'`` and ``compat='override'`` to ``xarray.concat()``, because the default behaviour of xarray will attempt to load coordinates in order to check their compatibility with one another. In future this [default will be changed](https://github.com/pydata/xarray/issues/8778), such that passing these two arguments explicitly will become unnecessary.
```
TomNicholas, line 370:
```suggestion
If you have any virtual coordinate variables, you will likely need to specify the keyword arguments `coords='minimal'` and `compat='override'` to `xarray.concat()`, because the default behaviour of xarray will attempt to load coordinates in order to check their compatibility with one another. In future this [default will be changed](https://github.com/pydata/xarray/issues/8778), such that passing these two arguments explicitly will become unnecessary.
```

virtualizarr/readers/api.py
TomNicholas, line 1: Oh I also moved this to a new `api.py` file.

virtualizarr/backend.py
```diff
@@ -159,10 +161,10 @@ def open_virtual_dataset(
...
         )
 
-    drop_variables, loadable_variables = check_for_collisions(
-        drop_variables,
-        loadable_variables,
-    )
+    # drop_variables, loadable_variables = check_for_collisions(
+    #     drop_variables,
+    #     loadable_variables,
+    # )
```
TomNicholas, line 167:
```suggestion
```

virtualizarr/readers/common.py
```diff
@@ -1,121 +1,107 @@
...
-            name: index
-            for name, index in ds.xindexes.items()
-            if name in loadable_variables
-        }
-    elif indexes != {}:
-        # TODO allow manual specification of index objects
-        raise NotImplementedError()
+    var_names_to_load: list[Hashable]
+    if isinstance(loadable_variables, list):
+        var_names_to_load = list(loadable_variables)
```
maxrjones, line 65: why do you need to cast this to a list if the code is only run on list instances?

virtualizarr/readers/common.py
```diff
@@ -1,121 +1,107 @@
...
-    # TODO See issue #124 for a suggestion of how to avoid calling xarray here.
+    return vds
+
+
+# TODO reimplement this using ManifestStore (GH #473)
+def replace_virtual_with_loadable_vars(
+    fully_virtual_dataset: xr.Dataset,
+    filepath: str,  # TODO won't need this after #473 because the filepaths will be in the ManifestStore
+    group: str | None = None,
+    loadable_variables: Iterable[Hashable] | None = None,
```
maxrjones, line 44: Why is the type `Iterable[Hashable]` here rather than `Iterable[str]` like elsewhere?

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -528,6 +528,7 @@ def test_append_virtual_ref_with_encoding(
             xrt.assert_equal(new_ds, expected_ds)
 
     ## When appending to a virtual ref with encoding, it succeeds
+    @pytest.mark.xfail
```
maxrjones, line 531: Can you add a reason for this xfail marker?

maxrjones (approved):
Looks good, thank you @TomNicholas! Just had a few nits

virtualizarr/readers/common.py
TomNicholas: Because I made the input type an `Iterable`, not a `list`

virtualizarr/readers/common.py
TomNicholas: Because technically xarray variable names can be `Hashable`. It should probably be generalized elsewhere too. I think I changed it here just to get rid of a mypy error.

virtualizarr/tests/test_writers/test_icechunk.py
TomNicholas: Apparently there was no reason for it - removed in https://github.com/zarr-developers/VirtualiZarr/pull/477/commits/6bf2615c1d80b30a392c1a5c9d742c3a69f6d6f5

PR #479, opened 2025-03-06
Manage environments with uv

### LINKED ISSUE #310: Use uv / pixi?

maxrjones:
Alternative to https://github.com/zarr-developers/VirtualiZarr/pull/407 that uses [uv](https://docs.astral.sh/uv/)

Closes #310 

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


virtualizarr/tests/test_readers/conftest.py
```diff
@@ -222,7 +222,7 @@ def skip_test_for_libhdf5_version():
     return libhdf5_version < Version("1.14")
 
 
-@pytest.fixture(params=["blosc_zlib", ""])
+@pytest.fixture(params=["zlib", ""])
```
maxrjones, line 225: This is because the netcdf4 wheels don't include plugins (xref https://github.com/Unidata/netcdf4-python/issues/1164). I'm not sure how important it is to use blosc specifically for this test.

TomNicholas:
> as long as all of the system-level dependencies distribute binary wheels, then pixi is not necessary?

I think that's correct.

> And at the risk (which I am happy to assume) of appearing naive, is uv doing something special to ensure the compatibility here, or does the advantage of uv over pip become simply performance?

I don't think `uv` is doing anything special. 

> I will admit I am surprised, but in a curious/intruiged way, to learn that this project would never need access to conda-style environment resolution, and would love to learn more!

IIUC then _users_ of this package might need conda-style environment resolution if they want to use (virtuali)zarr with some compressor that isn't distributed via wheels, but _developers_ of this package don't really need to worry about those cases, so long as we continue to distribute this package via conda-forge too.

maxrjones:
Based on the latest comments in https://github.com/zarr-developers/VirtualiZarr/issues/310, I'm going to close this PR in favor of #407

PR #480, opened 2025-03-07
Make virtual chunk staleness check opt out

### LINKED ISSUE #436: Virtual chunk staleness check should be opt-out, not opt-in
The `last_updated_at` argument to `.to_icechunk()` is optional, with the default being not to set any checksum in icechunk. This means that when reading virtual chunks from the icechunk store checking that the chunks are not stale is currently opt-in. 

It should be opt-out instead, as this is much safer, and there is arguably not a lot of reason to even turn this feature off. It could be implemented simply by changing the default behaviour to use the current time as the timestamp. 

cc @mpiannucci

TomNicholas:
- [x] Closes #436
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


virtualizarr/writers/icechunk.py
```diff
@@ -329,13 +329,17 @@ def write_manifest_virtual_refs(
         op_flags=[["readonly"]] * 3,  # type: ignore
     )
 
+    # Icechunk checksums currently store with second precision, so we need to make sure
+    # the checksum_date is at least one second in the future
+    timestamp_checksum = datetime.now(timezone.utc) + timedelta(seconds=1)
+
```
TomNicholas, line 335: @mpiannucci I feel like this +1 second business should be in the test, rather than in the business logic?

mpiannucci: Yes I agree

TomNicholas: We actually came to the opposite conclusion on a call just now. The long comments in the code hopefully explain our reasoning.

Issue #84, opened 2024-04-16
Support for groups


TomNicholas:
We should support generating references from files containing multiple groups in the same way that `xr.open_dataset` and `datatree.open_datatree` work. 

So we should add a new `open_virtual_datatree` function, and a new (optional) `group` kwarg to `open_virtual_dataset`.

This can be done right now using the datatree package (as an optional dependency imported inside `open_virtual_datatree`) but once that gets merged into xarray `main` (which is happening right now) we can get rid of that dependency.

See https://github.com/TomNicholas/VirtualiZarr/issues/78#issuecomment-2059737479 and #11.

cc @sharkinsspatial

TomNicholas:
One thing I realized about this is that concatenating multiple `DataTree` objects is currently a little awkward. I don't know if this is actually a common pattern, but imagine you had two netCDF files each with groups, and you wanted to concatenate group1 in file1 with group1 in file2 etc.

Adding `open_virtual_datatree` would allow you to open the files like this:
```python
vdt1 = open_virtual_datatree('file1.nc')
vdt1 = open_virtual_datatree('file2.nc')
```
but currently you can't do 
```python
combined_vdt = xr.concat([vdt1, vdt1], dim='time')
```
because `xr.concat` doesn't understand `DataTree` objects. To get around this you should be able to do

```python
from datatree import map_over_subtree

concat_datatrees = map_over_subtree(xr.concat)

combined_vdt = concat_datatrees([vdt1, vdt1], dim='time')
```

but it raises the question of whether the xarray DataTree upstream integration should include generalizing `xr.concat`.

jonas-spaeth:
Hi @TomNicholas

I often have the problem that I want to concat different datatrees into a single xr.Dataset again.
I came across your code above and tried it, but I get an error.

Generate some sample data:
```python
import datatree
import xarray as xr

ds1 = xr.Dataset(
    data_vars=dict(a=("x", [11, 22, 33])),
    coords=dict(x=[1,2,3])
)
ds2 = xr.Dataset(
    data_vars=dict(a=("x", [111, 222, 333])),
    coords=dict(x=[1,2,3])
)

mytree = datatree.DataTree.from_dict({"two_digits": ds1, "three_digits": ds2})
print(mytree)
```
output:
```
DataTree('None', parent=None)
├── DataTree('two_digits')
│       Dimensions:  (x: 3)
│       Coordinates:
│         * x        (x) int64 24B 1 2 3
[...]

PR #488, opened 2025-03-13
WIP: Refactor TIFF backend to use async_tiff and obstore

### LINKED ISSUE #291: open_virtual_dataset fails to open tiffs

maxrjones:
This PR does a few things, which may need to get split up but are convenient to prototype together:
- Refactor TIFFVirtualBackend to use async_tiff (closes #291) (would supersede #295, #297, #292)
- Adds a VirtualDataArrayAccessor and top-level `open_virtual_dataarray` for single IFD TIFF files (related to #84)
- Adds a `VirtualObjectStore` for testing functionality without needing to serialize to kerchunk or icechunk (relates to #124)

There's still a lot to do here, but I'm opening the PR early to make it easier to ask people questions.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


TomNicholas:
This is awesome!

I would ideally like to see each of your bullets be a separate PR. But I can see it might be easiest to develop them all here then split and merge separately.

virtualizarr/tests/data/LC08_L2SP_046027_20201229_20210310_02_T2_SR_B2.TIF
TomNicholas, line 1: Let's not commit an 80MB file to the repo. The 30KB one is okay, but ideally we would only commit artificially-created data that is literally as small as possible.

maxrjones: Sounds good, I'm remove it before finishing the PR. The 30 KB one is stripped rather than tiled and it was easier to implement tiled TIFFs first since I find it more intuitive to work with, which is why I added it for now.

https://github.com/zarr-developers/VirtualiZarr/issues/347 for TIFF along with a tiny in-repo test would be the best approach, but I don't want to get too distracted by other tempting issues 😅

kylebarron: FWIW I try to avoid ever committing a large file to the repo, because if you add it and remove it a few commits later, the entire file will still be in the Git history, so the repo will still be 80MB larger. I _think_ if you squash the merge you'll be fine though?

virtualizarr/accessor.py
```diff
@@ -31,7 +58,7 @@ def to_icechunk(
         append_dim: str | None = None,
         last_updated_at: datetime | None = None,
     ) -> None:
-        """
+        """o
```
kylebarron, line 61: typo?

virtualizarr/readers/tiff.py
```diff
@@ -1,20 +1,48 @@
...
+
+import numcodecs.registry as registry
+from async_tiff import TIFF
+
+from virtualizarr.codecs import numcodec_config_to_configurable
+from virtualizarr.manifests import ChunkManifest, ManifestArray
+from virtualizarr.manifests.utils import create_v3_array_metadata
+
+if TYPE_CHECKING:
+    from async_tiff._ifd import ImageFileDirectory
```
kylebarron, line 15: I think this should be exported from the top level now?
```suggestion
    from async_tiff import ImageFileDirectory
```

virtualizarr/readers/tiff.py
```diff
@@ -1,82 +1,197 @@
...
+            )
+        else:
+            for ind, ifd in enumerate(tiff.ifds):
+                manifest_dict[str(ind)] = TIFFVirtualBackend._construct_manifest_array(
+                    ifd=ifd, path=path
+                )
+        return ManifestGroup(manifest_dict=manifest_dict, attributes=attrs)
+
+    @staticmethod
+    def open_virtual_zarr(
```
TomNicholas, line 139: Let's call this something else, perhaps this?
```suggestion
    def create_manifeststore(
```

virtualizarr/readers/tiff.py
```diff
@@ -1,82 +1,197 @@
...
 
-        return construct_virtual_dataset(
-            virtual_vars=virtual_vars,
-            loadable_vars=loadable_vars,
-            indexes=indexes,
-            coord_names=coord_names,
-            attrs=attrs,
+        # Convert to a Zarr store
+        return xr.open_dataset(
+            manifest_store, engine="zarr", consolidated=False, zarr_format=3
```
TomNicholas, line 196: I'm confused - isn't this going to create a dataset with all loadable variables???

maxrjones: All the variables are "loadable", but not loaded unless you call something like .data on them.

virtualizarr/tests/data/LC08_L2SP_046027_20201229_20210310_02_T2_SR_B2.TIF
kylebarron: I think even if you used a fork, but then merged via squash or with a merge commit, it would include all the intermediate commits, which would include the large file in the history.

virtualizarr/readers/tiff.py
TomNicholas: > All the variables are "loadable", but not loaded unless you call something like .data on them.

That's what I mean - isn't this going to create a `Dataset` fully of lazy numpy arrays instead of `ManifestArrays`?

maxrjones: True, the `ManifestArray`s are be stored in `ManifestStore._manifest_group._manifest_dict` rather than in each `DataArray`. I'll need to give this some thought, and probably have a discussion about how to approach this. We could have a final step that replaces the lazy numpy arrays with ManifestArrays for non-loadable variables but that would require a lot more spaghetti in the `ManifestStore` internals.

maxrjones: Putting aside how majorly disruptive this would be, why must the data variables contain ManifestArrays, rather than indexing the ManifestArrays from the Zarr Store instance at write time? I'm thinking about whether we could use this approach, call `.load()` on loadable variables, then use isinstance(var, numpy.ndarray) and isinstance(var, xarray.core.indexing.MemoryCachedArray) after the concatenation, merging, etc. to determine whether the data or the VirtualChunkSpec should be stored in Icechunk.

If that's possible, we could provide all the functionality of both the Zarr and Xarray APIs. e.g., to_zarr now could be used to create a "native" zarr store from any archival file format, or virtualize.to_icechunk could create a virtual zarr store with some inlined data. if we are really concerned about accidental data loading we could break the Zarr API a bit to prevent loading any data not included in 'loadable_variables'.

TomNicholas: > why must the data variables contain `ManifestArrays`, rather than indexing the `ManifestArrays` from the Zarr Store instance at write time?

Because you can't do arbitrary transformations on `ManifestArrays` like you can on numpy arrays. IIUC your approach (which I think can be summarized as "lazily fetching the `ManifestArray` corresponding to this numpy array) would very often lead to people unable to write virtual variables to the store at write time, because of some step they did earlier.

maxrjones: I'm going to experiment a bit. Because if we have a `loadable_arrays` attribute on the ManifestStore, we can raise an error if Xarray calls `.get()` on any of the arrays that aren't set as loadable, preventing any eager computations that would prevent serialization. So, it wouldn't be a full feature complete dataset but you'd get better error messages than those relating to ChunkManagers. Still may be a bad idea, but I think there's no harm in spending an afternoon trying it out.

TomNicholas: My intuition is that this is just going to un-separate the concerns we want to separate, but I support messing around 😀

maxrjones:
I'm going to close this and rewrite the implementation after #477 and #490 are merged, which should make the PR much simpler. Hoping that the new PR will be ready for review by the middle of next week.

PR #458, opened 2025-03-02
Manifest store

TomNicholas:
Supercedes #427, as @ayushnag has made a great start but says he doesn't have much time to work on it in future. I couldn't get git to push more commits to his PR so I've just made this new one.

- [x] Closes #124
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage
- [ ] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`
- [ ] New functionality has documentation

Issue #16, opened 2024-03-08
Are manifests with missing chunks valid?


TomNicholas:

TomNicholas:
Apparently they are valid in the sense that zarr will just create a chunk full of `fill_value` if a chunk is not found in the store, see https://github.com/TomNicholas/VirtualiZarr/issues/32#issuecomment-1997467616.

PR #490, opened 2025-03-16
Add ManifestStore for loading data from ManifestArrays

### LINKED ISSUE #124: Loading data from ManifestArrays without saving references to disk first

maxrjones:
This PR allows loading data directly from ManifestArrays using [obstore](https://developmentseed.org/obstore/latest/). 
Possible TO-DOs:

- [ ] Implement `get_partial_values`, `exists`, `list_prefix`, and `list`.

General checklist
- [x] Closes #124 (one option, doesn't preclude a different FSSpec-backed VirtualStore)
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


virtualizarr/storage/_obstore.py
```diff
@@ -0,0 +1,246 @@
...
+
+_ALLOWED_EXCEPTIONS: tuple[type[Exception], ...] = (
+    FileNotFoundError,
+    IsADirectoryError,
+    NotADirectoryError,
+)
+
+
+class VirtualObjectStore(Store):
+    """A Zarr store that uses obstore for fast read/write from AWS, GCP, Azure.
```
TomNicholas, line 44: The fact this uses `obstore` is an implementation detail, as it could for example later be changed to use `fsspec` with no user-facing changes (I'm not actually suggesting we do that). 

What makes this store unique is the fact is uses a manifest, which is why I wanted to call it `ManifestStore` (though `VirtualStore` is okay too).

virtualizarr/storage/_obstore.py
```diff
@@ -0,0 +1,246 @@
...
+    get_zarr_metadata,
+    list_dir_from_xr_obj,
+    parse_manifest_index,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncGenerator, Iterable
+    from typing import Any
+
+    from obstore.store import ObjectStore as _UpstreamObjectStore
```
TomNicholas, line 29: What's the purpose of this funny type hint?

virtualizarr/storage/_obstore.py
```diff
@@ -0,0 +1,246 @@
...
+                    obs.store.HTTPStore,
+                    obs.store.S3Store,
+                    obs.store.LocalStore,
+                    obs.store.MemoryStore,
+                ),
+            ):
+                raise TypeError(f"expected ObjectStore class, got {store!r}")
+        super().__init__(read_only=True)
+        self.stores = stores
+        self.xr_obj: DataArray | Dataset = xr_obj
```
TomNicholas, line 87: You can add type hints for these in the class definition, outside of any of the methods.

TomNicholas, line 87: There should be `isinstance` checks here.

virtualizarr/storage/_obstore.py
```diff
@@ -0,0 +1,246 @@
...
+        key: str,
+        prototype: BufferPrototype,
+        byte_range: ByteRequest | None = None,
+    ) -> Buffer | None:
+        # docstring inherited
+        import obstore as obs
+
+        if "zarr.json" in key:
+            return get_zarr_metadata(self.xr_obj, key)
+        manifest_index = parse_manifest_index(key)
```
TomNicholas, line 118: I don't think I really understand the need for this `ManifestIndex` class. All it saves you is returning two separate variables in this line (`var_name` & `chunk_key`).

virtualizarr/storage/_obstore.py
```diff
@@ -0,0 +1,246 @@
...
+        store_request = find_matching_store(stores=self.stores, request_key=path)
+        # Transform the input byte range to account for the chunk location in the file
+        chunk_end_exclusive = offset + length
+        byte_range = _transform_byte_range(
+            byte_range, chunk_start=offset, chunk_end_exclusive=chunk_end_exclusive
+        )
+        # Actually get the bytes
+        try:
+            bytes = await obs.get_range_async(
+                self.stores[store_request.store_id],
```
TomNicholas, line 143: Perhaps better to pull this selection of store out of the `try...except...`?

virtualizarr/storage/_obstore.py
```diff
@@ -0,0 +1,246 @@
...
+                store_request.key,
+                start=byte_range.start,
+                end=byte_range.end,
+            )
+            return prototype.buffer.from_bytes(bytes)  # type: ignore[arg-type]
+        except _ALLOWED_EXCEPTIONS:
+            return None
```
TomNicholas, line 150: Why are any of these exceptions allowed? Don't we want to raise them?

virtualizarr/tests/__init__.py
```diff
@@ -39,6 +39,7 @@ def _importorskip(
 has_imagecodecs, requires_imagecodecs = _importorskip("imagecodecs")
 has_hdf5plugin, requires_hdf5plugin = _importorskip("hdf5plugin")
 has_zarr_python, requires_zarr_python = _importorskip("zarr")
+has_obstore, requires_obstore = _importorskip("obstore")
```
TomNicholas, line 42: To support #353 and #473 `obstore` would need to become a required dependency

virtualizarr/tests/test_integration.py
```diff
@@ -126,6 +127,15 @@ def roundtrip_as_in_memory_icechunk(vds: xr.Dataset, tmpdir, **kwargs):
     return xr.open_zarr(session.store, zarr_format=3, consolidated=False, **kwargs)
 
 
+def roundtrip_as_virtual_object_store(vds: xr.Dataset, tmpdir, **kwargs):
+    from obstore.store import LocalStore
+
+    store = LocalStore()
+    stores = {"file://": store}
+    # Convert dataset to a be obstore backed for loading
+    return vds.virtualize.to_xarray(stores)
```
TomNicholas, line 136: This new `storage` module deserves it's own test module, where you directly test the `ManifestStore` interface behaves as expected. Otherwise we're going to end up with every test in `test_integration.py`. 

I also heard there is a testing scaffold in zarr-python that we could use to help with testing the `Store` interface is correct too.

virtualizarr/accessor.py
```diff
@@ -88,6 +94,43 @@ def to_icechunk(
...
+        --------
+        The to_xarray() functionality is experimental.
+        """
+        store = VirtualObjectStore(self.ds, stores)
+        return open_dataset(
+            store,  # type: ignore[arg-type]
+            engine="zarr",
+            backend_kwargs={"zarr_format": 3, "consolidated": False},
+            **kwargs,
+        )
```
TomNicholas, line 132: I think adding this `.virtualize.to_xarray` accessor method is clever, but possibly premature. I'm not entirely sure what the best API would be, but I think in the meantime we should simply expose `VirtualObjectStore`/`ManifestStore` as public (developer) API, as I sketched out in #458.

virtualizarr/storage/_obstore.py
```diff
@@ -0,0 +1,246 @@
...
+    -----
+    Modified from https://github.com/zarr-developers/zarr-python/pull/1661
+    """
+
+    def __eq__(self, value: object):
+        NotImplementedError
+
+    def __init__(
+        self,
+        xr_obj: DataArray | Dataset,
```
TomNicholas, line 66: What happens if I pass a Dataset containing some numpy arrays?

TomNicholas (commented):
This is _awesome_ @maxrjones - I'm so impressed that you got this far this quickly. My comments are mostly nits about API choices.

virtualizarr/storage/_obstore.py
TomNicholas, line 1: I don't think this filename needs a preceding underscore.

virtualizarr/accessor.py
maxrjones: I find this super convenient. I made it private for the time being and can remove it later in this PR if you feel strongly about making the user need to create the store and pass it to open_dataset with the necessary kwargs.

virtualizarr/storage/_obstore.py
maxrjones: Currently `AttributeError: 'numpy.ndarray' object has no attribute 'metadata'` but I think passing a Dataset containing some numpy arrays is equivalent to inlining all variables, so I will be sure to test for that and ideally support it.

virtualizarr/tests/__init__.py
maxrjones: I think I'd rather it become required at that point rather than in this PR, but it's not a strong preference.

virtualizarr/storage/_obstore.py
maxrjones: This is how Zarr handles empty chunks (xref https://github.com/zarr-developers/VirtualiZarr/issues/16). I'll need to put some more thought into this and return with a more complete answer later on.

maxrjones:
Thanks for prompt review, @TomNicholas! I addressed the simpler comments and will build out the test suite and request a follow-up once that's done and all the exposed issues are solved.

virtualizarr/storage/obstore.py
```diff
@@ -0,0 +1,247 @@
...
+__all__ = ["VirtualStore"]
+
+_ALLOWED_EXCEPTIONS: tuple[type[Exception], ...] = (
+    FileNotFoundError,
+    IsADirectoryError,
+    NotADirectoryError,
+)
+
+
+class VirtualStore(Store):
```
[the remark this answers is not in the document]
maxrjones: I thought that `VirtualStore` makes more sense if it's part of the public API, since users don't generally interact much with `ChunkManifest` or even `ManifestArray` and so the `Manifest` terminology might be unfamiliar.

But I would rather the public API be through a function (e.g., `ds.virtualize.to_xarray(stores)` or `vds_to_xarray(ds, stores)`) rather than creating a class instance and passing it to the xarray API, in which case `ManifestStore` as a core (non-public) class would be perfect.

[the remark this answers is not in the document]
TomNicholas: > I would rather the public API be through a function (e.g., `ds.virtualize.to_xarray(stores)` or `vds_to_xarray(ds, stores))` rather than creating a class instance and passing it to the xarray API

Why? If we approach this as "what is the minimum amount of new API we can add to make this usable?" then that clearly is to add the store but not the accessor method.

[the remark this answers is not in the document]
maxrjones: > > I would rather the public API be through a function (e.g., ds.virtualize.to_xarray(stores) or vds_to_xarray(ds, stores)) rather than creating a class instance and passing it to the xarray API

> Why? If we approach this as "what is the minimum amount of new API we can add to make this usable?" then that clearly is to add the store but not the accessor method.

Because writing `zarr_format=3, consolidated=False` gets tedious and reducing two lines of code to one can actually dramatically effect ergonomics based on reactions to the "kerchunk" xarray backend.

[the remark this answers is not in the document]
TomNicholas: > writing `zarr_format=3`, `consolidated=False` gets tedious

We can get rid of `consolidated=False` with a change to xarray (see https://github.com/pydata/xarray/issues/10122), and I isn't `zarr_format=3` auto-detected anyway?

> reducing two lines of code to one

To me this is not a good enough reason to increase the API surface. Especially when this package is already confusing in terms of "is this xarray? zarr? virtual what?". IMO good API design involves the minimal surface area, chosen carefully. Also it's easier to add more surface later than it is to remove it.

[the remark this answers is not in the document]
TomNicholas: > based on reactions to the "kerchunk" xarray backend.

This is not an analogous use case anyway though - our equivalent of that is `xr.open_zarr(icechunkstore)`. Kerchunk doesn't even have an equivalent of the way we can turn virtual variables into loaded variables after opening the file.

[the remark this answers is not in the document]
maxrjones: The sticking point for me is exposing `ManifestStore` as part of the public API because we could provide a simpler, more readable API. If we're talking only about the private API then I'm fine going with your preferences here.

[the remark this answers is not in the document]
TomNicholas: tl;dr: I'm suggesting that `ManifestStore` is _public developer_ API, for authors of new backends, not _public user_ API, for users using virtualizarr on formats it already understands. We have more to discuss about the public API (below).

Fundamentally the actual abstraction is `ManifestStore`. The accessor method is merely indirection around that abstraction. I would like to expose `ManifestStore` publicly as part of developer API because I think it's a powerful and useful abstraction that might have uses beyond what we've anticipated here.

> simpler, more readable API.

I am also worried that calling `.to_xarray` on something that's already an `xarray.Dataset` is really unintuitive. It seems much more intuitive to me to call this method something like `.load()`.

---

Consider that this PR now gives us two ways to achieve the same thing (get an entirely in-memory xarray dataset from a filepath). The first is of course

```python
ds = xr.open_dataset(file)
ds = ds.load()
```

I would like our API to feel intuitive in relation to this pre-existing pattern.

Currently our API to achieve the same thing looks like this:

```python
ds = xr.open_virtual_dataset(file, loadable_variables=<list of all the variables>)
```
which is a little wonky, but made sense at the time.

Your suggestion would add a _second_ user-facing (i.e. not just labeled as for developers) way to achieve this:

```python
vds = xr.open_virtual_dataset(file)
ds = vds.virtualize.to_xarray()
```

whereas my suggestion of adding only `ManifestStore` to the developer API doesn't create two public user-facing ways to achieve the same thing. (It only does for developers, who should be given access to the real abstractions anyway.)

I think the _most_ intuitive API for loading variables would arguably be something like:

```python
vds = xr.open_virtual_dataset(file)
ds = vds.load()
```

and we may want to eventually replace `loadable_variables` with that. Then we have the best of all worlds: one and only one way for _users_ to load variables, that is exactly what they would intuitively expect it to be. But also we show developers and only developers how to access the actual abstraction that makes it work.

But adding `ManifestStore` and just `ManifestStore` to the developer API now allows us to defer this decision about `loadable_variables`/`.load()` until later, without adding more things that we would then need to deprecate.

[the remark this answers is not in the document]
TomNicholas: Sorry for the rant hahaha 😅 I think this might be interesting regardless though - it hadn't occurred to me until now that we could actually remove `loadable_variables`...

virtualizarr/storage/_obstore.py
TomNicholas: Have you read this? https://github.com/zarr-developers/VirtualiZarr/pull/427#discussion_r1943767851

Seems related.

virtualizarr/storage/_obstore.py
TomNicholas: I think we need to think more about this. 

Firstly for now any numpy arrays should definitely be caught by an explicit `instance` check at initialization time.

Secondly it's not clear to me which is more complicated:

1. Having `ManifestStore` wrap things that are not `ManifestArrays`, which creates a split in which arrays need to be read from one of `obstore`'s stores (you could actually handle it by putting the numpy arrays in a `MemoryStore`).
2. Having `ManifestStore` only wrap `ManifestArrays`, but as a result all readers need to split out only the virtual variables when creating the `ManifestStore` to load variables from.

Option 1 sounds more complicated to me...

TomNicholas: Actually here's an argument for forbidding non-`ManifestArrays` from `ManifestStore`. 

In #477 I'm refactoring all readers to work like this:
1. Create `ManifestArrays` for all variables in the given file,
2. Load only the subset of those arrays specified by `loadable_variables`.

This refactor is simpler, and it has the important property that it's impossible to even create a loaded variable in the `vds` that cannot be serialized back to zarr. (Our current system could easily use the other codepath that uses `xr.open_dataset(filepath)` to create a loaded variable directly, but one that we only discover cannot be represented in zarr at store write time (e.g. because it has an unsupported dtype).)

But if every reader works like this, none of them need a `ManifestStore` that contains anything but `ManifestArrays`.

virtualizarr/storage/obstore.py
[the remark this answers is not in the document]
maxrjones: Let's return to the `.load()` idea once the more immediate missing features and tests are completed, but thanks for thoroughly explaining your reasoning. I'm on board.

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,364 @@
...
+    async def get(
+        self,
+        key: str,
+        prototype: BufferPrototype,
+        byte_range: ByteRequest | None = None,
+    ) -> Buffer | None:
+        # docstring inherited
+        import obstore as obs
+
+        if "zarr.json" in key:
```
TomNicholas, line 247: ?
```suggestion
        if key.endswith("zarr.json"):
```

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+async def list_dir_from_manifest_dict(
+    manifest_arrays: ManifestDict, prefix: str
+) -> AsyncGenerator[str]:
+    """Create the expected results for Zarr's `store.list_dir()` from an Xarray DataArrray or Dataset
+
+    Parameters
+    ----------
+    vd : xarray DataArray or Dataset
```
TomNicholas, line 67: outdated

virtualizarr/manifests/group.py
```diff
@@ -0,0 +1,36 @@
+from typing import TypeAlias
+
+from zarr.core.group import GroupMetadata
+
+from virtualizarr.manifests import ManifestArray
+
+ManifestDict: TypeAlias = dict[str, ManifestArray]
```
TomNicholas, line 7: Rename to something that includes the word array? This makes it sound like it only contains the manifest itself

virtualizarr/manifests/group.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class ManifestGroup:
+    """
+    Virtualized representation of multiple ManifestArrays as a Zarr Group.
+    """
+
+    _manifest_dict: ManifestDict
```
TomNicholas, line 15: instead of wrapping a dict, couldn't the `ManifestGroup` _be_ dict (or at least be a `Mapping`). That would allow one fewer layers of `._manifest` later

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,364 @@
...
+    ) -> Buffer | None:
+        # docstring inherited
+        import obstore as obs
+
+        if "zarr.json" in key:
+            return get_zarr_metadata(self._manifest_group, key)
+        var, chunk_key = parse_manifest_index(key)
+        path = self._manifest_group._manifest_dict[var]._manifest._paths[*chunk_key]
+        offset = self._manifest_group._manifest_dict[var]._manifest._offsets[*chunk_key]
+        length = self._manifest_group._manifest_dict[var]._manifest._lengths[*chunk_key]
```
TomNicholas, line 252: Let's split this up for readability?

```suggestion
        marr = self._manifest_group._manifest_dict[var]
        manifest = marr._manifest

        path = manifest._paths[*chunk_key]
        offset = manifest._offsets[*chunk_key]
        length = manifest._lengths[*chunk_key]
```

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,364 @@
...
+    Warnings
+    --------
+    ManifestStore is experimental and subject to API changes without notice. Please
+    raise an issue with any comments/concerns about the store.
+
+    Notes
+    -----
+    Modified from https://github.com/zarr-developers/zarr-python/pull/1661
+    """
+
```
TomNicholas, line 186: Shouldn't we explicitly make the store read-only here somewhere?

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,364 @@
...
+            yield k
+
+
+def _transform_byte_range(
+    byte_range: ByteRequest | None, *, chunk_start: int, chunk_end_exclusive: int
+) -> RangeByteRequest:
+    """
+    Convert an incoming byte_range which assumes one chunk per file to a
+    virtual byte range that accounts for the location of a chunk within a file.
+    """
```
TomNicholas, line 343: this is really neat

virtualizarr/tests/test_manifests/test_group.py
```diff
@@ -0,0 +1,36 @@
...
+        "0.1.1": {"path": "s3://bucket/foo.nc", "offset": 400, "length": 100},
+    }
+    manifest = ChunkManifest(entries=chunk_dict)
+    chunks = (5, 1, 10)
+    shape = (5, 2, 20)
+    array_metadata = array_v3_metadata(shape=shape, chunks=chunks)
+    return ManifestArray(metadata=array_metadata, chunkmanifest=manifest)
+
+
+class TestManifestgGroup:
```
TomNicholas, line 22:
```suggestion
class TestManifestGroup:
```

virtualizarr/manifests/group.py
maxrjones: The manifest group class allows adding group level attributes in the group metadata (e.g., https://github.com/zarr-developers/VirtualiZarr/blob/5231800d94ab120c5bbf75c6a6f520993d63e2a3/virtualizarr/tests/test_manifests/test_store.py#L55-L57), which would map to dataset / datatree node attrs in the xarray model. IIUC the tradeoff for one less `._manifest` later would be less readable code if we have a `_zarr_group_metadata` key or something like that, or is there a different way to support the group level attributes?

virtualizarr/manifests/store.py
maxrjones: There's this line to make the store read-only - https://github.com/zarr-developers/VirtualiZarr/blob/5231800d94ab120c5bbf75c6a6f520993d63e2a3/virtualizarr/manifests/store.py#L216

Do you think something more is needed?

TomNicholas: Oh cool. That works. We also should document that the store is read-only more clearly (e.g. in the top-level class docstring).

pyproject.toml
```diff
@@ -38,7 +38,9 @@ remote = [
     "aiohttp",
     "s3fs",
 ]
-
+obstore = [
+    "obstore @ git+https://github.com/developmentseed/obstore@main#subdirectory=obstore",
+]
```
TomNicholas, line 43: The `ObjectStore` implementation was just merged into zarr-python! Does that make any difference for us here? Or  should we continue to use the `obstore` package directly?

maxrjones: @kylebarron would you be willing to take a quick look at https://github.com/zarr-developers/VirtualiZarr/blob/28255e7ddf1c39a55d94e792540a48d944c115c7/virtualizarr/manifests/store.py#L169-L337 and provide your opinion on whether I should be subclassing zarr's ObjectStore rather than building this separately? I'm leaning towards this being distinct enough to warrant its own implementation, but would appreciate a second opinion.

kylebarron: > The `ObjectStore` implementation was just merged into zarr-python! Does that make any difference for us here? Or should we continue to use the `obstore` package directly?

You should definitely use `obstore` from PyPI instead of from a git hash, regardless of whether you subclass the zarr store or not.

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,367 @@
...
+        except _ALLOWED_EXCEPTIONS:
+            return None
+
+    async def get_partial_values(
+        self,
+        prototype: BufferPrototype,
+        key_ranges: Iterable[tuple[str, ByteRequest | None]],
+    ) -> list[Buffer | None]:
+        # docstring inherited
+        raise NotImplementedError
```
kylebarron, line 282: If you plan to implement this in the future, you could even consider importing the private functions from the upstream zarr obstore file.

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,367 @@
...
+
+    def __eq__(self, value: object):
+        NotImplementedError
+
+    def __init__(
+        self,
+        manifest_group: ManifestGroup,
+        *,
+        stores: StoreDict,
+    ) -> None:
```
kylebarron, line 198: A docstring here would be helpful to explain that I think this `stores` argument works by prefix?

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,367 @@
...
+        f"Expected the one of stores.keys() to match the data prefix, got {stores.keys()} and {request_key}"
+    )
+
+
+class ManifestStore(Store):
+    """A Zarr store that uses obstore for fast read/write from AWS, GCP, Azure.
+
+    Parameters
+    ----------
+    stores : dict[prefix, obstore.store.ObjectStore]
```
kylebarron, line 174: fwiw you can use [sphinx interlinking](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) here to automatically link to the obstore docs. As in [zarr-developers/zarr-python@`a3afa44` (#1661)](https://github.com/zarr-developers/zarr-python/pull/1661/commits/a3afa44ad39da8646e48c7b549925b8cb88115ee) you just need to point sphinx to:

```
https://developmentseed.org/obstore/latest/objects.inv
```

Or we could prototype mkdocs-material here if you wanted 🙈

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,367 @@
...
+    # Assume "c" is the second part
+    # TODO: Handle scalar array case with "c" holds the data
+    if chunk_key_encoding == "/":
+        indexes = tuple(int(ind) for ind in parts[2:])
+    else:
+        indexes = tuple(int(ind) for ind in parts[2].split(chunk_key_encoding))
+    return var, indexes
+
+
+def find_matching_store(stores: dict[str, Any], request_key: str) -> StoreRequest:
```
kylebarron, line 140:
```suggestion
def find_matching_store(stores: StoreDict, request_key: str) -> StoreRequest:
```

pyproject.toml
kylebarron: And given that you manage multiple stores internally, I think it's fine to keep this as a separate implementation

virtualizarr/manifests/store.py
kylebarron: fwiw on the sphinx route you just need to add that configuration here: https://github.com/zarr-developers/VirtualiZarr/blob/develop/docs/conf.py

pyproject.toml
maxrjones: Do you think we should pin the specific version (e.g., v0.6.0)? I.e., do you expect we'll encounter breaking changing in the features we're using here?

kylebarron: I don't think it should be pinned. Perhaps a lower bound of 0.5 or 0.5.1, but not an upper bound.

 I don't expect any breaking changes, especially to the methods for actually accessing data.

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,367 @@
...
+        A dictionary with URI prefixes for different stores as keys
+    request_key : str
+        A string to match against the dictionary keys
+
+    Returns:
+    --------
+    StoreRequest
+    """
+    # Sort keys by length in descending order to ensure longer, more specific matches take precedence
+    sorted_keys = sorted(stores.keys(), key=len, reverse=True)
```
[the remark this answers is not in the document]
maxrjones: I'm actually reconsidering whether we need the str component at all or if we could just use the ObjectStore prefix for matching to the file paths. So we would first search through prefixes to see if there's a match with the start of the key, then if not use a generic S3 store, LocalStore, etc based on the parsed protocol. Meaning the type would just be iterable[ObjectStore]. What do you think of that alternative API design?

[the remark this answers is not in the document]
kylebarron: The prefix doesn't include the bucket. And you could have the same prefix in multiple buckets. So I don't think that would work. 

I think the current idea is fine for now.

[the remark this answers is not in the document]
maxrjones: Sounds good, I added a TODO to consider the sequence of tuples ideas. This is all internal API so it should be easy to reconsider when trying it out in the other parts of the library.

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,367 @@
...
+        var, chunk_key = parse_manifest_index(key)
+        marr = self._manifest_group._manifest_dict[var]
+        manifest = marr._manifest
+
+        path = manifest._paths[*chunk_key]
+        offset = manifest._offsets[*chunk_key]
+        length = manifest._lengths[*chunk_key]
+        store_request = find_matching_store(stores=self._stores, request_key=path)
+        # Get the  configured object store instance that matches the path
+        store = self._stores[store_request.store_id]
```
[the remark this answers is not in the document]
maxrjones: good call, thanks. updated in [`ed0c1a5` (#490)](https://github.com/zarr-developers/VirtualiZarr/pull/490/commits/ed0c1a51a3649a74c73895f6677818241797871c)

virtualizarr/manifests/store.py
maxrjones: thanks for the note, I added a TODO in [`709788b` (#490)](https://github.com/zarr-developers/VirtualiZarr/pull/490/commits/709788b0451f33d8552b0315c1896566ff06a00a)

virtualizarr/manifests/group.py
maxrjones: I added a TODO for this idea in [`6bfdf4d` (#490)](https://github.com/zarr-developers/VirtualiZarr/pull/490/commits/6bfdf4d4a67b6f9860316f365fc5c986ce256067)

virtualizarr/manifests/store.py
```diff
@@ -0,0 +1,377 @@
...
+    --------
+    StoreRequest
+    """
+    # Sort keys by length in descending order to ensure longer, more specific matches take precedence
+    sorted_keys = sorted(stores.keys(), key=len, reverse=True)
+
+    # Check each key to see if it's a prefix of the uri_string
+    for key in sorted_keys:
+        if request_key.startswith(key):
+            return StoreRequest(store=stores[key], key=request_key[len(key) :])
```
maxrjones, line 163: I think using `key=request_key[len(key) :]` makes some assumptions that are not universally valid. I will look into this tomorrow.

virtualizarr/manifests/group.py
```diff
@@ -0,0 +1,39 @@
...
+ManifestArrayVariableMapping: TypeAlias = dict[str, ManifestArray]
+
+
+class ManifestGroup:
+    """
+    Virtualized representation of multiple ManifestArrays as a Zarr Group.
+    """
+
+    # TODO: Consider refactoring according to https://github.com/zarr-developers/VirtualiZarr/pull/490#discussion_r2007805272
+    _manifest_arrays: ManifestArrayVariableMapping
```
TomNicholas, line 16: Why not just 
```suggestion
    _manifest_arrays: Mapping[str, ManifestArray]
```
The actual implementation can still use a dict, but this choice of typing clearly communicated that the `Mapping` is not meant to be mutable once set (because the `ManifestStore` is not meant to be writable).

virtualizarr/manifests/group.py
```diff
@@ -0,0 +1,39 @@
...
+        attributes : attributes to include in Group metadata
+        manifest_dict : ManifestArrayVariableMapping
+        """
+
+        self._metadata = GroupMetadata(attributes=attributes)
+        self._manifest_arrays = manifest_arrays
+
+    def __str__(self) -> str:
+        return (
+            f"ManifestArrayVariableMapping({self._manifest_arrays}, {self._metadata})"
```
TomNicholas, line 38:
```suggestion
            f"ManifestGroup(manifest_arrays={self._manifest_arrays}, metadata={self._metadata})"
```

TomNicholas (approved):

TomNicholas:
This has tests, which pass, so LGTM so we can iterate + build upon it.

PR #502, opened 2025-03-24
Fix metadata equality for nan fill value

### LINKED ISSUE #501: Bug with comparing ManifestArrays using `__eq__` which have `fill_value`=NaN
> I'm stuck on why this `assert_identical` on the datasets doesn't pass. It complains that all the coordinates are different, but every attribute I can think to check looks identical 😖

_Originally posted by @TomNicholas in https://github.com/zarr-developers/VirtualiZarr/pull/349#discussion_r1888894931_

I found out what the problem is - two `ManifestArrays` with identical `.metadata` will compare not equal if they both have `NaN` for a `fill_value`. This is because the `__eq__` check introspects deeper until it finds e.g. the `np.float32(nan)` type, but 

```python
In [4]: bool(np.float32('nan') == np.float32('nan'))
Out[4]: False
```

(See https://stackoverflow.com/a/10059796 for why numpy NaNs behave like this.)

The solution needs to be to actually check two `Metadata` classes are `__eq__` with dedicated code, not just trusting the python dataclasses' automatically-generated `__eq__` method to fi it.

I would ideally like to move that fix upstream into zarr-python.

TomNicholas:
This is a stopgap until https://github.com/zarr-developers/zarr-python/issues/2929 is fixed upstream. We will want the regression test regardless though.

- [x] Closes #501
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
 requires-python = ">=3.11"
 dynamic = ["version"]
 dependencies = [
-    "xarray>=2025.1.1",
+    #"xarray>=2025.3.0",
     "numpy>=2.0.0",
     "universal-pathlib",
     "numcodecs>=0.15.1",
@@ -85,7 +85,7 @@ all_writers = [
 # Dependency sets under dependencies-groups are NOT available via PyPI
 [dependency-groups]
 upstream = [
-    'xarray @ git+https://github.com/pydata/xarray',
+    #'xarray @ git+https://github.com/pydata/xarray',
     'universal_pathlib @ git+https://github.com/fsspec/universal_pathlib',
     'numcodecs @ git+https://github.com/zarr-developers/numcodecs',
     'ujson @ git+https://github.com/ultrajson/ultrajson',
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -85,7 +85,7 @@ all_writers = [
 # Dependency sets under dependencies-groups are NOT available via PyPI
 [dependency-groups]
 upstream = [
-    'xarray @ git+https://github.com/pydata/xarray',
+    #'xarray @ git+https://github.com/pydata/xarray',
```
TomNicholas, line 88:
```suggestion
    'xarray @ git+https://github.com/pydata/xarray',
```

pyproject.toml
```diff
@@ -95,7 +95,7 @@ upstream = [
     'fsspec @ git+https://github.com/fsspec/filesystem_spec',
     's3fs @ git+https://github.com/fsspec/s3fs',
     'kerchunk @ git+https://github.com/fsspec/kerchunk',
-    'icechunk @ git+https://github.com/earth-mover/icechunk#subdirectory=icechunk-python',
+    #'icechunk @ git+https://github.com/earth-mover/icechunk#subdirectory=icechunk-python',
```
TomNicholas, line 98:
```suggestion
    'icechunk @ git+https://github.com/earth-mover/icechunk#subdirectory=icechunk-python',
```

Issue #500, opened 2025-03-24
Use mkdocs-material instead of sphinx


maxrjones:
@kylebarron has been subtly 🙈 advocating for mkdocs over sphinx for Python projects because they [look nice](https://developmentseed.org/obstore/latest/) and don't involve any RST. I haven't used mkdocs, but am curious so I thought I'd move the comment to a specific issue for discussion.

_Originally posted by @kylebarron in https://github.com/zarr-developers/VirtualiZarr/pull/490#discussion_r2010683170_

TomNicholas:
I don't like writing RST but other than that don't have an opinion. If there is a solution that lets us have executable markdown docs that sounds great.

kylebarron:
What specifically do you mean by `executable`?

kylebarron:
I think the closest thing to what you probably mean is the https://github.com/danielfrg/mkdocs-jupyter. I've used it in Lonboard e.g. [here](https://developmentseed.org/lonboard/latest/examples/internet-speeds/), but in Lonboard I don't include the notebook outputs because the Jupyter notebook - ipywidgets - mkdocs - html integration doesn't work (at least for lonboard specifically)

TomNicholas:
We don't really need html integration to work for virtualizarr (the printable representation is fine), but we definitely do want the cells to actually be executed by the docs build. See https://github.com/zarr-developers/VirtualiZarr/pull/486

cc @ianhi

kylebarron:
Oh ok, so it's already in notebook form anyways. That should be easy (famous last words) to switch and prototype in mkdocs

ianhi:
My experience is that the mkdocs ecosystem is less well developed for notebooks than the equivalent using `myst`/`jupyterbook`

If the primary motivation for the swap is the looks we can also use https://jbms.github.io/sphinx-immaterial/

kylebarron:
https://github.com/danielfrg/mkdocs-jupyter allows you to execute notebooks when building docs.

To be clear I'm not even a committer here at this point, so don't change just for me. But I've had tons of pain in the past with both sphinx and RST. So from my perspective it's not so much as the looks as the ease-of-use for developing the site and writing documentation that mkdocs, mkdocs-material, and mkdocstrings make way easier and simpler today.

TomNicholas:
> To be clear I'm not even a committer here at this point, so don't change just for me. But I've had tons of pain in the past with both sphinx and RST. 

[...]

PR #503, opened 2025-03-24
Setup intersphinx mapping for docs

maxrjones:
@kylebarron made a great suggestion in https://github.com/zarr-developers/VirtualiZarr/pull/490#discussion_r2010683170 to use intersphinx references. This basically makes all the types in the function signatures clickable to get to the relevant documentation in NumPy, Zarr, Python, and Xarray, which is pretty great.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,6 +19,7 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "sphinx.ext.extlinks",
+    "sphinx.ext.intersphinx",
     "sphinx_copybutton",
     "sphinx_togglebutton",
     "sphinx_design",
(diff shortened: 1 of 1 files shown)


TomNicholas (approved):
Nice!

docs/conf.py
```diff
@@ -31,6 +32,16 @@
...
 }
 
+# Example configuration for intersphinx: refer to the Python standard library.
+# use in refs e.g:
+# :ref:`comparison manual <python:comparisons>`
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3/", None),
+    "numpy": ("https://numpy.org/doc/stable/", None),
+    "zarr": ("https://zarr.readthedocs.io/en/stable/", None),
+    "xarray": ("https://docs.xarray.dev/en/stable/", None),
```
kylebarron, line 42: Add obstore too? Or you want to add it in the other PR?

maxrjones: Yeah I figured I would add it in the other PR, since it's not used anywhere in VirtualiZarr yet.

docs/conf.py
```diff
@@ -31,6 +32,16 @@
     "discussion": ("https://github.com/zarr-developers/virtualizarr/discussions/%s", "D%s"),
 }
 
+# Example configuration for intersphinx: refer to the Python standard library.
+# use in refs e.g:
+# :ref:`comparison manual <python:comparisons>`
+intersphinx_mapping = {
```
kylebarron, line 38: Also might as well add all your dependencies. You should add all dependencies even optional ones IMO

- numcodecs: https://numcodecs.readthedocs.io/en/stable/objects.inv
- fsspec: https://filesystem-spec.readthedocs.io/en/latest/objects.inv
- requests: https://requests.readthedocs.io/en/latest/objects.inv
- aiohttp: https://docs.aiohttp.org/en/stable/objects.inv
- s3fs: https://s3fs.readthedocs.io/en/latest/objects.inv
- h5py: https://docs.h5py.org/en/stable/objects.inv

I didn't go through the whole list of optional dependencies but you get the picture.

At least with mkdocs, it's important to add all of these intersphinx linkings so that _all_ types get automatically linked. E.g. here where `RecordBatch` [automatically links to the arro3 docs](https://kylebarron.dev/arro3/latest/api/core/record-batch/#arro3.core.RecordBatch)

image

maxrjones: shucks, didn't see this before merging. Do you feel like opening a PR? Then you can definitely tell us what to do in https://github.com/zarr-developers/VirtualiZarr/issues/500 😆

kylebarron: If it's easy for you to make a new PR with just these changes that would be easier. I'm also not sure if sphinx actually shows the interlinking of all of these dependencies. But it doesn't hurt to have all of these interlinkings defined in your config; these links get resolved when building the docs site

PR #507, opened 2025-03-25
Test ManifestStore against S3 and fix key parsing

### LINKED ISSUE #97: Should we add minIO to test suite?
Pangeo-Forge uses minIO to test against s3-like buckets. This could be useful for the current s3 / fsspeec reading work.

maxrjones:
This PR test the ManifestStore against S3 using MinIO and fixes a small issue with translating ManifestArray keys to the appropriate key for `obstore.get()`.

- [x] Closes #97 
- [x] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


virtualizarr/tests/test_remote_integration/test_remote_icechunk.py
TomNicholas, line 1: Can this not just go in `test_icechunk.py`?

virtualizarr/tests/test_remote_integration/test_remote_icechunk.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # check array metadata
+    assert arr.metadata.zarr_format == 3
+    assert arr.shape == (2, 3)
+    assert arr.chunks == (2, 3)
+    assert arr.dtype == np.dtype("<i8")
+    assert arr.order == "C"
+    assert arr.fill_value == 0
+    # TODO check compressor, filters?
+    #
```
TomNicholas, line 74: None of this seems specific to remote files. Why would this not be sufficiently covered by just using a local icechunk store?

TomNicholas (commented):
This is cool, but I'm not totally sure I understand what you're intending to test here, or how it relates to the `ManifestStore` work. Surely dealing with remote object storage is the responsibility of `obstore` and `icechunk`?

maxrjones:
> This is cool, but I'm not totally sure I understand what you're intending to test here, or how it relates to the `ManifestStore` work. Surely dealing with remote object storage is the responsibility of `obstore` and `icechunk`?

I would like to test that `ManifestStore.get()` and `ManifestStore.list_prefix()` return the same expected results for files on both local and object storage. Since `ManifestStore` requires some URL manipulation, it is possible that we make assumptions valid for local filesystems but not object storage as mentioned as a concern in https://github.com/zarr-developers/VirtualiZarr/pull/490#discussion_r2010950114. 

I built this as a separate PR under the assumption that PR scope should be as constrained as possible and that there were some Icechunk tests that would benefit from minio., but didn't pick a good test to start from. I'll modify this PR to test https://github.com/zarr-developers/VirtualiZarr/blob/206277f68bdae637ccee3a178b42f06c10c93851/virtualizarr/tests/test_backend.py#L206-L232 instead, which is explicitly testing for S3 compatibilty.

TomNicholas:
> I would like to test that ManifestStore.get() and ManifestStore.list_prefix() return the same expected results for files on both local and object storage.

Thanks for the explanation, that makes a lot more sense.

I just want to avoid the rabbit hole of adding tests to VirtualiZarr that are really just testing that Icechunk does what it says it does.

> I built this as a separate PR under the assumption that PR scope should be as constrained as possible

FWIW I would be happy to merge your `ManifestStore` as it is and iterate on testing that.

maxrjones:
> FWIW I would be happy to merge your ManifestStore as it is and iterate on testing that.

Oh wow, that'd make this so much easier!

virtualizarr/manifests/store.py
```diff
@@ -160,7 +161,8 @@ def find_matching_store(stores: StoreDict, request_key: str) -> StoreRequest:
     # Check each key to see if it's a prefix of the uri_string
     for key in sorted_keys:
         if request_key.startswith(key):
-            return StoreRequest(store=stores[key], key=request_key[len(key) :])
+            parsed_key = urlparse(request_key)
+            return StoreRequest(store=stores[key], key=parsed_key.path)
```
maxrjones, line 165: This is the only non-test related change in the PR, which makes the ManifestStore work for both local and object stores

virtualizarr/tests/test_manifests/test_store.py
```diff
@@ -16,14 +16,11 @@
...
     ManifestStore,
 )
-from virtualizarr.tests import (
-    requires_obstore,
-)
+from virtualizarr.tests import requires_obstore
 
 
 @pytest.fixture()
-@requires_obstore
```
TomNicholas, line 25: Why does this test not require obstore?

virtualizarr/tests/test_manifests/test_store.py
```diff
@@ -33,19 +30,64 @@ def filepath(tmpdir):
...
+    store = obs.store.S3Store(
+        minio_bucket["bucket"],
+        aws_endpoint=minio_bucket["endpoint"],
+        access_key_id=minio_bucket["username"],
+        secret_access_key=minio_bucket["password"],
+        virtual_hosted_style_request=False,
+        client_options={"allow_http": True},
+    )
+    filepath = "data.tmp"
+    obs.put(
```
TomNicholas, line 65: You only did one put, but now there are 4 chunks in there?

virtualizarr/tests/test_manifests/test_store.py
```diff
@@ -55,71 +97,75 @@ def manifest_store(filepath, array_v3_metadata):
...
+        assert local_store.read_only
+        assert local_store.supports_listing
+        assert not local_store.supports_deletes
+        assert not local_store.supports_writes
+        assert not local_store.supports_partial_writes
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "manifest_store",
+        ["local_store", pytest.param("s3_store", marks=pytest.mark.minio)],
```
TomNicholas, line 115: Should the mark be `requires_minio`?

virtualizarr/tests/test_manifests/test_store.py
```diff
@@ -33,19 +30,64 @@ def filepath(tmpdir):
...
+            "offset": 8,
+            "length": 4,
+        },
+        "1.1": {
+            "path": f"s3://{minio_bucket['bucket']}/{filepath}",
+            "offset": 12,
+            "length": 4,
+        },
     }
     manifest = ChunkManifest(entries=chunk_dict)
```
TomNicholas, line 92: Is there a way to split this fixture up a bit, to more clearly delineate the creation of data in a (minio) s3-like obstore from the creation of the `ManifestStore` object which contains the byte references pointing to it?

TomNicholas (approved):
Nice! This does seem useful.

virtualizarr/tests/test_manifests/test_store.py
maxrjones: These markers don't make any impact on fixtures, so I removed it. The fixture is generated if any tests using the fixtures are captured by pytest, so the markers should all go on tests rather than fixtures.

virtualizarr/tests/test_manifests/test_store.py
maxrjones: They're synonymous, but I will re-add requires_minio since it's more descriptive

virtualizarr/tests/test_manifests/test_store.py
maxrjones: good call, done in [`874c6df` (#507)](https://github.com/zarr-developers/VirtualiZarr/pull/507/commits/874c6df4fa26a1271720f6f6a1c04575429d13ab)

virtualizarr/tests/test_manifests/test_store.py
maxrjones: I added a docstring explaining the approach in [`874c6df` (#507)](https://github.com/zarr-developers/VirtualiZarr/pull/507/commits/874c6df4fa26a1271720f6f6a1c04575429d13ab). This is kinda similar to libraries that concurrently compress each chunk (or not for uncompressed data), concatenate them, and then write then to a file in one go. At least, *I think that's a thing*.

Issue #498, opened 2025-03-21
Redefine Virtual Readers as `func(filepath) -> ManifestStore`


TomNicholas:
Once we've completed https://github.com/zarr-developers/VirtualiZarr/issues/473, the responsibility of each reader will essentially be reduced to just creating a fully virtual `ManifestStore` instance from the given filepath. 

This means that we could redefine what a "virtual reader" is, as simply a filetype-specific function of the form

```python
def virtual_reader(filepath: str) -> ManifestStore:
    ...
```

where there is one of these readers for TIFF files, one for HDF files, etc. etc.

As described in #473, `loadable_variables` is then not ever even passed to the reader function, as instead it's always handled by virtualizarr in some subsequent code like this:

```python

ds_virtual = manifeststore.to_virtual_ds()

ds_loadable = xr.open_zarr(manifeststore, drop_variables=list(ds_virtual.variables))

vds = xr.merge(ds_virtual, ds_loadable)
```

This is mostly great, BUT it does cause some issues for the special set of readers who read from other virtual references formats (i.e. DMR++ and Kerchunk) - see https://github.com/zarr-developers/VirtualiZarr/pull/477#issuecomment-2744448626.

Another issue it causes is that it becomes tricky to solve https://github.com/zarr-developers/VirtualiZarr/issues/489 - when reading inlined data the kerchunk reader would prefer to return a loadable variable, but this manifeststore-first architecture doesn't really allow that. Unless we used the ability of `obstore` to wrap an `MemoryStore` too...?

cc @maxrjones

TomNicholas:
This would also mean we had no need for a more complex entrypoint system like https://github.com/zarr-developers/VirtualiZarr/issues/245

maxrjones:
How do you think we should handle storage configuration in this new virual reader structure? We could have users separately configure on ObjectStore instance and pass that to the reader:

```Python
def virtual_reader(filepath: str, store: ObjectStore) -> ManifestStore:
    ...
```

Another option would be to accept arbitrary kwargs for storage creation and try to infer the storage type from the url:

```Python
def virtual_reader(url: str, storage_options: dict) -> ManifestStore:
    ...
```

Or perhaps we could define some sort of configuration container similar to [Icechunk's approach](https://icechunk.io/en/latest/icechunk-python/configuration/#virtual_chunk_containers).

[...]

PR #516, opened 2025-03-27
Add a function to produce a ManifestStore from HDF5 files

maxrjones:
First draft of first steps for addressing #498, based on changes originally proposed in #490

### DIFF
--- a/virtualizarr/readers/hdf/hdf.py
+++ b/virtualizarr/readers/hdf/hdf.py
@@ -42,8 +42,8 @@ h5py = soft_import("h5py", "For reading hdf files", strict=False)
 
 
 if TYPE_CHECKING:
-    from h5py import Dataset as H5Dataset  # type: ignore[import-untyped]
-    from h5py import Group as H5Group  # type: ignore[import-untyped]
+    from h5py import Dataset as H5Dataset
+    from h5py import Group as H5Group
     from obstore.store import ObjectStore
 
 FillValueType = Union[
--- a/virtualizarr/readers/hdf/hdf.py
+++ b/virtualizarr/readers/hdf/hdf.py
@@ -150,7 +150,7 @@ class HDFVirtualBackend(VirtualBackend):
                     )
                     if variable is not None:
                         manifest_dict[key] = variable
-        return ManifestGroup(manifest_arrays=manifest_dict, attributes=attrs)
+        return ManifestGroup(arrays=manifest_dict, attributes=attrs)
 
     @staticmethod
     def _create_manifest_store(
@@ -165,7 +165,7 @@ class HDFVirtualBackend(VirtualBackend):
             store=store, filepath=filepath, group=group
         )
         # Convert to a manifest store
-        return ManifestStore(stores={prefix: store}, manifest_group=manifest_group)
+        return ManifestStore(stores={prefix: store}, group=manifest_group)
 
     @staticmethod
     def open_virtual_dataset(
--- a/virtualizarr/readers/hdf/hdf.py
+++ b/virtualizarr/readers/hdf/hdf.py
@@ -228,7 +228,7 @@ class HDFVirtualBackend(VirtualBackend):
     def _dataset_chunk_manifest(
         path: str,
         dataset: H5Dataset,
-    ) -> Optional[ChunkManifest]:
+    ) -> ChunkManifest:
         """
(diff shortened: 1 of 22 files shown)


maxrjones:
FYI I'll return to this PR later to fix the failing test - the issue is that we now have support for multiple chunk_key_encodings but no way yet to specify which is used in the Zarr V3 array metadata.

TomNicholas:
This seems good - you could potentially go the whole way to using `open_zarr` to implement `loadable_variables` inside the HDF reader in this PR. Then we do that for every other reader, then we factor that outside of the readers (i.e. reduce them to each just being a function).

maxrjones:
> This seems good - you could potentially go the whole way to using `open_zarr` to implement `loadable_variables` inside the HDF reader in this PR. Then we do that for every other reader, then we factor that outside of the readers (i.e. reduce them to each just being a function).

[The abstraction described here](https://github.com/zarr-developers/VirtualiZarr/issues/498#issuecomment-2746332013) is a necessary prerequisite. Basically we need a utility function like `get_default_object_store(filepath) -> ObjectStore` that goes from a valid URL to an ObjectStore instance in order to make  the `store` argument in `_create_manifest_store()` optional.

While there's a lot of code in https://github.com/developmentseed/obstore/blob/main/obstore/python/obstore/fsspec.py that would help the development of that function, I think it's a big enough task to warrant it's own PR.

Somewhat relatedly, I'm planning to take tomorrow off and don't want to slow you down. So, my vote would be for you or @sharkinsspatial to take over this PR if you'd want to make progress before Monday.

maxrjones:
The typing error is because we don't have a way to handle "empty" chunk manifests in a ManifestArray. We need an analogous solution to [the current approach of an empty numpy array in a xarray Variable](https://github.com/zarr-developers/VirtualiZarr/blob/0cd29b8624bbcc703cd174218b5d0c21115f46dc/virtualizarr/readers/hdf/hdf.py#L358-L359), but I don't understand the circumstances requiring an empty chunk manifest well enough to propose a solution.

virtualizarr/readers/hdf/hdf.py
```diff
@@ -58,6 +62,111 @@
 
 
 class HDFVirtualBackend(VirtualBackend):
+    @staticmethod
+    def _construct_manifest_array(
```
TomNicholas, line 66: Every `staticmethod` here aside from `open_virtual_dataset` may as well just be an actual function. You will need to do that anyway later to refactor to make the reader just one function.

TomNicholas: But you could leave that to do in #498.

virtualizarr/tests/test_readers/test_hdf/test_hdf.py
```diff
@@ -16,19 +16,18 @@ class TestDatasetChunkManifest:
...
     def test_empty_dataset(self, empty_dataset_hdf5_file):
         f = h5py.File(empty_dataset_hdf5_file)
         ds = f["data"]
-        with pytest.raises(ValueError, match="no space allocated in the file"):
-            HDFVirtualBackend._dataset_chunk_manifest(
-                path=empty_dataset_hdf5_file, dataset=ds
-            )
+        manifest = HDFVirtualBackend._dataset_chunk_manifest(
+            path=empty_dataset_hdf5_file, dataset=ds
+        )
```
TomNicholas, line 29: Generally it's better to write tests that use higher-level public API instead of directly calling internals if you can help it, but that's just a nit.

TomNicholas (approved):

virtualizarr/readers/hdf/hdf.py
sharkinsspatial: 👍 That is the plan.  I think @maxrjones intent here was to make a very small change set to demonstrate the feasibility of using `ManifestStore` and that we would refactor the structure in a PR for https://github.com/zarr-developers/VirtualiZarr/issues/498.

virtualizarr/tests/test_readers/test_hdf/test_hdf.py
sharkinsspatial: 👍 I think the tests will be easier to structure when we move away from `VirtualBacked` base to something more functional and the returned `ManifestStore` will be simpler to introspect and reason about than a complete `xr.Dataset`.

PR #519, opened 2025-03-31
Run upstream tests nightly, commits to main/dev, and on labeled PRs

maxrjones:
The upstream build takes a while and IMO doesn't need to be run on every commit to PRs. This separates out the upstream tests so that they only get run on a commit to main/develop, on the nightly chron job, and when a PR is labeled with 'test-upstream'.


abarciauskas-bgse (approved):

TomNicholas:
Good idea, thanks!

PR #523, opened 2025-04-01
Properly extract chunk keys for arrays with a single chunk

maxrjones:
This PR improves ManifestStore to support ManifestArrays that only contain a single entry in paths, offsets, and lengths. It was originally included in https://github.com/zarr-developers/VirtualiZarr/pull/516 but I've pulled these changes out because they are also needed for the TIFF reader.

### DIFF
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -108,17 +108,12 @@ def get_zarr_metadata(manifest_group: ManifestGroup, key: str) -> Buffer:
         return dict_to_buffer(metadata, prototype=default_buffer_prototype())
 
 
-def parse_manifest_index(
-    key: str, chunk_key_encoding: str = "."
-) -> tuple[str, tuple[int, ...]]:
+def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int, ...]:
     """
     Splits `key` provided to a zarr store into the variable indicated
     by the first part and the chunk index from the 3rd through last parts,
     which can be used to index into the ndarrays containing paths, offsets,
     and lengths in ManifestArrays.
-
-    Currently only works for 1d+ arrays with a tree depth of one from the
-    root Zarr group.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


virtualizarr/manifests/store.py
```diff
@@ -108,36 +108,29 @@ def get_zarr_metadata(manifest_group: ManifestGroup, key: str) -> Buffer:
...
-    # Assume "c" is the second part
-    # TODO: Handle scalar array case with "c" holds the data
-    if chunk_key_encoding == "/":
-        indexes = tuple(int(ind) for ind in parts[2:])
-    else:
-        indexes = tuple(int(ind) for ind in parts[2].split(chunk_key_encoding))
-    return var, indexes
+    if key.endswith("c"):
+        # Scalar arrays hold the data in the "c" key
+        return (0,)
```
TomNicholas, line 129: I'm not sure I understand this. You're special-casing getting a scalar array by indexing into the underlying manifestarray as if it were 1D with length 1.

But AFAIK the ManifestArray doesn't currently make a special case for that? If anything it looks like scalars might not be implemented??

https://github.com/zarr-developers/VirtualiZarr/blob/25c0b3889d13bfc275d702437e15ae3a1492be98/virtualizarr/manifests/array.py#L65

I also don't understand how your adjustments to the test in this PR cover this if statement added here.

maxrjones: I'll work on adding a test for this. The existing changes to the tests were unrelated to this and instead properly specified the chunk_key_encoding.

maxrjones: I decided to just raise a NotImplementedError because this edge case isn't a priority, this PR is slowing progress on TIFF support, and this edge case will be easier to test for after the HDF5 reader uses ManifestStore

maxrjones:
@TomNicholas Sean and I think #516 is much ready to be merged, but I'd like to get this in first to keep changes isolated between PRs. Do you have any issues with the approach of punting on the scalar array special case?

TomNicholas:
If it currently doesn't work and you're just formalizing that by raising a `NotImplementedError` then I think that's totally fine. Might want to raise an issue to track it if there isn't one already though.

PR #529, opened 2025-04-03
Fix resource leak from unclosed dataset

chuckwondo:
Eliminate 219 of the following warnings emitted during run of tests (via `pixi run -e test run-tests`):

```
RuntimeWarning: deallocating CachingFileManager(...), but file is not already closed. This may indicate a bug.
    vds = replace_virtual_with_loadable_vars(
```

- [x] Tests passing


TomNicholas (approved):
Haha excellent! Thank you @chuckwondo

PR #538, opened 2025-04-07
Support writing xarray.DataTree to Icechunk store

### LINKED ISSUE #244: allow writing `DataTree` objects containing references to disk
In trying to create a nicer way to access ocean model output (several stacks of netcdf files where each stack can be concatenated, but not necessarily merged into a single `Dataset` object), I've been able to construct a `DataTree` object:
```python
hourly = xr.concat([virtualizarr.open_virtual_dataset(path, ...) for path in paths], dim="time", ...)
daily = xr.concat([virtualizarr.open_virtual_dataset(path, ...) for path in paths], dim="time", ...)
monthly = xr.concat([virtualizarr.open_virtual_dataset(path, ...) for path in paths], dim="time", ...)

tree = DataTree.from_dict({"/": ..., "/hourly": hourly, "/daily": daily, "/monthly": monthly})
```
but would then need a way to write that tree to disk.

The current file formats (except maybe parquet, but not sure), definitely support this since they're based on `zarr`, we'd just need to create a `DataTree` accessor and write the code to serialize `DataTree` objects containing `ManifestArray`s.

Edit: related to #84 and #11

chuckwondo:
- [x] Closes #244
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_integration.py
+++ b/virtualizarr/tests/test_integration.py
@@ -1,6 +1,7 @@
+from collections.abc import Mapping
 from os.path import relpath
 from pathlib import Path
-from typing import Callable, Concatenate, TypeAlias
+from typing import Any, Callable, Concatenate, TypeAlias, overload
 
 import numpy as np
 import pytest
@@ -23,7 +24,9 @@ from virtualizarr.translators.kerchunk import (
     dataset_from_kerchunk_refs,
 )
 
-RoundtripFunction: TypeAlias = Callable[Concatenate[xr.Dataset, Path, ...], xr.Dataset]
+RoundtripFunction: TypeAlias = Callable[
+    Concatenate[xr.Dataset | xr.DataTree, Path, ...], xr.Dataset | xr.DataTree
+]
 
 
 def test_kerchunk_roundtrip_in_memory_no_concat(array_v3_metadata):
@@ -111,7 +114,22 @@ def roundtrip_as_kerchunk_parquet(vds: xr.Dataset, tmpdir, **kwargs):
     return xr.open_dataset(f"{tmpdir}/refs.parquet", engine="kerchunk", **kwargs)
 
 
-def roundtrip_as_in_memory_icechunk(vds: xr.Dataset, tmpdir, **kwargs):
+@overload
+def roundtrip_as_in_memory_icechunk(
+    vdata: xr.Dataset, tmp_path: Path, **kwargs
+) -> xr.Dataset: ...
+@overload
+def roundtrip_as_in_memory_icechunk(
+    vdata: xr.DataTree, tmp_path: Path, **kwargs
+) -> xr.DataTree: ...
+
+
+def roundtrip_as_in_memory_icechunk(
+    vdata: xr.Dataset | xr.DataTree,
+    tmp_path: Path,
+    virtualize_kwargs: Mapping[str, Any] | None = None,
+    **kwargs,
+) -> xr.Dataset | xr.DataTree:
     from icechunk import Repository, Storage
 
     # create an in-memory icechunk store
@@ -120,7 +138,17 @@ def roundtrip_as_in_memory_icechunk(vds: xr.Dataset, tmpdir, **kwargs):
     session = repo.writable_session("main")
 
     # write those references to an icechunk store
-    vds.virtualize.to_icechunk(session.store)
+    vdata.virtualize.to_icechunk(session.store, **(virtualize_kwargs or {}))
+
+    if isinstance(vdata, xr.DataTree):
+        # read the dataset from icechunk
+        return xr.open_datatree(
+            session.store,  # type: ignore
+            engine="zarr",
+            zarr_format=3,
+            consolidated=False,
+            **kwargs,
+        )
 
     # read the dataset from icechunk
     return xr.open_zarr(session.store, zarr_format=3, consolidated=False, **kwargs)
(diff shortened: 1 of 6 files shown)


virtualizarr/writers/icechunk.py
```diff
@@ -118,6 +117,75 @@ def dataset_to_icechunk(
...
+        raise TypeError(
+            "last_updated_at: expected type datetime,"
+            f" but got type {type(last_updated_at)}"
+        )
+
+    if store.read_only:
+        raise ValueError("supplied store is read-only")
+
+    for path, subtree in dt.subtree_with_keys:
+        tree = cast(xr.DataTree, subtree)  # subtree is typed as Unknown
```
TomNicholas, line 174: What do you mean its typed as Unknown? `.subtree_with_keys` has a type hint - is it incorrect?

https://github.com/pydata/xarray/blob/08fa7b9fedaa977eed6ef4de1dbd8232b8b1ceed/xarray/core/treenode.py#L420

virtualizarr/writers/icechunk.py
```diff
@@ -118,6 +117,75 @@ def dataset_to_icechunk(
...
+        ds = tree.to_dataset()
```
TomNicholas, line 175: There's some important subtleties here with inherited coordinate variables. To get the same behaviour as  `DataTree.to_zarr()` you need to pass the `inherit` kwargs, which should be exposed publicly by this accessor method as `write_inherited_coords`. See 

https://github.com/pydata/xarray/blob/08fa7b9fedaa977eed6ef4de1dbd8232b8b1ceed/xarray/core/datatree_io.py#L124

virtualizarr/tests/test_integration.py
```diff
@@ -302,6 +326,93 @@ def test_datetime64_dtype_fill_value(
...
+                                ds1.coords[coord].attrs
+                                == roundtrip_vds1.coords[coord].attrs
+                            )
+                        for coord in ds2.coords:
+                            assert (
+                                ds2.coords[coord].attrs
+                                == roundtrip_vds2.coords[coord].attrs
+                            )
+                    else:
+                        # they are very very close! But assert_allclose doesn't seem to work on datetimes
```
TomNicholas, line 404: Is this comment copied from elsewhere or did you discover this?

TomNicholas (changes requested):
Thanks @chuckwondo ! This looks great, except for one important subtlety.

virtualizarr/writers/icechunk.py
chuckwondo: Ah, thanks for the code reference. I'll adjust accordingly.

virtualizarr/tests/test_integration.py
chuckwondo: I copied this test function from `test_kerchunk_roundtrip_concat` in the same file (which includes the very same comment), and adjusted for testing DataTree roundtrip, but with only the "in memory" roundtrip function.

However, since reading your comment here, I see that the comment is no longer accurate in either test function. That is, `assert_allclose` does work on datetimes, so I have adjusted both tests accordingly.

virtualizarr/writers/icechunk.py
TomNicholas: This is probably what it should have been

```python
def subtree_with_keys(self) -> Iterator[tuple[str, Self]]:
    ...
```

so that's something that apparently should be fixed upstream.

virtualizarr/writers/icechunk.py
```diff
@@ -118,6 +117,82 @@ def dataset_to_icechunk(
...
+        raise ValueError("supplied store is read-only")
+
+    for path, subtree in dt.subtree_with_keys:
+        tree = cast(xr.DataTree, subtree)  # subtree is typed as Unknown
+        at_root = tree is dt
+        ds = tree.to_dataset(write_inherited_coords or at_root)
+        store_path = StorePath(store, path="" if at_root else tree.relative_to(dt))
+        group = Group.from_store(store=store_path, zarr_format=3)
+
+        if len(ds):
```
TomNicholas, line 186: I see why you put this here (to only write if there are any variables), but I don't think we want this check. 

Instead, our write function should be able to write to a group even if there are no variables, because even just `ds.attrs` should be written to the store (as zarr group metadata). This may be an edge case that doesn't currently work for `Dataset.virtualize.to_icechunk()`...

chuckwondo: I removed the length check, and tests still passed. I also tweaked the test to nest 1 of the 2 virtual datasets in the datatree, just to be sure nesting below the root works, and all is good with that too.

IIRC, when I initially started on these changes, I encountered an error when attempting to write an "empty" dataset, which is why I originally added the length check. It seems that my subsequent changes addressed the cause of that problem, so removing the check now works.

TomNicholas (approved):
Thanks @chuckwondo !

virtualizarr/writers/icechunk.py
```diff
@@ -118,6 +117,81 @@ def dataset_to_icechunk(
...
+            f" but got type {type(last_updated_at)}"
+        )
+
+    if store.read_only:
+        raise ValueError("supplied store is read-only")
+
+    for path, subtree in dt.subtree_with_keys:
+        tree = cast(xr.DataTree, subtree)  # subtree is typed as Unknown
+        at_root = tree is dt
+        ds = tree.to_dataset(write_inherited_coords or at_root)
```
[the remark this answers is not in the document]
TomNicholas: Sorry my bad - I thought there was a problem with this code, but looking at it again it's all good. LGTM.

PR #539, opened 2025-04-08
Fix incorrect assertion in test_loadable_variables

chuckwondo:
test_loadable_variables was comparing actual to actual values rather than actual to expected values.


TomNicholas (approved):
Good catch!

PR #547, opened 2025-04-11
Refactor kerchunk readers to use manifest store

TomNicholas:
Aims to refactor all the kerchunk-based readers to use `ManifestStore`, but to do that this PR first adds a `ManifestStore.from_kerchunk_refs()` method.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage
- [ ] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`
- [ ] New functionality has documentation

Issue #452, opened 2025-02-27
Docs on how to write a custom reader


TomNicholas:
Related to #245 , but doesn't actually require it.

PR #549, opened 2025-04-13
Add ObjectStoreRegistry to support default stores

maxrjones:
This PR makes it possible to have an empty dict in `ManifestStores._stores` and automatically create an S3Store or LocalStore from a filepath.

Part of https://github.com/zarr-developers/VirtualiZarr/issues/498

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,3 +22,13 @@ repos:
     rev: ebf0b5e44d67f8beaa1cd13a0d0393ea04c6058d
     hooks:
       - id: validate-cff
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.15.0
+    hooks:
+      - id: mypy
+        files: src|tests
+        additional_dependencies:
+          # Package dependencies
+          - obstore>=0.5.1
+          # Tests
+          - pytest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,13 +22,3 @@ repos:
     rev: ebf0b5e44d67f8beaa1cd13a0d0393ea04c6058d
     hooks:
       - id: validate-cff
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.15.0
-    hooks:
-      - id: mypy
-        files: src|tests
-        additional_dependencies:
-          # Package dependencies
-          - obstore>=0.5.1
-          # Tests
-          - pytest
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,9 +177,9 @@ run-mypy = { cmd = "mypy virtualizarr" }
 # is used), reduces test hangs that appear to be macOS-related.
 run-tests = { cmd = "pytest -n auto --dist loadscope --run-network-tests --verbose" }
 run-tests-no-network = { cmd = "pytest -n auto --verbose" }
-run-tests-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov=term-missing" }
-run-tests-xml-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov-report=xml" }
-run-tests-html-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov-report=html" }
+run-tests-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov=term-missing" }
+run-tests-xml-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov-report=xml" }
+run-tests-html-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov-report=html" }
 
 # Define which features and groups to include in different pixi (similar to conda) environments)
 [tool.pixi.environments]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -227,6 +227,7 @@ module = [
     "numcodecs.*",
     "ujson",
     "zarr",
+    "requests",
 ]
 ignore_missing_imports = true
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,9 +177,9 @@ run-mypy = { cmd = "mypy virtualizarr" }
 # is used), reduces test hangs that appear to be macOS-related.
 run-tests = { cmd = "pytest -n auto --dist loadscope --run-network-tests --verbose" }
 run-tests-no-network = { cmd = "pytest -n auto --verbose" }
-run-tests-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov=term-missing" }
-run-tests-xml-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov-report=xml" }
-run-tests-html-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov-report=html" }
+run-tests-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov=term-missing" }
+run-tests-xml-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov-report=xml" }
+run-tests-html-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov-report=html" }
 
 # Define which features and groups to include in different pixi (similar to conda) environments)
 [tool.pixi.environments]
@@ -272,7 +272,6 @@ line-ending = "auto"
 known-first-party = ["virtualizarr"]
 
 [tool.coverage.run]
-include = ["virtualizarr/"]
 omit = ["conftest.py", "virtualizarr/tests/*"]
 
 [tool.coverage.report]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -201,8 +201,8 @@ readthedocs = "rm -rf $READTHEDOCS_OUTPUT/html && cp -r docs/_build/html $READTH
 
 # Define commands to run within the docs environment
 [tool.pixi.feature.minio.tasks]
-run-tests = { cmd = "pytest virtualizarr/tests/test_manifests/test_store.py --run-minio-tests --verbose" }
-run-tests-xml-cov = { cmd = "pytest virtualizarr/tests/test_manifests/test_store.py --run-minio-tests --verbose --cov-report=xml" }
+run-tests = { cmd = "pytest virtualizarr/tests/test_manifests/test_store.py virtualizarr/tests/test_readers/test_hdf/test_hdf_manifest_store.py --run-minio-tests --run-network-tests --verbose" }
+run-tests-xml-cov = { cmd = "pytest virtualizarr/tests/test_manifests/test_store.py virtualizarr/tests/test_readers/test_hdf/test_hdf_manifest_store.py --run-minio-tests --run-network-tests --verbose --cov-report=xml" }
 
 [tool.setuptools_scm]
 fallback_version = "9999"
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -156,7 +156,8 @@ def default_object_store(filepath: str) -> ObjectStore:
 
 class ObjectStoreRegistry:
     """
-    ObjectStoreRegistry maps the URL scheme and netloc to ObjectStore instances, and allows ManifestStores to read from different ObjectStore instances.
+    ObjectStoreRegistry maps the URL scheme and netloc to ObjectStore instances. This register allows
+    Zarr Store implementations (e.g., ManifestStore) to read from different ObjectStore instances.
     """
 
     _stores: dict[str, ObjectStore]
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import pickle
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, Mapping
+from typing import TYPE_CHECKING
 from urllib.parse import urlparse
 
 from zarr.abc.store import (
@@ -19,13 +19,16 @@ from virtualizarr.manifests.array import ManifestArray
 from virtualizarr.manifests.group import ManifestGroup
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncGenerator, Iterable
+    from collections.abc import AsyncGenerator, Iterable, Mapping
     from typing import Any
 
-    from obstore.store import S3Config
+    import xarray as xr
+    from obstore.store import (
+        ObjectStore,  # type: ignore[import-not-found]
+        S3Config,
+    )
     from zarr.core.buffer import BufferPrototype
     from zarr.core.common import BytesLike
-
 
 __all__ = ["ManifestStore"]
 
@@ -43,12 +46,6 @@ from typing import TYPE_CHECKING, Any
 from zarr.core.buffer import default_buffer_prototype
 
 from virtualizarr.vendor.zarr.metadata import dict_to_buffer
-
-if TYPE_CHECKING:
-    import xarray as xr
-    from obstore.store import (
-        ObjectStore,  # type: ignore[import-not-found]
-    )
 
 
 @dataclass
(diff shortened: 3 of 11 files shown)


virtualizarr/manifests/store.py
```diff
@@ -137,34 +140,32 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
-    Find the matching store based on the store keys and the beginning of the URI strings,
-    to fetch data from the appropriately configured ObjectStore.
+def _default_object_store(
+    filepath: str, config: Optional[S3Config] = {}
+) -> tuple[str, ObjectStore]:
+    import obstore as obs
+
+    parsed = urlparse(filepath)
+    if parsed.scheme == "s3":
+        if not config:
```
sharkinsspatial, line 150: I think this `config` dict needs to be initialized prior to assignment.

chuckwondo (changes requested):

virtualizarr/manifests/store.py
```diff
@@ -137,34 +140,32 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
+        config["virtual_hosted_style_request"] = False
+        bucket = parsed.netloc
+        return f"s3://{bucket}", obs.store.S3Store(
+            bucket=bucket,
+            **config,
+        )
+    elif parsed.scheme == "":
+        return "file://", obs.store.LocalStore()
+    else:
+        raise NotImplementedError(f"{parsed.scheme} is not yet supported")
```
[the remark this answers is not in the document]
chuckwondo: These changes correctly call S3Store, which also address the mypy errors.

virtualizarr/manifests/store.py
```diff
@@ -44,7 +45,9 @@
 from virtualizarr.vendor.zarr.metadata import dict_to_buffer
 
 if TYPE_CHECKING:
-    from obstore.store import ObjectStore  # type: ignore[import-not-found]
+    from obstore.store import (
+        ObjectStore,  # type: ignore[import-not-found]
+    )
```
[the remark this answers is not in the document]
TomNicholas: > There are 2 if TYPE_CHECKING: blocks in this file

I noticed the same thing and fixed in it #547. I assume this was caused by a bad merge somewhere.

virtualizarr/manifests/store.py
```diff
@@ -137,34 +140,32 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
+
+    parsed = urlparse(filepath)
+    if parsed.scheme == "s3":
+        if not config:
+            config["skip_signature"] = True
+        config["virtual_hosted_style_request"] = True
+        config["client_options"] = {"allow_http": True}
+        config["virtual_hosted_style_request"] = False
+        bucket = parsed.netloc
+        return f"s3://{bucket}", obs.store.S3Store(
```
sharkinsspatial, line 156: This is failing on this [test](https://github.com/zarr-developers/VirtualiZarr/blob/develop/virtualizarr/tests/test_backend.py#L222).  We'll likely need some additional logic to infer bucket region https://developmentseed.org/obstore/latest/troubleshooting/aws/#region-required

virtualizarr/manifests/store.py
```diff
@@ -387,6 +388,39 @@ def to_virtual_dataset(
...
+            if request_key.startswith(prefix):
+                # Return an existing configured store and parsed request path
+                return StoreRequest(
+                    store=self._stores[prefix], key=parsed_request_key.path
+                )
+        # Use anonymous default store if not in pre-configured stores
+        prefix, store = _default_object_store(request_key)
+        # Add to stores for future use
+        self._stores[prefix] = store
+        self._stores = _sort_stores_by_prefix_length(self._stores)
```
TomNicholas, line 420: Can we just quickly think through the implications of doing this at GET time rather than at `ManifestStore` creation time? It's a little weird that we create a `ManifestStore` which then later edits itself.

Doing it at GET time means if the chunk location URL is nonsense, an error won't be thrown until we actually attempt to read data, which is worse than failing early.

But doing it at `ManifestStore` creation time (which should be possible because the contents of the `ManifestStore` should be immutable once created) means we have to create obstore stores even for URLs that we never intend to read data from. This will happen, for example when converting kerchunk/DMR++ references to Icechunk. I suppose these credentials and so on won't be _tested_ until GET time regardless...

TomNicholas: I believe Icechunk does something more like the latter. 

The general problem is that overall to read data we need to provide the chunk locations (in the `ChunkManifests`), the store information, and potentially credentials too.

According to @paraseba in icechunk :
> you "configure" your virtual chunk containers in the repo ... with all data but credentials
>
> then you pass the credentials as you open the repo, only when you want to read data

If we did something similar I think that would mean passing (or creating default) obstore stores during `ManifestStore.__init__`, but not requiring credentials at that point. Then later if `.get()` is called then the credentials might be required. Can we do a pattern like that with obstore @kylebarron ?

virtualizarr/manifests/store.py
```diff
@@ -205,7 +206,7 @@ def __init__(
         self,
         group: ManifestGroup,
         *,
-        stores: StoreDict,  # TODO: Consider using a sequence of tuples rather than a dict (see https://github.com/zarr-developers/VirtualiZarr/pull/490#discussion_r2010717898).
+        stores: StoreDict = {},
```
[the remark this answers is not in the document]
kylebarron: IMO you should have ruff lints turned on that catch this for you.

virtualizarr/manifests/store.py
kylebarron: Obstore stores are immutable; you can't change the behavior of a store after construction. Though maybe you could write a custom credential provider, which would be called the first time obstore makes a request.

maxrjones:
Thanks for the reviews, all. FYI I want to revise this to follow the same design as https://docs.rs/datafusion/latest/datafusion/datasource/object_store/trait.ObjectStoreRegistry.html. So a ManifestStore would have a ObjectStoreRegistry which would implement the `register_store` and `get_store` methods. But I've been a bit busy with other tasks and probably won't be able to return to this PR until tomorrow afternoon.

pyproject.toml
```diff
@@ -177,9 +177,9 @@ run-mypy = { cmd = "mypy virtualizarr" }
 # is used), reduces test hangs that appear to be macOS-related.
 run-tests = { cmd = "pytest -n auto --dist loadscope --run-network-tests --verbose" }
 run-tests-no-network = { cmd = "pytest -n auto --verbose" }
-run-tests-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov=term-missing" }
-run-tests-xml-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov-report=xml" }
-run-tests-html-cov = { cmd = "pytest -n auto --run-network-tests --verbose --cov-report=html" }
+run-tests-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov=term-missing" }
+run-tests-xml-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov-report=xml" }
+run-tests-html-cov = { cmd = "pytest -n 1 --run-network-tests --verbose --cov=virtualizarr --cov-report=html" }
```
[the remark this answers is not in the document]
maxrjones: oops, I didn't mean to commit this. I'll revert the change.

virtualizarr/manifests/store.py
maxrjones: @TomNicholas are there other runtime configuration options that you foresee in VirtualiZarr's future that could impact how we design the configuration of storage instances?

TomNicholas: I think the only two scenarios that VirtualiZarr has to handle are:
1. you only want to use the `ManifestStore` to represent chunk references in-memory so you don't need credentials to the actual chunk locations,
2. you want to actually use the `ManifestStore` to read chunk data so you do need credentials.

We could have credentials specified or not specified per virtual chunk container like IC supports, but that seems like more of an edge case, the important thing is to be able to avoid specifying credentials at all at store construction time to support case (1).

virtualizarr/manifests/store.py
maxrjones: addressed in [`ca48b6c` (#549)](https://github.com/zarr-developers/VirtualiZarr/pull/549/commits/ca48b6c78b709246ed646c7938780751b109e0dc), thanks for the pointer @sharkinsspatial and the amazing docs @kylebarron!

virtualizarr/readers/hdf/hdf.py
```diff
@@ -152,20 +154,42 @@ def _construct_manifest_group(
...
+    @overload
     @staticmethod
     def _create_manifest_store(
         filepath: str,
         *,
         prefix: str,
-        store: ObjectStore,
+        store: ObjectStore | None = None,
+        group: str | None = None,
+    ) -> ManifestStore: ...
```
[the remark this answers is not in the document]
maxrjones: I'd rather address this comment separately if that's alright to keep this PR isolated to supporting default store instances

maxrjones:
I think this PR is ready for a round of reviews if anyone has some extra time 🙏

TomNicholas (approved):

> FYI I want to revise this to follow the same design as https://docs.rs/datafusion/latest/datafusion/datasource/object_store/trait.ObjectStoreRegistry.html

Cool idea. Nice to make that connection.

The main thing that's missing from this PR is documentation of how to instantiate `ManifestStore` in different scenarios. That could either be done now (perhaps as examples in the `ManifestStore.__init__` docstring) or it could be left until later. We are building up technical debt in this refactor in the sense that we will need to go back through and document how everything now works.

virtualizarr/manifests/store.py
```diff
@@ -137,34 +133,92 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
     return tuple(int(ind) for ind in parts[1].split(chunk_key_encoding))
 
 
-def find_matching_store(stores: StoreDict, request_key: str) -> StoreRequest:
-    """
-    Find the matching store based on the store keys and the beginning of the URI strings,
-    to fetch data from the appropriately configured ObjectStore.
+def _find_bucket_region(bucket_name: str) -> str:
```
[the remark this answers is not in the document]
maxrjones: What is your main source for understanding private methods? I've understood them to indicate to external users not to rely on the functions, but that it's fine to use inside the codebase. But, I've never checked my understanding against PEPs tbh.

.pre-commit-config.yaml
```diff
@@ -22,3 +22,13 @@ repos:
...
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.15.0
+    hooks:
+      - id: mypy
+        files: src|tests
+        additional_dependencies:
+          # Package dependencies
+          - obstore>=0.5.1
+          # Tests
+          - pytest
```
kylebarron, line 34: I don't love putting mypy in pre-commit because pre-commit by default doesn't have access to your dependencies. So either it won't type-check against your dependencies or you have to maintain a second declaration of all your (typed) dependencies here.

maxrjones:
> The main thing that's missing from this PR is documentation of how to instantiate ManifestStore in different scenarios. That could either be done now (perhaps as examples in the ManifestStore.__init__ docstring) or it could be left until later. We are building up technical debt in this refactor in the sense that we will need to go back through and document how everything now works.

sounds good, I can do this now.

virtualizarr/manifests/store.py
```diff
@@ -137,34 +133,92 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
+    import requests
 
-    Parameters:
-    -----------
-    stores : StoreDict
-        A dictionary with URI prefixes for different stores as keys
-    request_key : str
-        A string to match against the dictionary keys
+    resp = requests.head(f"https://{bucket_name}.s3.amazonaws.com")
```
kylebarron, line 139: Not sure if you're calling this from a sync or async context, but if it's an async context you should make this an async request.

virtualizarr/manifests/store.py
```diff
@@ -137,34 +133,92 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
-            return StoreRequest(store=stores[key], key=parsed_key.path)
-    # if no match is found, raise an error
-    raise ValueError(
-        f"Expected the one of stores.keys() to match the data prefix, got {stores.keys()} and {request_key}"
-    )
+
+    _stores: dict[str, ObjectStore]
+
+    @classmethod
+    def __init__(self, stores: dict | None = None):
```
kylebarron, line 176:
```suggestion
    def __init__(self, stores: dict[str, ObjectStore] | None = None):
```

virtualizarr/manifests/store.py
```diff
@@ -137,34 +133,92 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
+        --------
+        StoreRequest
+        """
+        # Check each key to see if it's a prefix of the uri_string
+        for prefix, store in self._stores.items():
+            if url.startswith(prefix):
+                # Return an existing configured store and parsed request path
+                return store
+        # Use anonymous default store if not in pre-configured stores
+        prefix, store = _default_object_store(url)
```
kylebarron, line 217: Another option, if you expect the user to have boto3 in their environment, is to use the default boto3 credentials (via the boto3 credential provider), which will let them access buckets through their default profile automatically.

But it's not great to have different functionality depending on whether boto3 is installed.

virtualizarr/manifests/store.py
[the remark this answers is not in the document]
TomNicholas: I think this is a very fuzzy thing in python in general. But the (undocumented) convention that I (and I think quite a few others) use is:

- `_` - preceded variables are only used within that module, liable to change at any time
- non-`_`-preceded functions/classes are treated as relatively stable by the rest of the codebase, but not for public use (i.e. they are stable developer API)
- anything listed in the API docs is fully public API, potentially with backwards-compatibility guarantees.

Tests shouldn't really call anything with an underscore, and ideally should only call fully public API. But we're very lax about this in VirtualiZarr.

But again this is just a nit, not important.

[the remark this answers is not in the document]
TomNicholas: I have no idea if that's encoded in any PEP, but it's roughly a description of what xarray does for example.

virtualizarr/manifests/store.py
```diff
@@ -137,34 +133,92 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
+        self._stores = _sort_stores_by_prefix_length(stores)
+
+    def register_store(self, url: str, store: ObjectStore):
+        """
+        Register a store using the given url
+
+        If a store with the same key existed before, it is replaced
+        """
+        self._stores[url] = store
+        self._stores = _sort_stores_by_prefix_length(self._stores)
```
kylebarron, line 190: This means that every time you insert a store, you sort the full list of stores..?

I think having a sequence as an internal data structure makes sense when the user passes in a list, because they can control the order of stores within that list.

But when the user adds and removes stores via e.g. `register_store`, I think a HashMap makes more sense under the hood (and is closer to what `DataFusionRegistry` is doing. It uses `dashmap`, a concurrent hashmap)

virtualizarr/manifests/store.py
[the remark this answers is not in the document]
maxrjones: thats for laying this out! I've found it a bit confusing, so I added a doc for other devs who aren't as familiar with xarray's code standards - https://github.com/zarr-developers/VirtualiZarr/pull/552.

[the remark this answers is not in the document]
TomNicholas: Good idea, thanks!

virtualizarr/manifests/store.py
maxrjones: thanks, we can keep this in mind if needed later on

.pre-commit-config.yaml
maxrjones: I removed this from pre-commit

maxrjones:
@TomNicholas I'm actually going to use your approval to merge this an add the additional docs separately as a new "How to build a reader" section of the contrib guide.

TomNicholas:
> add the additional docs separately as a new "How to build a reader" section of the contrib guide.

This is actually tracked already in #452 FYI

PR #562, opened 2025-04-23
WIP: Preferentially use to_dict for serializing codecs

maxrjones:
This breaks kerchunk's use of V2 metadata, but may help pickle V3 metadata. I'll see if both could work.


maxrjones:
This won't work

Issue #561, opened 2025-04-23
Fragility of url auto-parsing logic


TomNicholas:
We have code to automatically interpret a URI as being a local file, S3-compatible store, or HTTPStore using Obstore. 

https://github.com/zarr-developers/VirtualiZarr/blob/0bf830e63bc5bdedd87cd824d064071600d5b44a/virtualizarr/manifests/store.py#L142

But it feels a bit fragile and not really core to this package.

Can we just delegate the entire logic to another package? E.g. to `obstore` via https://github.com/developmentseed/obstore/blob/95e586fc6c8c9425fb682fdb9aedb16e0b0a2935/obstore/python/obstore/store.py#L656-L765

Related also to the API choice in #553, as we could choose to just make this the user's problem and stop guessing what they gave us.

cc @maxrjones

maxrjones:
> Can we just delegate the entire logic to another package? E.g. to obstore via [developmentseed/obstore@95e586f/obstore/python/obstore/store.py#L656-L765](https://github.com/developmentseed/obstore/blob/95e586fc6c8c9425fb682fdb9aedb16e0b0a2935/obstore/python/obstore/store.py#L656-L765)

I think this is the right call. That functionality wasn't in obstore when I started ManifestStore IIRC but now that it's released we should definitely use it.

> Related also to the API choice in https://github.com/zarr-developers/VirtualiZarr/issues/553, as we could choose to just make this the user's problem and stop guessing what they gave us.

I was originally thinking we should make it the user's problem, but have come to recognize that people really rely on and expect auto-determination for storage interfaces and think we should design the API with that reality in mind.

TomNicholas:
> That functionality wasn't in obstore when I started ManifestStore IIRC but now that it's released we should definitely use it.

Ohh cool.

> There are likely a large number of users relying on auth restricted HTTP endpoints and we won't have support for those via obstore currently

Uhh that does seem like a problem. I wonder if we can have some advanced API for opting in to using `fsspec` instead of `obstore`, but default to `obstore`

kylebarron:
[...]

PR #564, opened 2025-04-23
Allow passing config to default_object_store

maxrjones:


virtualizarr/manifests/store.py
```diff
@@ -139,25 +149,34 @@ def _find_bucket_region(bucket_name: str) -> str:
     return region
 
 
-def default_object_store(filepath: str) -> ObjectStore:
+def default_object_store(
+    filepath: str, storage_config: ObjectStoreOptions | None = None
+) -> ObjectStore:
     import obstore as obs
```
maxrjones, line 155: I'd like to replace this function with `obs.store.from_url()` but the upstream version doesn't seem to auto-infer the region. Am I missing anything @kylebarron? xref https://github.com/zarr-developers/VirtualiZarr/issues/561

maxrjones:
@TomNicholas I need to figure out how to handle the different kwargs for kerchunk vs virtualizarr HDF backends in the tests, but I think this is ready for you to try

virtualizarr/manifests/store.py
kylebarron: It's correct that `obstore.store.from_url` does not auto-infer the S3 region. This is an artifact of AWS-hosted S3 requiring the region but non-AWS-hosted S3-compatible stores not requiring the region. E.g. it [supports](https://github.com/apache/arrow-rs-object-store/blob/8d3d8319dbaae7c07d7bd52339a0d0cca033e9ac/src/parse.rs#L129-L130) `r2.cloudflarestorage.com` -style urls, but those don't have a region. So it's up to the user to pass in the region if required.

maxrjones:
The tests are harder to sort out than I expected because even a obstore based reader with use fsspec when automatically determining the filetype. I'm going to pivot to developer docs now which will help us design storage option configuration. In the meantime @TomNicholas  you probably want to specify the `filetype`.

virtualizarr/manifests/store.py
kylebarron: Here's something I whipped together, which is a quick port of [this file](https://github.com/apache/arrow-rs-object-store/blob/8d3d8319dbaae7c07d7bd52339a0d0cca033e9ac/src/parse.rs#L4)

```py
from __future__ import annotations

from enum import Enum

from obstore.store import S3Store


class ObjectStoreScheme(Enum):
    AZURE = "azure"
    FILE = "file"
    S3 = "s3"
    S3Like = "s3like"
    MEMORY = "memory"
    HTTP = "http"


def create_store(url_str: str, config: dict) -> ObjectStore:
    scheme, bucket, path = parse_url(url_str)
    if scheme == ObjectStoreScheme.S3:
        region = infer_region(...)
        return S3Store(...)
    elif scheme == ObjectStoreScheme.S3Like:
        # Don't infer region
        return S3Store(...)
    elif scheme == ObjectStoreScheme.AZURE:
        return AzureStore(...)
    elif scheme == ObjectStoreScheme.FILE:
        return LocalStore(...)
    elif scheme == ObjectStoreScheme.MEMORY:
        return MemoryStore(...)
    elif scheme == ObjectStoreScheme.HTTP:
        return HttpStore(...)
    else:
        raise ValueError(f"Unsupported URL scheme: {scheme}")


def parse_url(url_str: str) -> tuple[ObjectStoreScheme, str | None, str]:
    # scheme, bucket, path
    url = urlparse(url_str)
    if url.scheme == "file":
        return ObjectStoreScheme.FILE, None, url.path

    if url.scheme == "memory":
        return ObjectStoreScheme.MEMORY, None, url.path

    if url.scheme in ["s3", "s3a"]:
        assert url.netloc, f"Expected bucket in s3:// url, got: {url_str}"
        return ObjectStoreScheme.S3, url.netloc, url.path

    if url.scheme == "gs":
        assert url.netloc, f"Expected bucket in gs:// url, got: {url_str}"
        return ObjectStoreScheme.S3Like, url.netloc, url.path

    if url.scheme in ["az", "adl", "azure", "abfs", "abfss"]:
        assert url.netloc, f"Expected bucket in azure url, got: {url_str}"
        return ObjectStoreScheme.AZURE, url.netloc, url.path

    if url.scheme == "http":
        return ObjectStoreScheme.HTTP, None, url.path

    if url.scheme == "https":
        if url.netloc.endswith("amazonaws.com"):
            if url.netloc.startswith("s3"):
                region = url.netloc.split(".", maxsplit=2)[1]
                # TODO: return region from this fn
                bucket, path = url.path.removeprefix("/").split("/", maxsplit=1)
                return ObjectStoreScheme.S3, bucket, path
            else:
                return ObjectStoreScheme.S3, None, url.path
        if url.netloc.endswith("r2.cloudflarestorage.com"):
            return ObjectStoreScheme.S3Like, None, url.path
        if url.netloc.endswith("blob.core.windows.net") or url.netloc.endswith(
            "dfs.core.windows.net"
        ):
            return ObjectStoreScheme.AZURE, None, url.path

        return ObjectStoreScheme.HTTP, None, url.path

    raise ValueError("Unrecognized url")
```

Perhaps you'd want `create_store` to take in separate config dicts for better typing and so you can define configs for multiple different stores at once. Something like

```py
def create_store(
    url_str: str,
    s3_config: S3Config | None = None,
    azure_config: AzureConfig | None = None,
    gcs_config: GCSConfig | None = None,
) -> ObjectStore: ...
```

Issue #553, opened 2025-04-18
Function definition for `open_virtual_dataset` in next major release


maxrjones:
We discussed a pre-release of VirtualiZarr 2.0 within the next week. I wanted to raise some discussion points for the public `open_virtual_dataset` function definition.

While not included in the docstring, the `backend` parameter has been added to the function def since the last release. This means there are currently two ways to set the backend, `filetype` and `backend`. In the zen of Python, I think there should just be one way to specify a backend. Could we consolidate these parameters?

The internal `_create_manifest_store(...)` methods include a `store` parameter that can be used to virtualize objects requiring credentials. There's no equivalent option in `open_virtual_dataset`. Could both `reader_options` be consolidated with `virtual_backend_kwargs` to avoid the use of different function signatures depending on whether a backend uses Fsspec or Obstore.

This is a nit, but filepath is really filesystem oriented language and I think VirtualiZarr is about providing cloud native access patterns, such that URL would be more appropriate.

maxrjones:
I think any of these changes could be made with backwards compatibility + a warning about the new syntax

maxrjones:
I'm taking a "speak now or forever hold your peace" approach to this issue and hope people don't mind 🤞. I would prefer a `keep_variables` argument over a `drop_variables` argument (similar to this [request in xarray](https://github.com/pydata/xarray/issues/1754)). I understand that the original design was to mimic `open_dataset`. The tradeoffs here involve users needing to know which variables they want to drop and many readers first virtualizing all variables, then dropping some rather than the more efficient path of only virtualizing variables that will be kept.

TomNicholas:
> I think there should just be one way to specify a backend. Could we consolidate these parameters?

I agree that this has become messy. Could consolidating everything just look something like this?

```python
from typing import Callable, TypeAlias

from virtualizarr.manifests import ManifestStore


T_VirtualReader: TypeAlias = Callable[[str], ManifestStore]


def open_virtual_dataset(
    uri: str,
    reader: str | T_VirtualReader = None,
    reader_options: dict = {},
) -> xr.Dataset:
```

whose use would look like

```python
vds = open_virtual_dataset('file.tif', reader='tiff')
```

```python
from virtualizarr.readers import tiff_reader

vds = open_virtual_dataset('file.tif', reader=tiff_reader)
```

```python
from other.package import third_party_reader

vds = open_virtual_dataset('file.third', reader=third_party_reader)
```

[...]

PR #568, opened 2025-04-25
Propose design doc and roadmap for VirtualiZarr 2.0

maxrjones:
There's a lot of interconnected technical design issues open. This design doc should help us discuss solutions that impact multiple issues.

The key differences to the current design is using protocols more than ABCs, defining a "Reader" as a protocol required by VirtualiZarr that can be used by backends and the ManifestStore, and orienting backends around ManifestStore creation rather than Dataset creation.

Just a starting point building off ideas shared by @kylebarron and @sharkinsspatial.

### DIFF
--- a/design_notes/v2.md
+++ b/design_notes/v2.md
@@ -52,7 +52,7 @@ class Parser(protocol):
     def __call__(
         self,
         file_url: str,
-        object_reader: ObjectStore,
+        object_store: ObjectStore,
         *,
     ) → ManifestStore:
         ...
@@ -68,7 +68,7 @@ class Parser(protocol):
 ```python
  def open_virtual_dataset(
     file_url: str,
-    object_reader: ObjectStore,
+    object_store: ObjectStore,
     parser: Parser,
     *,
     drop_variables: Iterable[str] | None = None,
@@ -82,7 +82,7 @@ class Parser(protocol):
     """
     manifest_store = parser(
         file_url = file_url,
-        object_reader = object_reader,
+        object_store = object_store,
     )
     vds = manifest_store.to_virtual_dataset(
         loadable_variables=loadable_variables,
@@ -98,7 +98,7 @@ class Parser(protocol):
 ```python
  def open_virtual_mfdataset(
     paths: str | os.PathLike | Sequence[str | os.PathLike] | "NestedSequence[str | os.PathLike]",
-    object_reader: ObjectStore,
+    object_store: ObjectStore,
     parser: Parser,
     *,
     concat_dim: (


maxrjones:
@kylebarron @chuckwondo @d-v-b I wasn't able to request reviews from you all but it'd be great to get your thoughts on this if you have time

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Protocols
+
+> **Note**: Protocols over ABCs would allow third party backends to be less tightly coupled to VirtualiZarr
+> **Question:**: How do we simplify these protocol definitions using obspec?
+
+`FileReader` - read bytes from files on object storage or disk
+`VirtualZarrBackend` - construct a ManifestStore from a file
+
+```python!=
```
kylebarron, line 39: What does `!=` do?

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+    ) -> Bytes :
+        ...
+
+    async get(
+        path: str,
+        *,
+        options: int | None = None,
+        length: int | None = None
+    ) -> GetResult :
+        ...
```
kylebarron, line 69: It seems strange to me that you'd have a mix of sync and async methods in a single reader backend. I think you might want to have a `FileReader` and an `AsyncFileReader` as separate protocols so that you can swap between implementations of each of them.

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+    def open_virtual_store(
+        filepath: str,
+        *,
+        virtual_backend_kwargs: dict | None = None,
+        reader_options: ObspecReaderOptions | dict | None = None,
+    ) → VirtualStore:
+        ...
+```
+
+> **Question:**: How do we define this protocol to allow for strict typing for some readers (e.g., obstore-backed) and not others (e.g., fsspec-backed)?
```
kylebarron, line 90: The protocol will always be strongly typed; it just means in the fsspec case you have to manually verify the constraints, that the fsspec wrapper implements your protocol correctly using known fsspec methods.

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+        options: int | None = None,
+        length: int | None = None
+    ) -> GetResult :
+        ...
+
+```
+
+
+```python!=
+class VirtualZarrBackend(protocol):
```
kylebarron, line 75: Not everything will be a protocol. Only the "common interfaces" should be defined as protocols. 

E.g. if you only ever have one concrete implementation of "opening a virtual store", then `open_virtual_store` can be a top-level function, not its own protocol.

Here, `open_virtual_store` might _use_ the above FileReader. So you'd instead have something like `open_virtual_store(path: str, *, reader: FileReader)`.

kylebarron: A crucial concept here is that the end user needs to be able to pass in the protocol as a parameter in the top-level API somehow. The idea of the protocol is to make it easier to be _externally-implementable_. If you have a "closed" selection of types, where you know the only supported backends will be ones implemented in virtualizarr, then you don't necessarily need to go through the protocol definitions.

It's maybe ok if there's some default behavior that constructs a default filereader, but it needs to be overridden.

But right now you define `FileReader` as a protocol but never define it as a parameter in this document. What uses the `FileReader`? `open_virtual_store`?

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+> **Note**: Could ObspectReaderOptions by typed as follows:
+
+```python!=
+ObspecReaderOptions(TypedDict):
+    s3_config: S3Config | None = None,
+    azure_config: AzureConfig | None = None,
+    gcs_config: GCSConfig | None = None,
+    client_options: NotRequired[ClientConfig]
+    retry_config: NotRequired[RetryConfig]
+    credential_provider: NotRequired[Callable]
```
kylebarron, line 101: Similar to my [comment above](https://github.com/zarr-developers/VirtualiZarr/pull/568/files#r2060704770), you call this "obspec" reader options, but this is hard-coded to _obstore_ — a specific implementation of the obspec definitions.

Think in fsspec about the difference between `AbstractFileSystem` and `S3FileSystem`. There's no concept of `AbstractFileSystemOptions` because the options aren't a property of the _core base class_, the options are a property of the concrete s3 implementation.

design_notes/v2.md
maxrjones: In hackmd (where I drafted this), it adds line numbers and auto wrapping. In GitHub, it might prevent the Python code from correctly rendering 🤦‍♂️

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+    filepath: str,
+    *,
+    cache: bool = False,
+    drop_variables: Iterable[str] | None = None,
+    loadable_variables: Iterable[str] | None = None,
+    decode_times: bool | None = None,
+    cftime_variables: Iterable[str] | None = None,
+    indexes: Mapping[str, Index] | None = None,
+    virtual_backend_kwargs: dict | None = None,
+    reader_options: ObspecReaderOptions | dict | None = None,
```
kylebarron, line 121: This whole idea of passing in the `reader_options` is something we should avoid IMO, because it connects the public API into one specific implementation.

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+`FileReader` - read bytes from files on object storage or disk
+`VirtualZarrBackend` - construct a ManifestStore from a file
+
+```python!=
+class FileReader(protocol):
+    """
+    Protocol for file operations for downstream libraries
+    """
+
+    def read(self, size: int, /) -> bytes
```
chuckwondo, line 45:
```suggestion
    def read(self, size: int, /) -> bytes:
```

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+    async get_range_async(
+        path: str,
+        *,
+        start: int,
+        end: int | None = None,
+        length: int | None = None
+    ) -> Bytes :
+        ...
+
+    async get(
```
chuckwondo, line 63: `get_async`?

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+        path: str,
```
chuckwondo, line 64: missing `self`?

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+    """
+    Protocol for creating a virtual zarr dataset from a file
+    """
+
+    @staticmethod
+    def open_virtual_store(
+        filepath: str,
+        *,
+        virtual_backend_kwargs: dict | None = None,
+        reader_options: ObspecReaderOptions | dict | None = None,
```
chuckwondo, line 85: Perhaps we should allow users to optionally pass in their own reader. I'm thinking about the recent thread regarding the ability to support caching. If we simply allow the user to supply a caching reader through, no code changes would be required.

design_notes/v2.md
chuckwondo, line 90:
```suggestion
> **Question:** How do we define this protocol to allow for strict typing for some readers (e.g., obstore-backed) and not others (e.g., fsspec-backed)?
```

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+    ) → VirtualStore:
+        ...
+```
+
+> **Question:**: How do we define this protocol to allow for strict typing for some readers (e.g., obstore-backed) and not others (e.g., fsspec-backed)?
+
+> **Note**: Could ObspectReaderOptions by typed as follows:
```
chuckwondo, line 92:
```suggestion
> **Note:** Could ObspectReaderOptions be typed as follows?
```

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+
+```python!=
+ObspecReaderOptions(TypedDict):
+    s3_config: S3Config | None = None,
+    azure_config: AzureConfig | None = None,
+    gcs_config: GCSConfig | None = None,
```
chuckwondo, line 98:
```suggestion
ObspecReaderOptions(TypedDict):
    s3_config: S3Config | None = None
    azure_config: AzureConfig | None = None
    gcs_config: GCSConfig | None = None
```

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+    client_options: NotRequired[ClientConfig]
+    retry_config: NotRequired[RetryConfig]
+    credential_provider: NotRequired[Callable]
+```
+## Top-level API functions
+
+`open_virtual_dataset` - construct a Xarray Dataset with NumPy and/or ManifestArray-backed variables from a file
+`open_virtual_mfdataset` - construct a Xarray Dataset with NumPy and/or ManifestArray-backed variables from many files
```
chuckwondo, line 106:
```suggestion
- `open_virtual_dataset` - construct a Xarray Dataset with NumPy and/or ManifestArray-backed variables from a file
- `open_virtual_mfdataset` - construct a Xarray Dataset with NumPy and/or ManifestArray-backed variables from many files
```

design_notes/v2.md
chuckwondo: Yes, I believe this ties in with my comment about allowing the user to pass in their own reader, which would also mean that the `cache` parameter goes away, as the user would simply pass in a caching reader, it they like.

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+        reader_options=reader_options
+    )
+    ds = manifest_store.to_virtual_dataset(
+        loadable_variables=loadable_variables,
+        decode_times=decode_times,
+        indexes=indexes,
+    )
+    if cache:
+        ds = ds.virtualize.rename_paths(orig_fp)
+    return vds
```
chuckwondo, line 145:
```suggestion
    return ds
```

design_notes/v2.md
```diff
@@ -0,0 +1,173 @@
...
+    """
+    Backend for using any Kerchunk readers in VirtualiZarr
+    """
+
+    @staticmethod
+    def open_virtual_store(
+        filepath: str,
+        *,
+        virtual_backend_kwargs: dict | None = None,
+        reader_options: ObspecReaderOptions | dict | None = None,
```
chuckwondo, line 163: Again, we likely want to allow the user to provider their own reader directly.

design_notes/v2.md
maxrjones: I like the idea of a user specified "reader" rather than the `reader_options` parameter and the removal of the `cache` parameter.

That said, I am not sure how we'd define what a `reader` is acknowledging the oddity above of having `async` and `sync` in a single protocol (https://github.com/zarr-developers/VirtualiZarr/pull/568#discussion_r2060688872) and that different backends require different methods. For example, h5py-based backends require `read`, `seek` and `tell`. But others might not (e.g., `async_tiff`, maybe?). The ManifestStore used for all loadable variables requires only `get_range_async `. Would the "reader" be used for both, or would we have a "object_reader" and a "file_reader" as separate parameters, or some other solution?

chuckwondo: @maxrjones, I'm not familiar enough with all of this, but based upon the ["caching tweak" you recently created](https://github.com/zarr-developers/VirtualiZarr/blob/d6d6c25e76ac412555b06f91eec9d8548fc5f72d/virtualizarr/readers/hdf/hdf.py), I suspect what we might actually want to allow the user to pass in their own _store_ rather than their own _reader_, and that store could be a "caching" store.

I'm only basing this suggestion off or my recent gleaning of that "tweak" you created, so I could be missing something in how all these pieces fit together, so if my suggestion doesn't make sense, please add some information to help me better grasp everything.

design_notes/v2.md
```diff
@@ -0,0 +1,176 @@
+# VirtualiZarr V2 architecture discussion
+
+## Goals
+
+- Extensible and easily understandable entry-point system for new backends
+- Modular code base that uses Obspec for the object storage interface abstraction, Zarr as the core data model, and Xarray for the concatenation and merging of Zarr stores
+
+## Related issues
+
+- [Fragility of url auto-parsing logic](https://github.com/zarr-developers/VirtualiZarr/issues/561)
```
TomNicholas, line 10:
```suggestion
- [Fragility of url auto-parsing logic](https://github.com/zarr-developers/VirtualiZarr/issues/561)
- [Use obstore / obspec for globbing in `open_virtual_mfdataset`](https://github.com/zarr-developers/VirtualiZarr/issues/569)
```

design_notes/v2.md
```diff
@@ -0,0 +1,176 @@
...
+- [Support passing configuration options to default_object_store](https://github.com/zarr-developers/VirtualiZarr/issues/559)
+- [Redefine Virtual Readers as func(filepath) -> ManifestStore](https://github.com/zarr-developers/VirtualiZarr/issues/498)
+- [Make xarray an optional dependency?](https://github.com/zarr-developers/VirtualiZarr/issues/521)
+- [Re-implement loadable_variables using ManifestStore ](https://github.com/zarr-developers/VirtualiZarr/issues/473)
+- [Should we try to use Obstore for reading files?](https://github.com/zarr-developers/VirtualiZarr/issues/476)
+- [Challenges with earthaccess.open_virtual_mfdataset for grouped netCDFs with nested structure](https://github.com/zarr-developers/VirtualiZarr/issues/487)
+- [Docs on how to write a custom reader](https://github.com/zarr-developers/VirtualiZarr/issues/452)
+- [Make readers pluggable via entrypoint system](https://github.com/zarr-developers/VirtualiZarr/issues/245)
+- [Make Manifests' in-memory reference structure pluggable](https://github.com/zarr-developers/VirtualiZarr/issues/246)
```
TomNicholas, line 20: I think this one is unrelated to all the others, and doesn't need to be discussed in this doc.

design_notes/v2.md
```diff
@@ -0,0 +1,176 @@
...
+
+> **Note:** Protocols over ABCs would allow third party backends to be less tightly coupled to VirtualiZarr
+>
+> **Question:** How do we simplify these protocol definitions using obspec?
+
+- `FileReader`: read bytes from files on object storage or disk
+- `VirtualZarrBackend`: construct a ManifestStore from a file
+
+```python
+class FileReader(Protocol):
```
TomNicholas, line 41: This seems like a useful protocol, but it shouldn't live in VirtualiZarr, as it's a much more generally-useful thing. Xarray could use this, for example.

design_notes/v2.md
TomNicholas: @kylebarron for `VirtualZarrBackend` at least the motivating context is here: https://github.com/zarr-developers/VirtualiZarr/issues/498. I agree though that it doesn't need to be very complicated.

design_notes/v2.md
```diff
@@ -0,0 +1,176 @@
...
+
+
+```python
+class VirtualZarrBackend(protocol):
+    """
+    Protocol for creating a virtual zarr dataset from a file
+    """
+
+    @staticmethod
+    def open_virtual_store(
```
TomNicholas, line 84: A class which only has one static method might as well just be a function. Which takes us back to the suggestion in #498. Let's not over-complicate this.

design_notes/v2.md
TomNicholas: > what we might actually want to allow the user to pass in their own store rather than their own reader, and that store could be a "caching" store.

So the `CachingStore` would simply cache the entire file if any part of it is asked for? That sounds neat.

design_notes/v2.md
```diff
@@ -0,0 +1,176 @@
...
+    cftime_variables: Iterable[str] | None = None,
+    indexes: Mapping[str, Index] | None = None,
+    virtual_backend_kwargs: dict | None = None,
+    reader_options: ObspecReaderOptions | dict | None = None,
+    backend: type[VirtualBackend] | None = None
+) → Dataset
+    """
+    Method for creating virtual Xarray dataset, only implemented in top level API
+    """
+    # Guess backend if not provided based on filepath or magic bytes
```
TomNicholas, line 130: We may want to just get rid of the whole guessing-the-backend thing. It's complicated for little benefit IMO. Who is realistically using this library without knowing what archival format they are trying to virtualize?

design_notes/v2.md
```diff
@@ -0,0 +1,176 @@
...
+    if cache:
+        orig_fp = filepath
+        filepath = cache_file(orig_fp, reader_options)
+    if not backend:
+        backend = guess_backend(filepath, reader_options)
+    backend.open_virtual_store(
```
TomNicholas, line 136: `backend.open_virtual_store` should just be a function IMO (see other comments above). There's no state to keep track of on the `backend` object, and it only needs one method, so it doesn't need to be a class.

design_notes/v2.md
```diff
@@ -0,0 +1,176 @@
...
+        indexes=indexes,
+    )
+    if cache:
+        ds = ds.virtualize.rename_paths(orig_fp)
+    return ds
+```
+
+## Backends
+
+> **Question**: Is there a reason to have a backend per kerchunk reader rather than a single backend that could be used for all of them?
```
TomNicholas, line 153: Copying my DM reply to Max about this:

> 1. More kerchunk readers could be added at any time, by anyone, even outside of kerchunk
> 2. They read different filetypes
> 3. They potentially accept different kwargs
> 
> The fact those readers call kerchunk should just be an implementation detail

design_notes/v2.md
```diff
@@ -0,0 +1,176 @@
...
+
+```Python
+class KerchunkVirtualBackend():
```
TomNicholas, line 156: I don't see why we need this.

TomNicholas (commented):
Thanks for taking the initiative here @maxrjones ! It's clear from this that there was far too much intention just scattered around various issues.

design_notes/v2.md
```diff
@@ -0,0 +1,177 @@
...
+- [Should we try to use Obstore for reading files?](https://github.com/zarr-developers/VirtualiZarr/issues/476)
+- [Challenges with earthaccess.open_virtual_mfdataset for grouped netCDFs with nested structure](https://github.com/zarr-developers/VirtualiZarr/issues/487)
+- [Docs on how to write a custom reader](https://github.com/zarr-developers/VirtualiZarr/issues/452)
+- [Make readers pluggable via entrypoint system](https://github.com/zarr-developers/VirtualiZarr/issues/245)
+- [Make Manifests' in-memory reference structure pluggable](https://github.com/zarr-developers/VirtualiZarr/issues/246)
+- [Architecture diagram](https://github.com/zarr-developers/VirtualiZarr/issues/225)
+
+## Data structures
+
+- `ChunkManifest`: Contains paths, chunk byte offsets, lengths
```
abarciauskas-bgse, line 26:
```suggestion
- `ChunkManifest`: An object defining the paths, chunk byte offsets, lengths of one virtual array.
```

Perhaps we could define a virtual array as, not a data structure but a concept, where the virtual array is the representation of an array not by its actual values but in where to read those values from.

design_notes/v2.md
```diff
@@ -0,0 +1,177 @@
...
+- `ManifestArray`: Contains one ChunkManifest and Zarr V3 ArrayMetadata
```
abarciauskas-bgse, line 27: A chunk manifest and its assic
```suggestion
- `ManifestArray`: A chunk manifest and its associated Zarr V3 ArrayMetadata
```

design_notes/v2.md
```diff
@@ -0,0 +1,177 @@
...
+- `ManifestGroup`: Contains group metadata and ManifestGroup and/or ManifestArrays
+- `ManifestStore`: Contains a ManifestGroup and a FileReader
```
abarciauskas-bgse, line 29: Should we define a FileReader? And ObjectStoreRegistry?

abarciauskas-bgse:
Thanks for putting this together @maxrjones it is super helpful to have these conversations synthesized into this design doc, especially with the associated issues linked. Take my comments with a grain of salt as I haven't been involved directly in these discussions.

design_notes/v2.md
maxrjones: how do we standardize the function signatures on backends if it's just a function? I only know how to do this using ABCs or protocols TBH, which is why I left it as a protocol.

design_notes/v2.md
```diff
@@ -0,0 +1,96 @@
...
+- `VirtualZarrBackend`: construct a ManifestStore from a file
+
+```python
+from obspec import GetAsync, GetRangeAsync, Read, Seek, Tell
+
+class ObjectReader(GetAsync, GetRangeAsync):
+    ....
+
+class ReadableFile(Read, Seek, Tell):
+    ....
```
maxrjones, line 43: @kylebarron could `obspec` include protocols for `Readable/BufferedFile`? Or is there a different source for these definitions?

design_notes/v2.md
chuckwondo: Define a [callback protocol](https://typing.python.org/en/latest/spec/callables.html#callback-protocols).

maxrjones: Thanks so much, Chuck!

design_notes/v2.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+```python
+class Backend(protocol):
+    """
+    Callback protocol for creating a virtual zarr store from a file
+    """
+    def __call__(
+        url_string: str,
+        object_reader: ObjectReader,
```
maxrjones, line 57: We could make object_reader optional if there are no loadable variables.

design_notes/v2.md
```diff
@@ -0,0 +1,98 @@
...
+## Top-level API functions
+
+- `open_virtual_dataset` - construct a Xarray Dataset with NumPy and/or ManifestArray-backed variables from a file
+- `open_virtual_mfdataset` - construct a Xarray Dataset with NumPy and/or ManifestArray-backed variables from many files
+
+> **NOTE**: Here the construction of a virtual dataset from a manifest store would happen in the top-level function rather than the backends.
+
+```python
+ def open_virtual_dataset(
+    readable_file: ReadableFile,
```
sharkinsspatial, line 73: @maxrjones I just started playing with a PR following this proposed design and the discussion from https://github.com/zarr-developers/VirtualiZarr/issues/553.  I think one additional issue is that our `ChunkManifest` generation requires the full path string to the file being virtualized.  Looking through the `obstore` code I don't think we a way to reconstruct this full path in a generic way from either a `ReadableFile` or an `ObjectReader` so we may also need to pass the explicit path string in addition (even though it seems duplicative 😆).  I'll way for @kylebarron to confirm this is the case.

kylebarron: By "full path" you mean not just the path inside the bucket but also the protocol, bucket, and prefix? So you need `s3://bucket/path` or `abfs://container/path`?

The `ReadableFile` has a way to get the path not including the bucket and store prefix. https://developmentseed.org/obstore/latest/api/file/#obstore.ReadableFile.meta

And then the store allows you to get the `prefix` that was originally provided.

So you'd just need to keep a reference onto the store and infer the prefix yourself.

TomNicholas: > By "full path" you mean not just the path inside the bucket but also the protocol, bucket, and prefix?

Yes - the `ManifestArray` needs to hold all the path information that Icechunk will eventually need to read back from that object.

This is important - should the reader function signature actually be something more like:

```python
def reader(bucket_config: StoreConfig, *files: ReadableFile) -> ManifestStore:
    ...
```

design_notes/v2.md
TomNicholas: > We could make object_reader optional if there are no loadable variables.

Well then it would be optional in general because we don't know if there are any `loadable_variables` at `ManifestStore` construction time.

design_notes/v2.md
```diff
@@ -0,0 +1,98 @@
...
+- [Challenges with earthaccess.open_virtual_mfdataset for grouped netCDFs with nested structure](https://github.com/zarr-developers/VirtualiZarr/issues/487)
+- [Docs on how to write a custom reader](https://github.com/zarr-developers/VirtualiZarr/issues/452)
+- [Make readers pluggable via entrypoint system](https://github.com/zarr-developers/VirtualiZarr/issues/245)
+- [Architecture diagram](https://github.com/zarr-developers/VirtualiZarr/issues/225)
+
+## Data structures
+
+- `ChunkManifest`: An object defining the paths, chunk byte offsets, lengths of one virtual array.
+- `ManifestArray`: A chunk manifest and its associated Zarr V3 ArrayMetadata
+- `ManifestGroup`: ManifestGroups are a way to organizing child groups and arrays into a data hierarchy. Contains Zarr `GroupMetadata` and one or more ManifestGroups and/or ManifestArrays.
```
kylebarron, line 27:
```suggestion
- `ManifestGroup`: ManifestGroups are a way to organize child groups and arrays into a data hierarchy. Contains Zarr `GroupMetadata` and one or more ManifestGroups and/or ManifestArrays.
```

design_notes/v2.md
```diff
@@ -0,0 +1,98 @@
...
+- `ManifestStore`: A ManifestGroup and its associated ObjectReader (see protocol below).
+- `xarray.Dataset`: Contains ManifestArray-backed variables and (optionally) in-memory NumPy-backed variables.
+
+## Definitions
+
+- `ObjectReader` - An implementation of the `GetAsync` and `GetRangeAsync` protocols defined in obspec. `ObjectReader`s are used to load arrays using `ManifestStore`.
```
kylebarron, line 33: Does this also need `List` support?

And this says that it requires `GetAsync` and `GetRangeAsync`, but that implies that 1) all backends need to be async-capable and that 2) the `ObjectReader` itself will be async (or internally have an event loop that blocks on all the async calls)

It's annoying but unavoidable that async makes you deal with [colored functions](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/).

You should decide whether the required performance characteristics of Virtualizarr require exposing an async API to the user. If not, then maybe only use blocking calls here (`Get` and `GetRange` and `GetRanges`). If so, then you might need an `ObjectReader` and `AsyncObjectReader`.

design_notes/v2.md
```diff
@@ -0,0 +1,98 @@
...
+- `ReadableFile` - A reader that implements the `Read`, `Seek`, and `Tell` methods using a similar interface as a [Python BufferedReader](https://docs.python.org/3/library/io.html#io.BufferedReader). `ReadableFile`s are used by a `Backend` to construct a `ManifestStore`.
```
kylebarron, line 34: Why do you want to use a `ReadableFile`? Isn't this part of what we're trying to get away from in leaving fsspec? Having a stateful `ReadableFile` is less predictable than calling range requests directly.

design_notes/v2.md
```diff
@@ -0,0 +1,98 @@
...
+- `Backend` - A backend is a function that constructs a `ManifestStore` from a `ReadableFile`, `ObjectReader`, and optionally `kwargs` specified using `virtual_backend_kwargs`.
+
+
+## Protocols
+
+```python
+from obspec import GetAsync, GetRangeAsync, Read, Seek, Tell
```
kylebarron, line 41: `Read`, `Seek`, and `Tell` don't exist in Obspec, but you could define similar methods yourself.

design_notes/v2.md
maxrjones: Since both ffspec and obstore have ways to create a readable file object from a store and a file url, I think this could work:
```
    def __call__(
        file_url: str,
        object_reader: ObjectReader,
        *,
        virtual_backend_kwargs: dict | None = None,
    ) → ManifestStore:

> This is important - should the reader function signature actually be something more like:

This is pretty close to a previous edition - it's was challenging to support interoperability between different ObjectReaders.

File: design_notes/v2.md (hunk 31 above)
Comment (maxrjones) [reply design_notes/v2.md:33]: The async get range method is primarily used within `ManifestStore`'s implementation of Zarr's async Store API. So I think we definitely want an async version - https://github.com/zarr-developers/VirtualiZarr/blob/5646b675e37ded8d18037bf0aab9a0786650c46c/virtualizarr/manifests/store.py#L322-L328

(note I should change this to use `store.get_range_async` to allow interoperability following obspec).

I'll revise to add an `ObjectReader` and `AsyncObjectReader` with the `AsyncObjectReader` required by `ManifestStore` and the `ObjectReader` available for use in backends.

File: design_notes/v2.md (hunk 32 above)
Comment (maxrjones) [reply design_notes/v2.md:34]: As one example, the h5py library is used inside the hdf5 backend for VirtualiZarr and [operates on file-like objects](https://docs.h5py.org/en/stable/high/file.html#python-file-like-objects).

File: design_notes/v2.md [hunk 34]
```diff
@@ -0,0 +1,97 @@
...
+- `ManifestArray`: A chunk manifest and its associated Zarr V3 ArrayMetadata
+- `ManifestGroup`: ManifestGroups are a way to organize child groups and arrays into a data hierarchy. Contains Zarr `GroupMetadata` and one or more ManifestGroups and/or ManifestArrays.
+- `ManifestStore`: An implementation of Zarr's store API that contains a ManifestGroup and an AsyncObjectReader (see protocol below).
+- `xarray.Dataset`: Contains ManifestArray-backed variables and (optionally) in-memory NumPy-backed variables.
+
+## Definitions
+
+- `AsyncObjectReader` - An implementation of the `GetAsync` and `GetRangeAsync` protocols defined in `obspec`. `AsyncObjectReader`s are used to load arrays in a `ManifestStore`. `AsyncObjectReader`s may be used by a `Backend` to construct a `ManifestStore`.
+- `ReadableFile` - A reader that implements the `Read`, `Seek`, and `Tell` methods using a similar interface as a [Python BufferedReader](https://docs.python.org/3/library/io.html#io.BufferedReader). `ReadableFile`s may be used by a `Backend` to construct a `ManifestStore`. If required, the `Backend` should construct the `ReadableFile` from the `AsyncObjectReader` and file URL.
```
maxrjones, line 34: Maybe we should just take out all the `ReadableFile` parts and include guidance in https://virtualizarr.readthedocs.io/en/latest/custom_readers.html on creating a file object from an `AsyncObjectReader` and `file_url`.

design_notes/v2.md
```diff
@@ -0,0 +1,97 @@
...
+class ReadableFile(Read, Seek, Tell):
+    ...
+```
+
+```python
+class Backend(protocol):
+    """
+    Callback protocol for creating a virtual zarr store from a file
+    """
+    def __call__(
```
TomNicholas, line 54:
```suggestion
    @staticmethod
    def __call__(
```

chuckwondo: **WARNING**: I suggest _removing_ the `@staticmethod` decorator and _adding_ a `self` parameter instead.

The problem with making `__call__` a static method is that the only way to invoke it is by _explicit_ usage because it becomes shadowed by _instantiation_.

For example, let's assume we have a class `Foo` that implements `__call__` as a static method, like so (the implementation details are irrelevant):

```python
class Foo:
    @staticmethod
    def __call__(an_arg: int, *, a_kwarg: str) -> str:
        return f"{an_arg}: {a_kwarg}"
```

Unfortunately, we cannot call `__call__` _implicitly_ because the expression `Foo()` will _instantiate_ `Foo` (i.e., call `Foo.__new__` to create an instance of `Foo`, followed by invoking `__init__` on the new instance).

The only way to call `__call__` is _explicitly_: `Foo.__call__(42, "hello")`. Further, attempting to call it _implicitly_ with the expression `Foo(42, "hello")` will produce the following error because it will not actually call `__call__`, but rather `__init__`:

    TypeError: Foo.__init__() takes 1 positional argument but 3 were given

Worse still, if we _also_ implement `__init__` with the _same signature_ as `__call__`, the expression `Foo(42, "hello")` will perhaps very confusingly (for someone expecting `__call__` to be invoked) _not_ invoke `__call__`, but will instead _successfully_ cause `__init__` to be invoked (rather than raising the error above since our now _explicit_ `__init__` implementation takes the additional args).

TomNicholas: Thanks @chuckwondo. The objective here is to type hint a function with named kwargs. My understanding was that instead of using `Callable` the best way to do this was to use a protocol with a `__call__` method. Is there a better way?

chuckwondo: Yep, that's correct. Just drop the `@staticmethod` decorator and add the `self` parameter. The bottom line is that when defining a callback protocol, the convention is to _not_ use the `@staticmethod` decorator, and to add the `self` parameter.

kylebarron: @chuckwondo Are those not interchangeable? In obstore I define my protocol with a staticmethod and it works fine for classes too.

chuckwondo: > @chuckwondo Are those not interchangeable? In obstore I define my protocol with a staticmethod and it works fine for classes too.

No, those are not totally interchangeable, per my previous details about class instantiation "shadowing" the static call (thus forcing explicit use of `__call__` for static class methods). I'm curious to see the obstore code where it works fine. Perhaps your usage is different from the problematic usage I described above. Can you point me to it?

kylebarron: I meant they're interchangeable in the type system.

This type checks and runs fine:

```py
from typing import Protocol


class CallProtocol1(Protocol):
    @staticmethod
    def __call__(arg: int) -> None: ...


class CallProtocol2(Protocol):
    def __call__(self, arg: int) -> None: ...


def functional(arg: int) -> None:
    print(f"hello world, {arg}")


class ClassBased:
    def __call__(self, arg: int) -> None:
        print(f"hello world, {arg}")


def accepts_call_protocol1(arg: CallProtocol1):
    arg(1)


def accepts_call_protocol2(arg: CallProtocol2):
    arg(2)


def main():
    instance = ClassBased()

    accepts_call_protocol1(functional)
    accepts_call_protocol1(instance)

    accepts_call_protocol2(functional)
    accepts_call_protocol2(instance)


if __name__ == "__main__":
    main()
```

Prints:
```
hello world, 1
hello world, 1
hello world, 2
hello world, 2
```

TomNicholas: Hang on. Reading this again:

> The problem with making `__call__` a static method is that the only way to invoke it is by explicit usage because it becomes shadowed by instantiation.

We are not planning to instantiate this class - it's purely for type hinting purposes. We're never actually going to inherit from or directly use this class.

All we need is for this to work:

```python
from typing import Protocol


class Backend(Protocol):
    @staticmethod
    def __call__(x: int) -> int:
        ...


def custom_backend(x: int) -> int:
    return x + 1


def open_virtual_dataset(x: int, backend: Backend) -> int:
    return backend(x)


# if this type hints as correct then we're good
open_virtual_dataset(1, custom_backend)
```
When I run `mypy` on this script there are no errors.

In fact, you never actually can instantiate `Backend`, because if you try to do `Backend()` you get `TypeError: Protocols cannot be instantiated` (which is because [`typing.Protocol` is an instance of `abc.ABCMeta`](https://typing.python.org/en/latest/spec/protocol.html#defining-a-protocol)). 

A user could in theory still subclass `Backend`, which then could lead to the shadowing issue, but this is meant to be a private class, and we could at least forbid subclassing in static typing land by using `@final`.

chuckwondo: Yes, indeed. I wasn't questioning the typing details.

In your example, using `@staticmethod` doesn't cause a problem here because you have passed an instance of the class (in addition to a plain function), not the class itself to the 2 higher-order functions.

Therefore, my point is that marking a method as static when it is not (and _cannot_ implicitly be) used as a static method can lead to confusion and problems (per my previous details). In other words, it's counterintuitive to define a static method when you end up invoking it _only_ via an _instance_ of the class. Just make it an _instance_ method.

It's certainly not a deal-breaker, but IMO, it's a code smell. Since there is no intent to use `__call__` as a static method (and attempting to do so can lead to confusion, per my previous details) because we're using it _specifically_ to define a _callback protocol_ (i.e., to define a _function_ signature that cannot be described by the limitations of `Callable`, and not define any kind of typical class protocol), my suggestion is to use the syntax that more accurately conveys this case, which is to drop the decorator and add the self parameter.

To extend your example, add the following:

```python
class ClassBasedStatic:
    @staticmethod
    def __call__(arg: int) -> None:
        print(f"hello world, {arg}")
```

Since we're marking the method as static, it would only seem natural to want to be able to do the following to have `__call__` invoked:

```python
ClassBasedStatic(42)
```

However, that's _not_ what happens. Instead, you get this:

```
TypeError: ClassBasedStatic.__init__() takes 1 positional argument but 2 were given
```

Thus you also _cannot_ pass the class to either function either. These won't work, even though it would be easy to think they should, given that `__call__` is a static method, nor do they type check:

```python
accepts_call_protocol1(ClassBasedStatic)
accepts_call_protocol2(ClassBasedStatic)
```

chuckwondo: @TomNicholas, I was still typing my reply as your reply posted, so I didn't see it until after I also submitted.

Yes, I agree that there is no intent to "implement" the protocol in the sense that we don't plan to define classes that adhere to it. This is purely for defining a function signature that cannot be defined by `Callable` syntax.

However, I still feel it's just a bit counterintuitive to use the `@staticmethod` decorator in this case, and you certainly don't see it used in the [callback protocols](https://typing.python.org/en/latest/spec/callables.html#callback-protocols) docs either.

I do agree that adding `@final` as a decorator on the protocol itself will solidify the intent.

design_notes/v2.md
```diff
@@ -0,0 +1,98 @@
...
+- [Re-implement loadable_variables using ManifestStore ](https://github.com/zarr-developers/VirtualiZarr/issues/473)
+- [Should we try to use Obstore for reading files?](https://github.com/zarr-developers/VirtualiZarr/issues/476)
+- [Challenges with earthaccess.open_virtual_mfdataset for grouped netCDFs with nested structure](https://github.com/zarr-developers/VirtualiZarr/issues/487)
+- [Docs on how to write a custom reader](https://github.com/zarr-developers/VirtualiZarr/issues/452)
+- [Make readers pluggable via entrypoint system](https://github.com/zarr-developers/VirtualiZarr/issues/245)
+- [Architecture diagram](https://github.com/zarr-developers/VirtualiZarr/issues/225)
+
+## Data structures
+
+- `ChunkManifest`: An object defining the paths, chunk byte offsets, lengths of one virtual array.
```
norlandrhagen, line 25: Should `ChunkManifest` be changed to `ManifestChunk` (It sounds terrible I know) to match the `ManifestArray`, `ManifestGroup` and `ManifestStore` naming convention?

design_notes/v2.md
chuckwondo: At the end of the day, either approach will work, but given 2 competing approaches that satisfy our requirements, I simply recommend the approach that better conveys the intent, which, IMO, is to drop the staticmethod decorator and add the self parameter, which also aligns with what is shown in the official docs.

design_notes/v2.md
TomNicholas: If we wanted to make them consistent I think we would instead have just `Manifest`, or change the others to be `ChunkManifestArray` etc. But I don't really think this matters very much.

design_notes/v2.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Top-level API functions
+
+- `open_virtual_dataset` - construct a Xarray Dataset with NumPy and/or ManifestArray-backed variables from a file
+- `open_virtual_mfdataset` - construct a Xarray Dataset with NumPy and/or ManifestArray-backed variables from many files
+
+> **NOTE**: Here the construction of a virtual dataset from a manifest store would happen in the top-level function rather than the backends.
+
+```python
+ def open_virtual_dataset(
```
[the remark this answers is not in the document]
TomNicholas: `object_reader` + `object_reader`? Is that what you meant to write?

design_notes/v2.md
TomNicholas: > However, I still feel it's just a bit counterintuitive to use the @staticmethod decorator in this case, and you certainly don't see it used in the [callback protocols](https://typing.python.org/en/latest/spec/callables.html#callback-protocols) docs either.
>
> I do agree that adding @final as a decorator on the protocol itself will solidify the intent.

Okay, thanks for the clarification. I was confused because as you say either will work. But you've convinced me that the better approach is not to use `staticmethod`, and the official docs imply that it's recommended to avoid `staticmethod`.

design_notes/v2.md
maxrjones: `ChunkManifest` is the only object that really directly contains a manifest (of chunks, in this case). If we were to rename these, I would chose `ChunkManifest`, `VirtualArray`, `VirtualGroup`, and `VirtualStore`.

design_notes/v2.md
kylebarron: > Yes, I agree that there is no intent to "implement" the protocol in the sense that we don't plan to define classes that adhere to it. This is purely for defining a function signature that cannot be defined by `Callable` syntax.

On the contrary I absolutely expect we'll have **classes** that implement `__call__` and not just functions. How else would we customize the configuration of these readers?

To make this concrete, look at obstore. Obstore defines the concept of "credential providers". For S3 this is just [a callback](https://developmentseed.org/obstore/latest/api/store/aws/#obstore.store.S3CredentialProvider.__call__) that [returns](https://developmentseed.org/obstore/latest/api/store/aws/#obstore.store.S3Credential) `access_key_id`, `secret_access_key`, `token`.

Of course it's permitted for users to have totally custom logic encapsulated in a single function, but for our built-in credential providers, like [`Boto3CredentialProvider`](https://developmentseed.org/obstore/latest/api/auth/boto3/#obstore.auth.boto3.Boto3CredentialProvider), we allow passing a custom credential configuration, for example the `boto3.Session` or `botocore.session.Session`. 

Likewise, I expect that users will want to customize the behavior of each of these protocols using the built-in classes exposed by virtualizarr; so it would be clean to expose classes that implement `__call__`.

kylebarron: And to recap the staticmethod discussion; I think either way is fine. Happy to go with dropping the `@staticmethod` decorator.

design_notes/v2.md
kylebarron: > `ChunkManifest` is the only object that really directly contains a manifest (of chunks, in this case). If we were to rename these, I would chose `ChunkManifest`, `VirtualArray`, `VirtualGroup`, and `VirtualStore`.

FWIW I like these names

TomNicholas: > Is the only object that really directly contains a manifest (of chunks, in this case). 

That's not true - all of `ManifestArray`, `ManifestGroup`, `ManifestStore` contain manifests, they are just wrapped up in lower-level objects. Containing manifests is their defining feature!

If anything I regret the "virtual" nomenclature, not the "manifest" nomenclature.

design_notes/v2.md
chuckwondo: Yep, allowing class implementations is certainly acceptable, so we would then avoid adding the `@final` decorator to the protocol, mentioned more recently.

Given your additional examples, which require creating _instances_ of classes implementing the protocol, I think they give further credence to dropping the `@staticmethod` decorator on the `__call__` method.

design_notes/v2.md
TomNicholas: Literally a `ManifestArray` is defined by only two attributes (`._metadata` and `._manifest`), and `zarr.Array` also has `._metadata`.

maxrjones: Perhaps "contain" wasn't the correct word to use. My point is that ChunkManifest **is** a Manifest, so it makes sense to include "Manifest" in the name. In contrast, the others **have** a manifest.

For example, the most important feature of `ManifestStore` is that it redirects an incoming key and byte range to a different key and byte range. IMO "VirtualStore" or "IndirectionStore" (same message, more boring) is much more aligned to why it exists.

This doesn't matter a lot so I'm fine with just leaving it all, but I think the "virtual" nomenclature is more clear for objects that don't have storing references as their primary purpose.

TomNicholas: I can see the argument for having everything in the library be "virtual-something" though...

design_notes/v2.md
maxrjones: In the latest version, the `object_reader` also used by parsers to get the byte ranges for chunks rather than only to load data so I think it's definitely required in all cases.

design_notes/v2.md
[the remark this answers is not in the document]
maxrjones: sorry, I changed this to `file_url` and `object_reader` in [`37f4f1a` (#568)](https://github.com/zarr-developers/VirtualiZarr/pull/568/commits/37f4f1a65da85656198f9f910216fe76bfcd2498) because there's no good way to define the full url, which is needed for the `ChunkManifest`, from a ReadableFile. Hopefully requiring a `object_reader` instance rather than a storage configuration will solve a lot of the issues.

norlandrhagen:
Not sure which Phase in the roadmap it fits into, but having [https support for virtual stores](https://github.com/earth-mover/icechunk/issues/526) would be really clutch for a lot of usecases.

design_notes/v2.md
```diff
@@ -0,0 +1,127 @@
...
+    compat: "CompatOptions" = "no_conflicts",
+    preprocess: Callable[[Dataset], Dataset] | None = None,
+    data_vars: Literal["all", "minimal", "different"] | list[str] = "all",
+    coords="different",
+    combine: Literal["by_coords", "nested"] = "by_coords",
+    parallel: Literal["dask", "lithops", False] | Executor = False,
+    join: "JoinOptions" = "outer",
+    attrs_file: str | os.PathLike | None = None,
+    combine_attrs: "CombineAttrsOptions" = "override",
+    **kwargs,
```
maxrjones, line 121: @TomNicholas the lack of guessing functionality means that all inputs to `open_virtual_mfdataset` would need to use the same `ObjectReader` and `Parser`.

kylebarron, line 121: I'd suggest considering removing `**kwargs` from any top-level function because it's often unclear where they go. Instead, if those kwargs are intended for the object reader, they should be passed into the reader before the instance is passed into this function.

maxrjones: It provides these arguments to `open_virtual_dataset`:
```Python
    decode_times: bool | None = None,
    cftime_variables: Iterable[str] | None = None,
    indexes: Mapping[str, Index] | None = None,
  ```
Since I think the motivation was to follow the design of https://docs.xarray.dev/en/stable/generated/xarray.open_mfdataset.html closely, I doubt we'll want to change this even though I also dislike pass-through `**kwargs`.

Comment (kylebarron) [reply design_notes/v2.md:121]: If you know it's always going to be those 5 arguments, why not re-declare them here so you get accurate type hinting?

Comment (maxrjones) [reply design_notes/v2.md:121]: I defer to Tom on this to decide the preferred balance of mirroring Xarray vs accurate type hinting.

File: design_notes/v2.md [hunk 39]
```diff
@@ -0,0 +1,127 @@
...
+    manifest_store = parser(
+        file_url = file_url,
+        object_reader = object_reader,
+    )
+    ds = manifest_store.to_virtual_dataset(
+        loadable_variables=loadable_variables,
+        decode_times=decode_times,
+        indexes=indexes,
+    )
+    return ds
```
maxrjones, line 92:
```suggestion
    vds = manifest_store.to_virtual_dataset(
        loadable_variables=loadable_variables,
        decode_times=decode_times,
        indexes=indexes,
        drop_variables=drop_variables
    )
    return vds
```
@sharkinsspatial @TomNicholas each of the parsers/backends right now has redundant code to deal with `drop_variables`. I think we should instead have `drop_variables` handled either in the top level `open_virtual_dataset` via `return ds.drop_vars(drop_variables)` or in `to_virtual_dataset` which could just skip any variables in `drop_variables`.

design_notes/v2.md
maxrjones: We decided to match xarray's `**kwargs` syntax in today's coordination call.

kylebarron: Do you mean we _are_ having `**kwargs` in our signature or we're going to hoist the specific argument parameters that xarray supports into our signature so that we don't have to accept `**kwargs` here?

maxrjones: > Do you mean we are having **kwargs in our signature

Yes, following the design pattern established in https://docs.xarray.dev/en/stable/generated/xarray.open_mfdataset.html

design_notes/v2.md
```diff
@@ -0,0 +1,128 @@
...
+- [Re-implement loadable_variables using ManifestStore ](https://github.com/zarr-developers/VirtualiZarr/issues/473)
+- [Should we try to use Obstore for reading files?](https://github.com/zarr-developers/VirtualiZarr/issues/476)
+- [Challenges with earthaccess.open_virtual_mfdataset for grouped netCDFs with nested structure](https://github.com/zarr-developers/VirtualiZarr/issues/487)
+- [Docs on how to write a custom reader](https://github.com/zarr-developers/VirtualiZarr/issues/452)
+- [Make readers pluggable via entrypoint system](https://github.com/zarr-developers/VirtualiZarr/issues/245)
+- [Architecture diagram](https://github.com/zarr-developers/VirtualiZarr/issues/225)
+
+## Data structures
+
+- `ChunkManifest`: An object defining the paths, chunk byte offsets, lengths of one virtual array.
```
maxrjones, line 25:
```suggestion
- `Manifest`: An object defining the paths, chunk byte offsets, lengths of one virtual array.
```

design_notes/v2.md
TomNicholas: For now, we just want the signature of `open_virtual_X` to match the signature of `open_X` as closely as possible. Even though that signature includes untyped `**kwargs` (for `open_mfdataset`). We're choosing for now to value consistency with xarray over strong typing.

maxrjones:
@TomNicholas do you rather merge this into the repo as documentation or close it as not needed?

Issue #565, opened 2025-04-23
Virtualize Native Zarr V2 format


norlandrhagen:
This [PR](https://github.com/zarr-developers/VirtualiZarr/pull/271) should soon be merged. We intentionally didn't included reading from V2 Zarr. At some point, we should include this functionality. @sharkinsspatial mentioned some codec changes that might make this a good deal easier. 

- Zarr V2 functionality.

TomNicholas:
@neilSchroeder do you want to just have a go at finishing #806? Davis already explained what to do, and I am happy to provide further help. Otherwise this might take a month or two for me to get to perhaps.

neilSchroeder:
Thanks for the rapid response @TomNicholas. I'll take a look at it and see if I can't make some progress on it.

PR #570, opened 2025-04-25
Allows the zarr_reader to use the new Obstore functionallity released in `zarr-python 3.0.7` 

norlandrhagen:
- [ ] Towards https://github.com/zarr-developers/VirtualiZarr/issues/565
- [ ] Tests passing
- [ ] Full type hint coverage

cc @maxrjones

### DIFF
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -145,7 +145,8 @@ def default_object_store(filepath: str) -> ObjectStore:
     parsed = urlparse(filepath)
 
     if parsed.scheme in ["", "file"]:
-        return obs.store.LocalStore()
+        return obs.store.LocalStore(parsed.path)
+
     if parsed.scheme == "s3":
         bucket = parsed.netloc
         return obs.store.S3Store(
--- a/virtualizarr/readers/zarr.py
+++ b/virtualizarr/readers/zarr.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 import asyncio
 from pathlib import Path  # noqa
 from typing import (
+    TYPE_CHECKING,
     Any,
-    Hashable,
     Iterable,
     Mapping,
     Optional,
@@ -12,7 +12,6 @@ from typing import (
 
 import numpy as np
 from xarray import Dataset, Index
-from zarr.api.asynchronous import open_group as open_group_async
 from zarr.core.metadata import ArrayV3Metadata
 
 from virtualizarr.manifests import (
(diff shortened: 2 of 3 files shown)


virtualizarr/readers/zarr.py
```diff
@@ -39,6 +42,7 @@
 
 
 import zarr
+from zarr.storage import ObjectStore
```
TomNicholas: You have imported two different `ObjectStore` classes under the same name! They need to be disambiguated - I suggest just using the full name for both everywhere in this file, i.e.

```python
import zarr
import obstore

zarr.storage.ObjectStore
obstore.store.ObjectStore
```

virtualizarr/readers/zarr.py
```diff
@@ -22,9 +21,13 @@
     ManifestStore,
 )
 from virtualizarr.manifests.manifest import validate_and_normalize_path_to_uri  # noqa
+from virtualizarr.manifests.store import default_object_store
```
TomNicholas: Does this still exist?

virtualizarr/manifests/store.py
```diff
@@ -145,7 +145,8 @@ def default_object_store(filepath: str) -> ObjectStore:
     parsed = urlparse(filepath)
 
     if parsed.scheme in ["", "file"]:
-        return obs.store.LocalStore()
+        return obs.store.LocalStore(parsed.path)
```
TomNicholas: This seems worrying - do we not already have a test that would catch this?

virtualizarr/readers/zarr.py
```diff
@@ -177,15 +185,19 @@ def open_virtual_dataset(
         virtual_backend_kwargs: Optional[dict] = None,
         reader_options: Optional[dict] = None,
     ) -> Dataset:
+        if virtual_backend_kwargs:
+            raise NotImplementedError(
+                "Zarr reader does not understand any virtual_backend_kwargs"
+            )
```
TomNicholas: I think these are gone now

TomNicholas (dismissed):

virtualizarr/manifests/store.py
norlandrhagen: I believe this was done in the general parser refactor.

maxrjones: I'm confused by this conversation, is this still an issue after the removal of default_object_store?

norlandrhagen: 🤷

TomNicholas:
There's almost no changes in this PR now relative to `develop`...

maxrjones (approved):
The obstore-related change suggested in https://github.com/zarr-developers/VirtualiZarr/pull/570#issuecomment-3069683575 was included in https://github.com/zarr-developers/VirtualiZarr/pull/660, so I updated the PR title to be more descriptive about the scope of your PR which changes an internal variable name.

PR #579, opened 2025-04-29
Remove the 'rectilinear' adjective for arrays in the FAQ

maxrjones:
Rectilinear is a very specific type of grid structure and I'm quite certain that VirtualiZarr can handle a broader set of grid types. I know people are already using Zarr for HEALPix and I think the virtual zarr approach has been demonstrated for curvilinear grids. I think 'structured' is a more appropriate description of the limitations of VirtualiZarr's approach.


maxrjones:
> You're right to clarify this, but I think your clarification needs clarifying :D
> 
> I guess Zarr doesn't actually care about grids at all, only about multidimensional arrays. So we shouldn't even mention grids or grid-related words.

Yeah only the PR description mentions grids rather so I'm not sure more text adjustments would help. What do you think would about adding a figure:
requirements

TomNicholas:
I like the idea of an image, but I find this particular image really confusing 😕

TomNicholas (approved):

Issue #521, opened 2025-03-31
Make xarray an optional dependency?


TomNicholas:
I realised that the in-progess `ManifestStore` refactor would actually allow us to separate concerns so much that we could potentially make xarray an optional dependency, where you only need xarray installed if you want to use its API to manipulate virtual zarr stores (e.g. by concatenating them).

The result could work like this:

```python
# use a virtual reader directly - no xarray needed
ms: ManifestStore = manifeststore_from_hdf('file.nc')

# write to some virtual references format directly - no xarray needed
# this would use `IcechunkStore.set_virtual_refs()` as it currently does
ms.to_icechunk(icechunkstore)
```
or if you want to work in xarray space you can move to it:

```python
# xarray required to convert to virtual dataset representation
vds: xr.Dataset = ms.to_virtual_dataset(loadable_variables=...)

# (or just go straight there using our existing API)
vds: xr.Dataset = vz.open_virtual_dataset('file.nc', reader=manifeststore_from_hdf, loadable_variables=...)

# xarray required to do manipulating in xarray space
vds_combined: xr.Dataset = xr.concatenate(vds1, vds2, ...)

# write to some virtual references format - xarray required to write the non-virtual variables
# this would use `Dataset.to_zarr(icechunkstore)` internally as well as `IcechunkStore.set_virtual_refs()` it currently does
vds.to_icechunk(icechunkstore)
```

Advantages:
1) Total separation of concerns between virtualizing files into the zarr data model and manipulating them using the xarray data model,
2) Can create virtual zarr references for data that xarray cannot even represent (e.g. multiple arrays with non-alignable dimensions in the same group).

Disadvantages:
1) Might be less clear for non-expert users, because there are now two ways to read and write references. I still think we would present the xarray interface as the standard UI, we would just mention that this is possible in a developers section of the docs, as `ManifestStore` is only supposed to be developer API anyway.

maxrjones:
I think namespace packages could help with the separation of concerns, in addition to the optional dependencies approach. For example:

`virtualizarr.manifests` - In-memory manifest representations (Chunk, Array, Group, Store)
`virtualizarr.readers` - All "canonical" reader functions (HDF5, TIFF, etc.)
`virtualizarr.xarray` - Access to Xarray's manipulation functionality
`virtualizarr.writers` - Write to Icechunk or Kerchunk

[...]

PR #591, opened 2025-05-11
Add `.to_icechunk()` method to `ManifestGroup`

TomNicholas:
This (WIP) PR adds a `to_icechunk()` method directly on `ManifestGroup`, and refactors the `.to_icechunk()` virtual dataset accessor method to go via a the `ManifestGroup` method.

The point is 
1. To further modularize the code by providing the dual of going from `ManifestStore` to virtual dataset, by adding the ability to go back from a virtual dataset to a ManifestStore,
2. To therefore allow users to create and persist virtual references for a single file without ever using xarray, i.e. enable #521.

After this PR we should do the same for `.to_kerchunk()` -  at that point we could actually make xarray an optional dependency if we want to.

- [x] Closes first step of #521
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/group.py
+++ b/virtualizarr/manifests/group.py
@@ -1,12 +1,16 @@
 from __future__ import annotations
 
 import textwrap
-from typing import Iterator, Mapping
+from time import datetime
+from typing import TYPE_CHECKING, Iterator, Mapping, Optional
 
 import xarray as xr
 from zarr.core.group import GroupMetadata
 
 from virtualizarr.manifests import ManifestArray
+
+if TYPE_CHECKING:
+    from icechunk import IcechunkStore  # type: ignore[import-not-found]
 
 
 class ManifestGroup(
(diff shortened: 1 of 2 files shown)


virtualizarr/manifests/group.py
```diff
@@ -131,3 +135,70 @@ def to_virtual_dataset(self) -> xr.Dataset:
...
+
+        attributes = vds.attrs
+        # TODO test this is correct
+        attributes["dimension_names"] = " ".join(list(vds.dims))
+        # TODO test this constructor round-trips coordinates
+        attributes["coordinates"] = " ".join(list(vds.coords))
+
+        return cls(arrays=manifestarrays, attributes=attributes)
+
+    def to_icechunk(
```
TomNicholas, line 171: The point of adding this on `ManifestGroup` instead of `ManifestStore` initially is to make it easier to compose to deal with virtual datatrees later.

virtualizarr/manifests/group.py
```diff
@@ -131,3 +135,70 @@ def to_virtual_dataset(self) -> xr.Dataset:
...
+
+        manifestarrays = {name: var.data for name, var in vds.variables.items()}
+
+        attributes = vds.attrs
+        # TODO test this is correct
+        attributes["dimension_names"] = " ".join(list(vds.dims))
+        # TODO test this constructor round-trips coordinates
+        attributes["coordinates"] = " ".join(list(vds.coords))
+
+        return cls(arrays=manifestarrays, attributes=attributes)
```
maxrjones, line 169: Could this functionality be moved to a different PR to constrain the scope of the ManifestGroup.to_icechunk feature?

Issue #38, opened 2024-03-15
Treatment of final smaller chunk in the zarr model


TomNicholas:
The `ManifestArray` code right now might not be treating carefully enough the possibility of the last chunk being smaller than the rest. For example if you concatenate two arrays that each have uniform chunks except for the final chunk, you could end up with an array that has variable-length chunks in the middle somewhere 😕 

One solution to this might be to generalise `ManifestArray.chunks` to be able to represent rectilinear arrays of chunks, but just have a check on the constructor to forbid creating an array that actually has any chunk other than the last be of different length. This check could then be relaxed in future once we have variable-length chunks supported upstream in zarr.

TomNicholas:
We need some test cases to start explicitly checking behaviour for these cases. Unfortunately kerchunk doesn't seem to have any (see https://github.com/fsspec/kerchunk/issues/436). This is a little worrying because this seems like an important set of edge cases to check, so we should try to find / create some files like this ourselves and add them to the test suite of this library.

TomNicholas:
> have a check on the constructor to forbid creating an array that actually has any chunk other than the last be of different length.

A better place to have this check would probably be in the serialization step. It makes perfect sense to have a `ManifestArray` that points to variable-length chunks, the problem is just that neither of the available on-disk formats (kerchunk or zarr) support variable-length chunks yet. (A warning instead of an error in the constructor would be more appropriate.)

PR #594, opened 2025-05-13
Raise error on concatenation with partial boundary chunks

maxrjones:
Addresses https://github.com/zarr-developers/VirtualiZarr/issues/38

Some notes:
- As far as I can tell HDF5, similar to Zarr, pads the boundary chunks to a full chunk so that virtualizing a single file works fine and the issue is only with concatenation.
- Icechunk doesn't allow setting these virtual chunks (nice!), so I roundtripped as a kerchunk dict.
- I'm surprised by the outcome of this test - it seems that it's just not possible to load the array but there's no error message


virtualizarr/manifests/utils.py
```diff
@@ -157,6 +157,21 @@ def _remove_element_at_position(t: tuple[int, ...], pos: int) -> tuple[int, ...]
     return tuple(new_l)
 
 
+def check_no_partial_chunks_on_concat_axis(
```
TomNicholas, line 160: The place to test this is here:

https://github.com/zarr-developers/VirtualiZarr/blob/627a9049e9eb0f2e69a40f3d511ce7ebac3ee45d/virtualizarr/tests/test_manifests/test_array.py#L352

because we should be able to test this error message without ever doing IO.

virtualizarr/manifests/array_api.py
```diff
@@ -75,7 +76,9 @@ def concatenate(
         axis = axis % first_arr.ndim
 
     arr_shapes = [arr.shape for arr in arrays]
+    arr_chunks = [arr.chunks for arr in arrays]
     check_same_shapes_except_on_concat_axis(arr_shapes, axis)
+    check_no_partial_chunks_on_concat_axis(arr_shapes, arr_chunks, axis)
```
TomNicholas, line 81: Are we also missing a `check_same_chunks` function call here? 😬 Or does it happen inside `check_combinable_zarr_arrays`?

virtualizarr/manifests/utils.py
```diff
@@ -157,6 +157,21 @@ def _remove_element_at_position(t: tuple[int, ...], pos: int) -> tuple[int, ...]
...
+    """Check that there are no partial chunks along the concatenation axis"""
+
+    partial_chunks = any(
+        shape[axis] % chunk[axis] > 0 for shape, chunk in zip(shapes, chunks)
+    )
+
+    if partial_chunks:
+        raise ValueError(
+            f"Cannot concatenate arrays with shapes {[shape for shape in shapes]}  and chunks {[chunk for chunk in chunks]} because only regular chunk shapes are currently supported."
+        )
```
TomNicholas, line 172: This is neat, but it could be made more specific. We could do something like this:

```python
def has_partial_chunks(shape: tuple[int, ...], chunk_shape: tuple[int, ...]) -> bool:
    """Check that a single array has no partial chunks along any axis"""
    # TODO is that actually the requirement? Or is it merely that there not be partial chunks along the concatenation axis
    return any(
        shape[axis] % chunk[axis] > 0 for axis in range(len(shape))
    )


# loop over the arrays to be concatenated
for i, (shape, chunk_shape) in enumerate(zip(shapes, chunks)):
    if has_partial_chunks(shape, chunk):
        raise ValueError(
            f"Cannot concatenate arrays with partial chunks because only regular chunk grids are supported, but array {i} has shape={shape} and chunks={chunks}"
        )

```

TomNicholas: That would tell us _which_ array is the offending one

TomNicholas (commented):

> As far as I can tell HDF5, similar to Zarr, pads the boundary chunks to a full chunk so that virtualizing a single file works fine and the issue is only with concatenation.

We should explicitly test this then. Have a test that we can roundtrip one file with partial chunks without any concatenation involved.

> Icechunk doesn't allow setting these virtual chunks (nice!)

What do you mean? Icechunk refuses to accept virtual chunks whose chunk shape is smaller than the chunk shape of the array they are being appended to?

> I'm surprised by the outcome of this test - it seems that it's just not possible to load the array but there's no error message

What do you mean? The tests seem to pass?

maxrjones:
>> I'm surprised by the outcome of this test - it seems that it's just not possible to load the array but there's no error message

> What do you mean? The tests seem to pass?

Sorry, I didn't update the PR description after making subsequent commits. I was describing the situation after [`e8300d4` (#594)](https://github.com/zarr-developers/VirtualiZarr/pull/594/commits/e8300d4278e1988e0fce266d0d710a7bd2a6a136).

virtualizarr/manifests/array_api.py
maxrjones: It happens inside `check_combinable_zarr_arrays`. I guess that `check_combinable_zarr_arrays` does all the checks that do not depend on the `axis` of concatenation.

virtualizarr/manifests/utils.py
maxrjones: Good call, I modified it slightly to indicate if there are multiple offenders (which should always be the case because otherwise the prior `all_chunks_equal` check would've been hit).

maxrjones: Actually I should use your code since it's probably not ideal to have a really long error message, which would happen if someone's trying to concatenate thousands of files. I'll update this tomorrow.

PR #595, opened 2025-05-15
Support using an ObjectStore directly with ManifestStore

maxrjones:
Part of the migration detailed in #568

### DIFF
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -240,13 +240,18 @@ class ManifestStore(Store):
     """
 
     _group: ManifestGroup
-    _store_registry: ObjectStoreRegistry
+    _store_registry: ObjectStoreRegistry | None
+    _store: ObjectStore | None
 
     def __eq__(self, value: object):
         NotImplementedError
 
     def __init__(
-        self, group: ManifestGroup, *, store_registry: ObjectStoreRegistry | None = None
+        self,
+        group: ManifestGroup,
+        *,
+        store: ObjectStore | None = None,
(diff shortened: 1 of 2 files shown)


virtualizarr/manifests/store.py
```diff
@@ -241,19 +241,26 @@ class ManifestStore(Store):
...
     def __eq__(self, value: object):
         NotImplementedError
 
     def __init__(
-        self, group: ManifestGroup, *, store_registry: ObjectStoreRegistry | None = None
+        self,
+        group: ManifestGroup,
+        *,
+        store: ObjectStore | None = None,
+        store_registry: ObjectStoreRegistry | None = None,
```
TomNicholas, line 254: I'm not sure I understand the purpose of this PR. If you only want to pass a single `ObjectStore`, couldn't you just pass an `ObjectStoreRegistry` with just one element in it?

maxrjones: Let's add this to the agenda for tomorrow's sync. `ObjectStoreRegistry` adds some complexity, but allows combining `ManifestGroup`s that point to files in different storage locations (e.g., an S3 bucket and Azure container) in a single `ManifestStore`. If we think that's worthwhile complexity, we don't need this PR at all.

TomNicholas: > combining ManifestGroups that point to files in different storage locations

If one file becomes one `ManifestStore`, why would we ever need that?

I had thought the point of the registry was to allow opening metadata formats like kerchunk/DMR++ with one objectstore, and loading the data they point to from another bucket using a different objectstore.

PR #598, opened 2025-05-16
Add V2 design for parser protocol and open_virtual_dataset

### LINKED ISSUE #245: Make readers pluggable via entrypoint system

### LINKED ISSUE #553: Function definition for `open_virtual_dataset` in next major release

maxrjones:
Following design structure in #568 

Builds on changes in #595, which should be merged first.

Closes #553 
Closes #245

### DIFF
--- a/virtualizarr/testing/utils.py
+++ b/virtualizarr/testing/utils.py
@@ -10,7 +10,7 @@ from virtualizarr.manifests import (
 from virtualizarr.manifests.utils import create_v3_array_metadata
 
 
-def _put_fake_data(store: ObjectStore, filepath: str):
+def put_fake_data(store: ObjectStore, filepath: str):
     """
     Puts a sequence of 16 bytes in a file, which can simulate storing
     4 contiguous uncompressed 4-byte chunks (or 8 2-byte chunks, etc). This
@@ -52,7 +52,6 @@ def fake_parser(filepath: str, object_reader: ObjectStore) -> ManifestStore:
     --------
     ManifestStore
     """
-    _put_fake_data(object_reader, filepath=filepath)
     chunk_dict = {
         "0.0": {"path": filepath, "offset": 0, "length": 4},
         "0.1": {"path": filepath, "offset": 4, "length": 4},
--- a/virtualizarr/testing/utils.py
+++ b/virtualizarr/testing/utils.py
@@ -6,6 +6,7 @@ from virtualizarr.manifests import (
     ManifestArray,
     ManifestGroup,
     ManifestStore,
+    ObjectStoreRegistry,
 )
 from virtualizarr.manifests.utils import create_v3_array_metadata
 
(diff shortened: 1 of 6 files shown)


virtualizarr/testing/utils.py
```diff
@@ -0,0 +1,78 @@
...
+    -----------
+    filepath : str
+        Filepath for storing temporary testing file
+    store : ObjectStore
+        ObjectStore instance for holding the file
+    Returns:
+    --------
+    ManifestStore
+    """
+    _put_fake_data(object_reader, filepath=filepath)
```
TomNicholas, line 55: Shouldn't you do this outside of the `fake_parser`?

virtualizarr/v2/__init__.py
TomNicholas, line 1: Why do we need a `v2` namespace? Shouldn't we just be directly replacing the functions we already have?

maxrjones: The purpose is to allow updating parsers incrementally, but no worries if you prefer to do it all in one go with Sean's branch.

TomNicholas: I would prefer to do it all in one go if we can!

PR #602, opened 2025-05-23
Fix upstream tests by removing fsspec pin

maxrjones:
Fixing dependency resolution, See https://github.com/fsspec/s3fs/pull/966 for context


maxrjones:
I'm going to merge this despite the failures because those are unrelated and this change will unblock development elswhere.

PR #486, opened 2025-03-12
Start Executing Docs

ianhi:
Beginning of executable docs (#83). All notebooks will automatically be executed as part of the docs build per the docs here:  https://myst-nb.readthedocs.io/en/latest/computation/execute.html#

It will be useful to set up the build failing on unexepcted errors per the docs here: https://myst-nb.readthedocs.io/en/latest/computation/execute.html#error-reporting-warning-vs-failure

Currently has multiple errors in the `usage.ipynb` file @TomNicholas

PR #606, opened 2025-05-26
WIP: Use mkdocs-material for documentation

maxrjones:
I haven't updated all the pages yet for the new format, so please don't nit yet but I'd really appreciate some high level feedback on using mkdocs material and plugins to address #500 and #83.


TomNicholas:
I was secretly hoping you would do this @maxrjones 😆 

It looks beautiful!

> find it easier to navigate with the page nav on the left and section nav on the right.

This is definitely an improvement.

> I also prefer executable markdown cells over needing to convert the files to Jupyter notebooks.

Are cells being executed in the generation of these docs? I know that in #486 we found that there were a few bugs in the executable code examples. If possible I think having one PR that changes the docs format and a follow-up that makes cells executable would be the easiest way.

PR #611, opened 2025-06-10
Update type-ignore coverage

maxrjones:
This fixes the failing mypy check

### DIFF
--- a/virtualizarr/translators/kerchunk.py
+++ b/virtualizarr/translators/kerchunk.py
@@ -37,7 +37,7 @@ def to_kerchunk_json(v2_metadata: ArrayV2Metadata) -> str:
             if codec is not None
         ]  # type: ignore[assignment]
     if v2_metadata.compressor:
-        zarray_dict["compressor"] = v2_metadata.compressor.get_config()  # type: ignore[assignment]
+        zarray_dict["compressor"] = v2_metadata.compressor.get_config()  # type: ignore[union-attr]
 
     return json.dumps(zarray_dict, separators=(",", ":"), cls=NumpyEncoder)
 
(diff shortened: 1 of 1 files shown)


virtualizarr/translators/kerchunk.py
```diff
@@ -37,7 +37,7 @@ def to_kerchunk_json(v2_metadata: ArrayV2Metadata) -> str:
             if codec is not None
         ]  # type: ignore[assignment]
     if v2_metadata.compressor:
-        zarray_dict["compressor"] = v2_metadata.compressor.get_config()  # type: ignore[assignment]
+        zarray_dict["compressor"] = v2_metadata.compressor.get_config()  # type: ignore[union-attr]
```
chuckwondo: Since the type for `v2_metadata.compressor` is `zarr.core.metadata.v2.CompressorLikev2`, which is a type alias for `dict[str, JSON] | numcodecs.abc.Codec | None`, and `numcodecs.abc.Codec` is the only type from that union that defines a `get_config` method, you might want to add a bit of safety (and perhaps also be able to eliminate the "ignore" comment completely) by doing something like the following instead:

```python
if isinstance(compressor := v2_metadata.compressor, numcodecs.abc.Codec):
    zarray_dict["compressor"] = compressor.get_config()
```

(On a related note, I'd suggest also opening an issue in zarr-python to change that type alias to drop the `None` from the union. Including `None` in a type alias union can be problematic, or at least cumbersome. When something is "optional", it should be indicated separately from the non-None type. So instead of typing something optional as `CompressorLikev2`, which is _automatically_ optional since `None` is part of the alias, it would be typed as `CompressorLikev2 | None`. In other words, you can _never_ define the type as _not_ optional when `None` is part of the type alias.)

chuckwondo (changes requested):
I've locally tried my suggestion, and it indeed allows removal of the "ignore" comment.

Issue #617, opened 2025-06-17
Remove vendored zarr code


TomNicholas:
We are apparently still vendoring some code from zarr-python, even though a while ago we added an explicit dependency on it (in #482). That vendored code should have been removed then in favour of directly importing things.

TomNicholas:
This came up when I got this error trying to run tests locally

```python
ImportError while loading conftest '/Users/tom/Documents/Work/Code/VirtualiZarr/conftest.py'.
conftest.py:16: in <module>
    from virtualizarr.manifests import ChunkManifest, ManifestArray
virtualizarr/__init__.py:3: in <module>
    from virtualizarr.accessor import (
virtualizarr/accessor.py:7: in <module>
    from virtualizarr.manifests import ManifestArray
virtualizarr/manifests/__init__.py:7: in <module>
    from virtualizarr.manifests.store import ManifestStore, ObjectStoreRegistry  # type: ignore # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
virtualizarr/manifests/store.py:20: in <module>
    from virtualizarr.vendor.zarr.core.metadata import dict_to_buffer
virtualizarr/vendor/zarr/core/metadata.py:6: in <module>
    from zarr.core.metadata.v3 import V3JsonEncoder
E   ImportError: cannot import name 'V3JsonEncoder' from 'zarr.core.metadata.v3' (/Users/tom/Documents/Work/Code/VirtualiZarr/.venv/lib/python3.13/site-packages/zarr/core/metadata/v3.py)
```
But actually this might be related to the zarr v3 dtypes refactor?

PR #545, opened 2025-04-11
Test endianess support against zarr-python data types refactor

TomNicholas:
This PR is for testing support for @d-v-b's unmerged zarr data types refactor in https://github.com/zarr-developers/zarr-python/pull/2874. Particularly we care about supporting big-endian data.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage
- [ ] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`
- [ ] New functionality has documentation

PR #618, opened 2025-06-17
Zarr data types refactor compatibility

TomNicholas:
Supercedes #545 now that https://github.com/zarr-developers/zarr-python/pull/2874 has been merged upstream.

The upstream changes are already being tested (and unsurprisingly causing some issues - see also #617) in our upstream dev CI tests. But here we will x-fail stuff, start trying to fix things, and bump required dependencies.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

FYI @d-v-b

### DIFF
--- a/virtualizarr/tests/test_parsers/test_fits.py
+++ b/virtualizarr/tests/test_parsers/test_fits.py
@@ -11,9 +11,6 @@ pytest.importorskip("astropy")
 
 @requires_kerchunk
 @requires_network
-@pytest.mark.xfail(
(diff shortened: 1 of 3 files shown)


TomNicholas:
@d-v-b what would be the preferred way to replace the usage of `V3JsonEncoder` (and the older `dict_to_buffer` that we apparently vendored in https://github.com/zarr-developers/VirtualiZarr/blob/develop/virtualizarr/vendor/zarr/core/metadata.py?

Issue #616, opened 2025-06-17
Move `Parser` protocol to `parsers.typing`


TomNicholas:
The `Parser` class added in #601 should be advertised publicly in API docs but it's really a typing protocol - it implements nothing. So it should live in a dedicated typing namespace to make this clearer. Otherwise someone might be confused why using `from VirtualiZarr.parsers import Parser` doesn't work.

PR #601, opened 2025-05-20
Refactor codebase to support a new simplified Parser->ManifestStore model.

sharkinsspatial:
- [x] Closes 
https://github.com/zarr-developers/VirtualiZarr/issues/561
https://github.com/zarr-developers/VirtualiZarr/issues/553
https://github.com/zarr-developers/VirtualiZarr/issues/498
https://github.com/zarr-developers/VirtualiZarr/issues/476
https://github.com/zarr-developers/VirtualiZarr/issues/559
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage
- [ ] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`
- [ ] New functionality has documentation

PR #619, opened 2025-06-17
Docs and API follow-ups to #601

TomNicholas:
This started out as a targeted PR to address #616 and ended up as an attempt to address all the uncheck bullets from #601 (i.e. everything in the docs that touches the concept of parsers).

- [x] Follow up to #601
- [ ] ~~Tests added~~
- [ ] Tests passing
- [x] Full type hint coverage

fyi @sharkinsspatial @maxrjones @chuckwondo


docs/custom_readers.md
TomNicholas, line 1: Don't know why `git mv` didn't understand that I was just renaming this file. (It has a bunch of other changes too.)

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+
+    # parse the file contents to extract its metadata
+    # this is generally where the format-specific logic lives
+    manifestgroup: ManifestGroup = extract_metadata(readable_file)
+
+    # optionally create an object store registry, used to actually load chunk data from file later
+    registry = ObjectStoreRegistry({store_prefix: object_store})
+
+    # construct the Manifeststore from the parsed metadata and the object store registry
+    return ManifestStore(group=manifestgroup, store_registry=registry)
```
TomNicholas, line 30: Writing this out made me realize it's a bit weird that exactly one `ObjectStore` is required by the call signature, but not actually technically needed by the code...

TomNicholas:
Okay I've split that out in #621. The rest of this PR basically does a few (related) things:
1. grep for "reader" and replace with "parser"
2. modify any language referring to parsers / ManifestStore to be up-to-date with the changes in #601
3. update code examples to use parser and obstore instead of reader

Those could be separated further if it would help, but it's now already down to being a pure docs (including docs examples) PR.

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+
+
+vds = vz.open_virtual_dataset(
+    file_url,
+    object_store=object_store,
+    parser=custom_parser,
+)
+```
+
+All parsers _must_ follow this exact call signature, enforced at runtime by checking against the the :py:class:`virtualizarr.parsers.typing.Parser` typing protocol.
```
chuckwondo: remove duplicate "the":

```suggestion
All parsers _must_ follow this exact call signature, enforced at runtime by checking against the :py:class:`virtualizarr.parsers.typing.Parser` typing protocol.
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+
+    # construct the Manifeststore from the parsed metadata and the object store registry
+    return ManifestStore(group=manifestgroup, store_registry=registry)
+
+
+vds = vz.open_virtual_dataset(
+    file_url,
+    object_store=object_store,
+    parser=custom_parser,
+)
```
chuckwondo, line 37: I suggest we use context managers in examples to show recommended usage to ensure resources are properly managed to avoid leaks:

```suggestion
with vz.open_virtual_dataset(
    file_url,
    object_store=object_store,
    parser=custom_parser,
) as vds:
    ...
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+def custom_parser(file_url: str, object_store: ObjectStore) -> ManifestStore:
+    # access the file's contents, e.g. using the ObjectStore instance
+    readable_file = obstore.open_reader(object_store, file_url)
+
+    # parse the file contents to extract its metadata
+    # this is generally where the format-specific logic lives
+    manifestgroup: ManifestGroup = extract_metadata(readable_file)
+
+    # optionally create an object store registry, used to actually load chunk data from file later
+    registry = ObjectStoreRegistry({store_prefix: object_store})
```
chuckwondo, line 27: `store_prefix` is undefined. Do we want to add a line or 2 of code/comment about it, or at least a comment referring users to a section of the docs covering registries?

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+    parser=custom_parser,
+)
+```
+
+All parsers _must_ follow this exact call signature, enforced at runtime by checking against the the :py:class:`virtualizarr.parsers.typing.Parser` typing protocol.
+
+```{note}
+The object store registry can technically be empty, but to be able to read actual chunks of data back from the `ManifestStore` later the registry needs to contain at least one `ObjectStore` instance.
+
+The only time you might want to use an empty object store registry is if you are attempting to parse a custom metadata-only references format without touching the original files they refer to - i.e. a format like Kerchunk or DMR++, that doesn't contain actual binary data values.
```
chuckwondo:
```suggestion
The only time you might want to use an empty object store registry is if you are attempting to parse a custom metadata-only references format without touching the original files they refer to -- i.e., a format like Kerchunk or DMR++, that doesn't contain actual binary data values.
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+    file_url,
+    object_store=object_store,
+    parser=custom_parser,
+)
+```
+
+All parsers _must_ follow this exact call signature, enforced at runtime by checking against the the :py:class:`virtualizarr.parsers.typing.Parser` typing protocol.
+
+```{note}
+The object store registry can technically be empty, but to be able to read actual chunks of data back from the `ManifestStore` later the registry needs to contain at least one `ObjectStore` instance.
```
chuckwondo:
```suggestion
The object store registry can technically be empty, but to be able to read actual chunks of data back from the `ManifestStore` later, the registry needs to contain at least one `ObjectStore` instance.
```

docs/custom_parsers.md
chuckwondo: Tangentially, can we rename the `object_store` parameter to simply `store`? That would be consistent with the names `store_prefix` and `store_registry` elsewhere.

However, would that then cause potential confusion with `zarr.abc.Store`? If so, then wouldn't `store_prefix` and `store_registry` also cause confusion about what type of store they are related to (obstore or zarr)?

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+Note using xarray like this would produce an entirely non-virtual dataset, so is equivalent to passing
+
+```python
+ds = vz.open_virtual_dataset(
+    file_url,
+    object_store=object_store,
+    parser=parser,
+    loadable_variables=<all_the_variable_names>,
+)
+```
```
chuckwondo, line 96: Related to my other suggestions about using a context manager, do we also want to add context managers to these examples as well?

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+
+        # construct the Manifeststore from the parsed metadata
+        return ManifestStore(...)
+
+
+vds = vz.open_virtual_dataset(
+    file_url,
+    object_store=object_store,
+    parser=CustomParser(**kwargs),
+)
```
chuckwondo, line 135: Context manager here too?

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+This helps to keep format-specific parser configuration separate from kwargs to `open_virtual_dataset`.
+
+## How to write your own custom parser
+
+As long as your custom parser callable follows the interface above, you can implement it in any way you like.
+However there are few common approaches.
+
+### Typical VirtualiZarr parsers
+
+The recommended way to implement a custom parser is simply to parse the given file yourself, and construct the `ManifestStore` object explicitly component by component, extracting the metadata that you need.
```
chuckwondo:
```suggestion
The recommended way to implement a custom parser is simply to parse the given file yourself, and construct the `ManifestStore` object explicitly, component by component, extracting the metadata that you need.
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+2. Read metadata to determine how many arrays there are in the file, their shapes, chunk shapes, dimensions, codecs, and other metadata.
+3. For each array in the file:
+  4. Create a `zarr.core.metadata.ArrayV3Metadata` object to hold that metadata, including dimension names. At this point you may have to define new Zarr codecs to support deserializing your data (though hopefully the standard Zarr codecs are sufficient).
+  5. Extract the byte ranges of each chunk and store them alongside the fully-qualified filepath in a `ChunkManifest` object.
+  6. Create one `ManifestArray` object, using the corresponding `ArrayV3Metadata` and `ChunkManifest` objects.
+7. Group `ManifestArrays` into one or more `ManifestGroup` objects. Ideally you would only have one group, but your format's data model may preclude that. If there is group-level metadata attach this to the `ManifestGroup` object as a `zarr.metadata.GroupMetadata` object. Remember that `ManifestGroups` can contain other groups as well as arrays.
+8. Instantiate the final `ManifestStore` using the top-most `ManifestGroup` and return it.
```
chuckwondo: The numbering and indentation caused everything to be rendered as top-level items.

This indentation correctly nests sub-items, and by simply numbering everything with `1.`, the markdown renderer will correctly number things so we don't have to concern ourselves with getting the numbers correct.

```suggestion

1. Extract file header or magic bytes to confirm the file passed is the format your parser expects.
1. Read metadata to determine how many arrays there are in the file, their shapes, chunk shapes, dimensions, codecs, and other metadata.
1. For each array in the file:
   1. Create a `zarr.core.metadata.ArrayV3Metadata` object to hold that metadata, including dimension names. At this point you may have to define new Zarr codecs to support deserializing your data (though hopefully the standard Zarr codecs are sufficient).
   1. Extract the byte ranges of each chunk and store them alongside the fully-qualified filepath in a `ChunkManifest` object.
   1. Create one `ManifestArray` object, using the corresponding `ArrayV3Metadata` and `ChunkManifest` objects.
1. Group `ManifestArrays` into one or more `ManifestGroup` objects. Ideally you would only have one group, but your format's data model may preclude that. If there is group-level metadata attach this to the `ManifestGroup` object as a `zarr.metadata.GroupMetadata` object. Remember that `ManifestGroups` can contain other groups as well as arrays.
1. Instantiate the final `ManifestStore` using the top-most `ManifestGroup` and return it.
```

virtualizarr/parsers/typing.py
```diff
@@ -0,0 +1,34 @@
...
+    ----------
+    file_url
+        The URI or path to the input file (e.g., "s3://bucket/file.nc").
+    object_store
+        An obstore ObjectStore instance for accessing the file specified in the `file_url` parameter.
+
+    Returns
+    -------
+    ManifestStore
+        A ManifestStore which provides a Zarr representation of the parsed file.
```
chuckwondo:
```suggestion
        A ManifestStore providing a Zarr representation of the parsed file.
```

virtualizarr/parsers/hdf/__init__.py
```diff
@@ -0,0 +1 @@
+from virtualizarr.parsers.hdf.hdf import HDFParser  # noqa: F401
```
chuckwondo:
```suggestion
from virtualizarr.parsers.hdf.hdf import HDFParser

__all__ = ["HDFParser"]
```

docs/usage.md
```diff
@@ -367,9 +379,11 @@ You can open existing Kerchunk `json` or `parquet` references as Virtualizarr vi
 This may be useful for manipulating them or converting existing kerchunk-formatted references to other reference storage formats such as [Icechunk](https://icechunk.io/).
 
 ```python
-vds = open_virtual_dataset('combined.json', filetype='kerchunk')
+from virtualizarr.parsers import KerchunkJSONParser, KerchunkParquetParser
+
+vds = open_virtual_dataset('combined.json', , object_store=LocalStore, parser=KerchunkJSONParser())
```
chuckwondo:
```suggestion
vds = open_virtual_dataset('combined.json', object_store=LocalStore, parser=KerchunkJSONParser())
```

docs/usage.md
```diff
@@ -367,9 +379,11 @@ You can open existing Kerchunk `json` or `parquet` references as Virtualizarr vi
...
 # or
-vds = open_virtual_dataset('combined.parquet', filetype='kerchunk')
+vds = open_virtual_dataset('combined.parquet', object_store=LocalStore, parser=KerchunkParquetParser())
```
chuckwondo, line 422: context manager here and above too?

docs/usage.md
```diff
@@ -257,8 +269,8 @@ However, you should first be confident that the archival files actually do have
 If you're happy to load 1D dimension coordinates into memory, you can use their values to do the ordering for you!
 
 ```python
-vds1 = open_virtual_dataset('air1.nc')
-vds2 = open_virtual_dataset('air2.nc')
+vds1 = open_virtual_dataset('air1.nc', object_store=LocalStore, parser=HDFParser())
+vds2 = open_virtual_dataset('air2.nc', object_store=LocalStore, parser=HDFParser())
```
chuckwondo, line 290: How about this (and same for snippet above too)?

```suggestion
with (
    open_virtual_dataset('air1.nc', object_store=LocalStore, parser=HDFParser()) as vds1,
    open_virtual_dataset('air2.nc', object_store=LocalStore, parser=HDFParser()) as vds2,
):
    ...
```

docs/usage.md
```diff
@@ -143,6 +153,8 @@ To decode time variables according to the CF conventions upon loading, you must
 ```python
 vds = open_virtual_dataset(
     'air.nc',
+    object_store=LocalStore,
+    parser=HDFParser(),
     loadable_variables=['air', 'time'],
     decode_times=True,
 )
```
chuckwondo, line 172: context manager?

docs/usage.md
```diff
@@ -102,7 +107,12 @@ Loading the variables during virtual dataset creation has several benefits detai
 You can use the `loadable_variables` argument to specify variables to load as regular variables rather than virtual variables:
 
 ```python
-vds = open_virtual_dataset('air.nc', loadable_variables=['air', 'time'])
+vds = open_virtual_dataset(
+    'air.nc',
+    object_store=LocalStore,
+    parser=HDFParser(),
+    loadable_variables=['air', 'time'],
+)
```
chuckwondo, line 128: context manager?

docs/usage.md
```diff
@@ -86,11 +88,14 @@ _The only use case for a virtual dataset is [combining references](#combining-vi
...
+To open remote files as virtual datasets use the necessary obstore store and [obstore credentials](https://developmentseed.org/obstore/latest/authentication/), for example:
 
 ```python
-aws_credentials = {"key": ..., "secret": ...}
-vds = open_virtual_dataset("s3://some-bucket/file.nc", reader_options={'storage_options': aws_credentials})
+from obstore.store import S3Store
+
+# set S3 credentials as environment variables
+
+vds = open_virtual_dataset("s3://some-bucket/file.nc", object_store=S3Store, parser=HDFParser())
```
chuckwondo, line 111: context manager?

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+```python
+import xarray.testing as xrt
+
+from virtualizarr.parsers import HDFParser
+
+manifest_store = HDFParser("file.nc", object_store=obstore.LocalStore)
+actual = xr.open_dataset(manifest_store, engine="zarr")
+
+expected = xr.open_dataset(manifest_store, backend="h5netcdf")
+xrt.assert_identical(actual, expected)
```
chuckwondo:
```suggestion

with (
    xr.open_dataset(manifest_store, engine="zarr") as actual,
    xr.open_dataset(manifest_store, backend="h5netcdf") as expected,
):
    xrt.assert_identical(actual, expected)
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+1. All arrays must have dimension names, specified in the `ArrayV3Metadata` objects.
+2. All arrays in the same group with a common dimension name must have the same length along that common dimension.
+
+You also may want to set the `coordinates` field of the group metadata to tell xarray to set those variables as coordinates upon conversion.
+
+## Testing your new parser
+
+The fact we can read data from the `ManifestStore` is useful for testing that our parser implementation behaves as expected.
+
+If we already have some other way to read data directly into memory from that archival file format - for example a conventional xarray IO backend - we can compare the results of opening and loading data via the two approaches.
```
chuckwondo:
```suggestion
If we already have some other way to read data directly into memory from that archival file format -- for example, a conventional xarray IO backend -- we can compare the results of opening and loading data via the two approaches.
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+
+You can therefore use a function which returns in-memory kerchunk JSON references inside your parser, then simply call `KerchunkJSONParser` and return the result.
+
+```{note}
+Whilst this might be the quickest way to get a custom parser working, we do not really recommend this approach, as:
+1. The Kerchunk in-memory nested dictionary format is very memory-inefficient compared to the numpy array representation used internally by VirtualiZarr's `ChunkManifest` class,
+2. The Kerchunk package in general has a number of known bugs, often stemming from a lack of clear internal abstractions and specification,
+3. This lack of data model enforcement means that the dictionaries returned by different Kerchunk parsers sometimes follow inconsistent schemas ([for example](https://github.com/fsspec/kerchunk/issues/561)).
+
+Nevertheless this approach is currently used by VirtualiZarr internally, at least for the FITS, netCDF3, and (now-deprecated original implementation of the) HDF5 file format parsers.
```
chuckwondo:
```suggestion
Nevertheless, this approach is currently used by VirtualiZarr internally, at least for the FITS, netCDF3, and (now-deprecated original implementation of the) HDF5 file format parsers.
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Kerchunk-based parsers
+
+The Kerchunk package includes code for parsing various array file formats, returning the result as an in-memory nested dictionary objects, following the [Kerchunk references specification](https://fsspec.github.io/kerchunk/spec).
+These references can be directly read and converted into a `ManifestStore` by VirtualiZarr's :py:class:~`virtualizarr.parsers.KerchunkJSONParser` and :py:class:~`virtualizarr.parsers.KerchunkParquetParser`.
+
+You can therefore use a function which returns in-memory kerchunk JSON references inside your parser, then simply call `KerchunkJSONParser` and return the result.
+
+```{note}
+Whilst this might be the quickest way to get a custom parser working, we do not really recommend this approach, as:
```
chuckwondo: nit: add blank line

```suggestion
Whilst this might be the quickest way to get a custom parser working, we do not really recommend this approach, as:

```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+### Parsing a pre-existing index file
+
+A custom parser can parse multiple files, perhaps by passing a glob string and looking for expected file naming conventions, or by passing additional parser-specific keyword arguments.
+This can be useful for reading file formats which include some kind of additional "index" sidecar file, but don't have all the information necessary to construct the entire `ManifestStore` object from the sidecar file alone.
+
+```{note}
+If you do have some type of custom sidecar metadata file which contains all the information necessary to create the `ManifestStore`, then you should just create a custom parser for that metadata file format instead!
+Examples of this approach which come packaged with VirtualiZarr are the `DMRPPparser` and the `KerchunkJSONparser`
```
chuckwondo:
```suggestion
Examples of this approach which come packaged with VirtualiZarr are the `DMRPPparser` and the `KerchunkJSONparser`.
```

README.md
```diff
@@ -42,15 +43,20 @@ You now have a choice between using VirtualiZarr and Kerchunk: VirtualiZarr prov
 
 VirtualiZarr version 1 (mostly) achieves [feature parity](https://virtualizarr.readthedocs.io/en/latest/faq.html#how-do-virtualizarr-and-kerchunk-compare) with kerchunk's logic for combining datasets, providing an easier way to manipulate kerchunk references in memory and generate kerchunk reference files on disk.
 
+VirtualiZarr version 2 (unreleased) will bring:
```
chuckwondo, line 46: nit: this doesn't impact rendering, it just adds a bit of vertical spacing for readability of the raw text for those editing the markdown. I've made a few such suggestions elsewhere, in case you're wondering why I'm suggesting the addition of a blank line. This is something a markdown linter would suggest too.

```suggestion
VirtualiZarr version 2 (unreleased) will bring:

```

docs/api.rst
```diff
@@ -20,6 +20,24 @@ Reading
 
     open_virtual_dataset
 
+Parsers
+-------
+
+Each parser understands how to read a specific file format, and one parser must be passed to :py:func:`~virtualizarr.open_virtual_dataset`
```
chuckwondo:
```suggestion
Each parser understands how to read a specific file format, and a parser must be passed to :py:func:`~virtualizarr.open_virtual_dataset`.
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+2. Allows reading chosen variables into memory (e.g. via the `loadable_variables` kwarg, or reading from the `ManifestStore` using zarr-python directly).
+3. Provides a way to combine arrays of chunk references using a convenient API (the Xarray API).
+4. Allows persisting these references to storage for later use, in either the Kerchunk or Icechunk format.
+
+**VirtualiZarr parsers are responsible for the entirety of step (1).**
+In other words, all of the assumptions required to map the data model of an archival file format to the Zarr data model, and the logic for doing so for a specific file, together constitute a parser.
+
+**The ObjectStore instances are responsible for fetching the bytes in step (2).**
+
+This design provides a neat separation of concerns, which is helpful in two ways:
```
chuckwondo, line 61: nit: add blank line:

```suggestion
This design provides a neat separation of concerns, which is helpful in two ways:

```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+```{note}
+The object store registry can technically be empty, but to be able to read actual chunks of data back from the `ManifestStore` later the registry needs to contain at least one `ObjectStore` instance.
+
+The only time you might want to use an empty object store registry is if you are attempting to parse a custom metadata-only references format without touching the original files they refer to - i.e. a format like Kerchunk or DMR++, that doesn't contain actual binary data values.
+```
+
+## What is the responsibility of a parser?
+
+The VirtualiZarr package really does four separate things.
+In order, it:
```
chuckwondo:
```suggestion
The VirtualiZarr package really does four separate things, in order:
```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+1. The Xarray data model is subtly different from the Zarr data model (see below), so as the final objective is to create a virtual store which programmatically maps Zarr API calls to the archival file format at read-time, it is useful to separate that logic up front, before we convert to use the xarray virtual dataset representation and potentially subtly confuse matters.
+2. It also allows us to support reading data from the file via the `ManifestStore` interface, using zarr-python and obstore, but without using Xarray.
+
+## Reading data from the `ManifestStore`
+
+As well as being a well-defined representation of the archival data in the Zarr model, you can also read chunk data directly from the `ManifestStore` object.
+
+This works because the `ManifestStore` class is an implementation of the Zarr-Python `zarr.abc.Store` interface, and uses the [obstore](https://developmentseed.org/obstore/latest/) package internally to actually fetch chunk data when requested.
+
+Reading data from the `ManifestStore` can therefore be done using the zarr-python API directly
```
chuckwondo: nit: add blank line (and `:`)

```suggestion
Reading data from the `ManifestStore` can therefore be done using the zarr-python API directly:

```

docs/custom_parsers.md
```diff
@@ -0,0 +1,221 @@
...
+
+This virtual dataset object is then returned to the user, so `open_virtual_dataset` is really a very thin wrapper around the parser and object store you pass in.
+
+## Parser-specific keyword arguments
+
+The `Parser` callable does not accept arbitrary optional keyword arguments.
+
+However, extra information is often needed to fully map the archival format to the Zarr data model, for example if the format does not include array names or dimension names.
+
+Instead, to pass arbitrary extra information to your parser callable, it is recommended that you bind that information to class attributes (or use `functools.partial`), e.g.
```
chuckwondo:
```suggestion
Instead, to pass arbitrary extra information to your parser callable, it is recommended that you bind that information to class attributes (or use `functools.partial`). For example:
```

docs/usage.md
```diff
@@ -22,12 +22,14 @@ We can open a virtual representation of this file using {py:func}`open_virtual_d
 
 ```python
+from virtualizarr.parsers import HDFParser
+from obstore.store import LocalStore
 
-vds = open_virtual_dataset('air.nc')
+vds = open_virtual_dataset('air.nc', object_store=LocalStore, parser=HDFParser())
```
chuckwondo, line 26: context manager?

also, a few lines above, I suggest using a context manager for opening the air_temperature tutorial dataset

docs/usage.md
```diff
@@ -382,11 +396,13 @@ You can dis-ambuiguate kerchunk references containing relative paths by passing
 
 vds = open_virtual_dataset(
     'relative_refs.json',
-    filetype='kerchunk',
-    virtual_backend_kwargs={'fs_root': 'file:///some_directory/'}
+    object_store=LocalStore,
+    parser=KerchunkJSONParser(
+        fs_root='file:///data_directory/',
+    )
 )
```
chuckwondo, line 439: context manager?

chuckwondo (changes requested):
Great documentation! This really helped me start to wrap my head around things.

Most of my suggestions are minor format/syntax/grammar suggestions, but there are also a few regarding use of context managers in examples, and a naming question (which would be best addressed in a separate PR, if is makes sense).

README.md
TomNicholas: Adding a markdown linter to the CI seems like a good idea.

docs/custom_parsers.md
TomNicholas: On context managers: Do we really need to? It makes all the examples more complex to read...

On renaming: I agree this is potentially confusing. I think I would prefer everything be `object_store`, but then on the other hand we do have type hints to help disambiguate... Doesn't help that `zarr.storage.ObjectStore` is a `zarr.abc.Store` that wraps an `obstore.Store` 🙃 Is the word `object` redundant in any way? Might we want to generalize that later?

chuckwondo: Adding context managers would certainly add a minor amount of complexity to the examples, but my fear is that most readers of any code examples (regardless of library) tend to _repeat_ the same patterns, even if those patterns are likely _not_ ideal for production code. How many context managers have I already had to add to the codebase itself to resolve problems (both in main code and test code)?

At the very least, I recommend a very obvious, bold warning in at least one place in the docs (ideally somewhere most readers are likely to see) that very clearly indicates that use of context managers is recommended for production code, but for brevity, code examples will not use them. And the callout should show an explicit example of the recommended practice, so that the syntax is visually imprinted in the reader's mind.

chuckwondo: My preference is to make repeated use of context managers throughout the examples, so that the repetition is imprinted in the reader's mind, and will be the syntax they repeat, rather than repeatedly _not_ using context managers.

Even with a big, bold warning somewhere in the docs, I suspect the reader will repeat what they see, not what the warning says, because that's what they would repeatedly see in the examples. I recommend repeating the recommended practice, not repeating the "poor" practice simply for saving a modicum of keystrokes/simplification.

Of course, if I'm outvoted, I won't block things.

TomNicholas: That's totally reasonable. My only remaining concern is that it's tricker to do that in narrative documentation than in real code, because I need text between opening the virtual dataset and using the virtual dataset. But this isn't going to work if users copy it verbatim:

```python
with open_virtual_dataset() as vds:
    ...
```

some explanatory text

```python
vds.virtualize.to_kerchunk()
```

In the docs I can't really wrap all later uses of `vds` inside the context manager, unless I keep opening it again and again, which also wouldn't be very clear. It feels like a compromise either way.

FWIW all your arguments could apply to the xarray documentation too, but they don't use context managers there either

https://docs.xarray.dev/en/stable/user-guide/io.html#reading-and-writing-files

chuckwondo: Fair point about interleaving prose with code. Perhaps we can at least find a good place to put a callout explaining that use of context managers is strongly recommended to prevent memory/resource leaks in critical code (along with a code example), but that for convenience throughout the docs, context managers might be dropped.

TomNicholas: Added a note in https://github.com/zarr-developers/VirtualiZarr/pull/619/commits/41f38abf349e852de3a2ccd2c2f9e67a2520f0d8

docs/custom_parsers.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/custom_parsers.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/usage.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/usage.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/usage.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/usage.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/usage.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/usage.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/usage.md
TomNicholas: See https://github.com/zarr-developers/VirtualiZarr/pull/619/files#r2164943080

docs/custom_parsers.md
TomNicholas: > a section of the docs covering registries

TomNicholas: Tracked in https://github.com/zarr-developers/VirtualiZarr/issues/629

docs/usage.md
```diff
@@ -20,12 +20,29 @@ We can open a virtual representation of this file using [virtualizarr.open_virtu
...
+    VirtualiZarr has various "parsers" which understand different file formats. You must supply a parser, and as all netCDF4 files are really HDF5 files, here we used the `HDFParser`.
+
+!!! important
+
+    It is good practice to use `open_virtual_dataset` as a context manager to automatically close file handles.
+    For example the above code would become:
+
+    ```python
+    with vz.open_virtual_dataset('air.nc', object_store=LocalStore, parser=HDFParser()) as vds:
+	    # do things with vds
```
chuckwondo, line 40:
```suggestion
	# do things with vds
```

docs/usage.md
chuckwondo, line 1: Thanks for the note about recommending context managers!

docs/usage.md
chuckwondo: Weird. The indentation appears off here, but looking at the rendered docs, it looks fine, so you can ignore this suggestion.

Issue #596, opened 2025-05-15
Failing test to due to upstream indexer changes


maxrjones:
`virtualizarr/tests/test_xarray.py::TestConcat::test_concat_along_new_dim` started failing after https://github.com/pydata/xarray/pull/10277 was merged.

keewis:
this actually looks like a bug in `virtualizarr`. The problematic part is this: https://github.com/zarr-developers/VirtualiZarr/blob/43209b32b8d84f7f57f91942c5b8d3b6cf2db0f4/virtualizarr/manifests/array.py#L241-L247 where the error message says that `None` would be fine but the condition checks for `dim_indexer is None`. I'd rewrite the condition to
```python
if not isinstance(dim_indexer, (int, slice, np.ndarray)) and dim_indexer not in (None, Ellipsis):
```
(could also just change the second part to `and dim_indexer is not None` or put parentheses around `isinstance(...) or dim_indexer is None`, but `Ellipsis` is also a singleton)

maxrjones:
With xarray just being released, I think we need to address this soon. I tried the suggestion above and it didn't work as-is..

maxrjones:
There's more work needed following up on #612. For example, this fails:
```Python
import xarray as xr
import warnings
warnings.filterwarnings("ignore",
                       message="Numcodecs codecs are not in the Zarr version 3 specification*",
                       category=UserWarning)
xr.set_options(display_style="html")

from urllib.parse import urlparse

import obstore as obs
import xarray as xr

from virtualizarr import open_virtual_dataset
from virtualizarr.parsers.hdf.hdf import Parser as HDFParser

file_urls = [
    "s3://smn-ar-wrf/DATA/WRF/DET/2022/12/31/12/WRFDETAR_01H_20221231_12_000.nc",
    "s3://smn-ar-wrf/DATA/WRF/DET/2022/12/31/12/WRFDETAR_01H_20221231_12_001.nc",
]

parsed = urlparse(file_urls[0])
s3store = obs.store.S3Store(
    bucket=parsed.netloc,
    client_options={"allow_http": True},
    skip_signature=True,
    virtual_hosted_style_request=False,
    region="us-west-2",
)
virtual_datasets = [
    open_virtual_dataset(file_url=url, parser=HDFParser(), object_store=s3store)
    for url in file_urls
]

virtual_ds = xr.concat(virtual_datasets, dim='time', coords='minimal', compat='override')
print(repr(virtual_ds))
```

[...]

PR #612, opened 2025-06-11
Support ManifestArray expansion via None in the indexer

maxrjones:
This PR adds ManifestArray support this component of the array API:
> Each None in the selection tuple must expand the dimensions of the resulting selection by one dimension of size 1. The position of the added dimension must be the same as the position of None in the selection tuple.

Fixes #596

PR #641, opened 2025-06-29
Allow concatenating empty variables

maxrjones:
The dimensionality check should not apply to variables with no dimensions.

- [x] Closes https://github.com/zarr-developers/VirtualiZarr/issues/596
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


virtualizarr/tests/test_xarray.py
```diff
@@ -878,3 +880,223 @@ def test_drop_variables(netcdf4_file):
         file_url=netcdf4_file, object_store=store, parser=parser, drop_variables=["air"]
     ) as vds:
         assert "air" not in vds.variables
+
+
+def test_xarray_concat_empty_data_var(tmp_path: Path):
```
TomNicholas: This test seems extremely long and complicated to test such a tiny change. Couldn't we just test this with something like

```python
marr = create_manifest_array(shape=())
marr[indexer]
```

TomNicholas (commented):
Thank you for tracking this down @maxrjones !

However I think we need to be more careful about the expected behaviour here - look at what numpy does:

```python
In [1]: import numpy as np

In [2]: arr = np.array(3)

In [3]: arr
Out[3]: array(3)

In [4]: arr.ndim
Out[4]: 0

In [5]: arr[...]
Out[5]: array(3)

In [6]: arr[...].shape
Out[6]: ()

In [7]: arr[None]
Out[7]: array([3])

In [8]: arr[None].ndim
Out[8]: 1
```

I don't think this current change will produce that behaviour - notice the subtle difference between `None` and `Ellipsis`.

virtualizarr/tests/test_xarray.py
maxrjones: thanks, I was trying for a simplification of https://github.com/zarr-developers/VirtualiZarr/issues/596#issuecomment-2977861926 but your suggestion seems sufficient.

TomNicholas:
@maxrjones do you want me to take this PR over? It's one of the very few blockers remaining to releasing v2.0

maxrjones:
> @maxrjones do you want me to take this PR over? It's one of the very few blockers remaining to releasing v2.0

That would be amazing! you could also close this and start afresh in a new branch if you prefer.

> It's one of the very few blockers remaining to releasing v2.0

we also may need to consider backporting a fix to a patch release since this is blocking V1.* usage with newer xarray releases 😞

virtualizarr/tests/test_xarray.py
TomNicholas: I simplified the tests, and checked that the new tests would fail without the bugfix here.

maxrjones:
> I don't think this current change will produce that behaviour - notice the subtle difference between None and Ellipsis.

Did the existing fix turn out to be sufficient for these differences?

TomNicholas:
I _think_ that in conjunction with #612 it is.

Issue #638, opened 2025-06-27
Move current examples to a V1 folder and add warning


norlandrhagen:
The [listed examples](https://github.com/zarr-developers/VirtualiZarr/tree/develop/examples) are all in V1 syntax. 

We could:
- move them all to a V1 folder or 
- update all of them to V2 syntax

TomNicholas:
Can we do option 1 now (with a big disclaimer somewhere) and keep an issue open to track the need to do option 2 at some point later?

PR #644, opened 2025-06-30
Adds note that examples in `examplesV1/` are pre-2.0 VirtualiZarr

norlandrhagen:
- [x] Closes https://github.com/zarr-developers/VirtualiZarr/issues/638
Moves the current examples into a `v1` directory and adds a note.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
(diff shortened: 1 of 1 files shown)


TomNicholas:
Thank you - can we also add a note on this page?

https://virtualizarr--644.org.readthedocs.build/en/644/examples/

PR #651, opened 2025-07-01
Fix mkdocs site_url and don't use directory urls

### LINKED ISSUE #643: Old documentation links 404
I just came across a pre-mkdocs link in the wild that pointed to https://virtualizarr.readthedocs.io/en/latest/usage.html. 

With the mkdocs change that link became https://virtualizarr.readthedocs.io/en/latest/usage/ (which is just a shorthand way of writing https://virtualizarr.readthedocs.io/en/latest/usage/index.html). So the original link 404s (and some css and js loads fail as well which make it really unpleasant to look at).

I think there are a few options:

1) Do nothing - there probably aren't too many external links to virtualizarr docs yet.
2) Add redirects to go from /usage.html to usage/index.html
3) Fix the css and js loads on the 404 page so it is styled

maxrjones:
I think the should make the old usage.html relative link still work and also customize the 404 page, but there may be more debugging required based on the RTD preview.

- [ ] Closes #643
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -4,7 +4,7 @@ repo_name: zarr-developers/VirtualiZarr
 repo_url: https://github.com/zarr-developers/VirtualiZarr
 site_description: Cloud-Optimize your Scientific Data as Virtual Zarr stores, using Xarray syntax.
 site_author: VirtualiZarr developers
-site_url: https://virtualizarr.readthedocs.io/en/latest/
+site_url: !ENV [READTHEDOCS_CANONICAL_URL, 'https://virtualizarr.readthedocs.io/en/latest/']
 docs_dir: docs
 use_directory_urls: false
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,6 +8,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+        exclude: mkdocs.yml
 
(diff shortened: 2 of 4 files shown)


mkdocs.yml
```diff
@@ -4,8 +4,9 @@ repo_name: zarr-developers/VirtualiZarr
 repo_url: https://github.com/zarr-developers/VirtualiZarr
 site_description: Cloud-Optimize your Scientific Data as Virtual Zarr stores, using Xarray syntax.
 site_author: VirtualiZarr developers
-site_url: https://virtualizarr.readthedocs.io/en/stable/index.html
+site_url: https://virtualizarr.readthedocs.io/en/latest/
```
chuckwondo: Any reason you can't just do the following to avoid adding the `prep-site-url` command in `pyproject.toml`?

```
site_url: !ENV [READTHEDOCS_CANONICAL_URL, 'https://virtualizarr.readthedocs.io/en/latest/']
```

Reference: https://www.mkdocs.org/user-guide/configuration/#environment-variables

maxrjones: Amazing, thank you @chuckwondo! I tried searching for this solution but did not find that page. I applied your suggestion in [`b584a25` (#651)](https://github.com/zarr-developers/VirtualiZarr/pull/651/commits/b584a251ab97c6cd8d12b77f1ff183cbbe69ad28).

chuckwondo (approved):
LGTM!

Issue #647, opened 2025-07-01
Can't isel on VirtualiZarr of multi-dimensional NetCDF files


alexgleith:
I have what I would hope is a simple use case for VirtualiZarr, which is to create a lazy-loaded xarray of existing NetCDF files.

Code I'm using currently is this:

```python
from virtualizarr import open_virtual_mfdataset
from obstore.store import HTTPStore
from virtualizarr.parsers import HDFParser

files = [
    "https://thredds.nci.org.au/thredds/fileServer/gb6/BRAN/BRAN2020/month/ocean_temp_mth_1993_01.nc",
    "https://thredds.nci.org.au/thredds/fileServer/gb6/BRAN/BRAN2020/month/ocean_temp_mth_1993_02.nc",
]

parser = HDFParser()
store = HTTPStore(url="https://thredds.nci.org.au")

ds = open_virtual_mfdataset(files, object_store=store, parser=parser)
ds
```

This works ok, although datetimes are in float64.

<img width="716" alt="Image" src="https://github.com/user-attachments/assets/454369c7-0d01-4d4c-894c-b50b58957499" />

When it comes to selecting a single time and depth though, it fails.

```python
one = ds.temp.isel(Time=0, st_ocean=0)
one
```

With this error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 one = ds.temp.isel(Time=0, st_ocean=0)
      2 one

File /opt/homebrew/lib/python3.11/site-packages/xarray/core/dataarray.py:1555, in DataArray.isel(self, indexers, drop, missing_dims, **indexers_kwargs)
   1550     return self._from_temp_dataset(ds)
   1552 # Much faster algorithm for when all indexers are ints, slices, one-dimensional
   1553 # lists, or zero or one-dimensional np.ndarray's
-> 1555 variable = self._variable.isel(indexers, missing_dims=missing_dims)
   1556 indexes, index_variables = isel_indexes(self.xindexes, indexers)
   1558 coords = {}

File /opt/homebrew/lib/python3.11/site-packages/xarray/core/variable.py:1023, in Variable.isel(self, indexers, missing_dims, **indexers_kwargs)
   1020 indexers = drop_dims_from_indexers(indexers, self.dims, missing_dims)
   1022 key = tuple(indexers.get(dim, slice(None)) for dim in self.dims)
-> 1023 return self[key]

[...]

PR #654, opened 2025-07-02
Proof of concept: lazily loading datasets with VirtualiZarr

maxrjones:
Just showing how simple it would be to support the ask from https://github.com/zarr-developers/VirtualiZarr/issues/647. I think we should consider it. 

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/__init__.py
+++ b/virtualizarr/__init__.py
@@ -4,7 +4,11 @@ from virtualizarr.accessor import (
     VirtualiZarrDatasetAccessor,
     VirtualiZarrDataTreeAccessor,
 )
-from virtualizarr.xarray import open_virtual_dataset, open_virtual_mfdataset
+from virtualizarr.xarray import (
+    open_dataset,
+    open_virtual_dataset,
+    open_virtual_mfdataset,
+)
 
 try:
     __version__ = _version("virtualizarr")
@@ -18,4 +22,5 @@ __all__ = [
     "VirtualiZarrDataTreeAccessor",
     "open_virtual_dataset",
     "open_virtual_mfdataset",
+    "open_dataset",
 ]
--- a/virtualizarr/tests/test_xarray.py
+++ b/virtualizarr/tests/test_xarray.py
@@ -11,6 +11,7 @@ import xarray.testing as xrt
 from xarray import Dataset, open_dataset
 from xarray.core.indexes import Index
 
+from virtualizarr import open_dataset as vz_open_dataset
 from virtualizarr import open_virtual_dataset, open_virtual_mfdataset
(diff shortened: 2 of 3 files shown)


TomNicholas:
> > what I would hope is a simple use case for VirtualiZarr, which is to create a lazy-loaded xarray of existing NetCDF files.
>
> this statement is absolutely true - it is a simple use case.

I don't understand why this would be useful. It

1. isn't serialized permanently
2. doesn't involve or allow lazy concatenation (unless xarray supported lazy concatenation)
3. saves the user only 2 or 3 lines of code
4. offers nothing beyond what `ManifestStore` now offers

It increases the API surface without adding the ability to do anything new - it's totally the same as just

```python
manifest_store = HDFParser()('file.nc')
ds = xr.open_zarr(manifest_store)
```

maxrjones:
> > > what I would hope is a simple use case for VirtualiZarr, which is to create a lazy-loaded xarray of existing NetCDF files.
> > 
> > 
> > this statement is absolutely true - it is a simple use case.
> 
> I don't understand why this would be useful. It
> 
> 1. isn't serialized permanently
> 2. doesn't involve or allow lazy concatenation (unless xarray supported lazy concatenation)
> 3. saves the user only 2 or 3 lines of code
> 4. offers nothing beyond what `ManifestStore` now offers
> 
> It increases the API surface without adding the ability to do anything new - it's totally the same as just
> 
> ```python
> manifest_store = HDFParser()('file.nc')
> ds = xr.open_zarr(manifest_store)
> ```

`virtualizarr.open_mfdataset()` is where the user would actually save effort. What is the equivalent for going from `virtualizarr.open_virtual_mfdataset()` to an actual xarray dataset without temporarily serializing to kerchunk/icechunk?

TomNicholas:
My interpretation of the ask from #647 is that he had a lazy xarray data that he created without using virtualizarr at all, and wanted to somehow serialize that as virtual references. Which isn't possible unless somehow that references information is still in there somewhere to be extracted.

TomNicholas:
If xarray supported lazy concatenation then the whole thing becomes a lot more interesting.

TomNicholas:
Happy to discuss this with you further, but shall we close this for now? I don't think 3 lines of code in a dangling PR is a very good way to track this...

TomNicholas:
xref https://github.com/pydata/xarray/issues/4628 and https://github.com/pydata/xarray/pull/10402

TomNicholas:
Note that the reason I didn't put effort into that and instead made virtualizarr is that you can't serialize that entire datacube to kerchunk / icechunk if you do it this way.

PR #660, opened 2025-07-06
Use ObjectStoreRegistry in Parser protocol

### LINKED ISSUE #629: Document the object store registry
> a section of the docs covering registries

_Originally posted by @TomNicholas in https://github.com/zarr-developers/VirtualiZarr/pull/619#discussion_r2164955852_

### LINKED ISSUE #656: Align function signature for `ObjectStoreRegistry` with `object_store`'s trait
I based the `ObjectStoreRegistry` off the [trait of the same name](https://docs.rs/object_store/0.12.2/object_store/registry/trait.ObjectStoreRegistry.html) in [object_store](https://crates.io/crates/object_store), but did not perfectly align the argument names. We should fix this so that if [obstore](https://github.com/developmentseed/obstore) implements a `ObjectStoreRegistry` class via either python bindings or a pure python implementation we can rely on that rather than maintaining the code internally.

maxrjones:
This PR makes several improvements to ObjectStoreRegistry.
	
- Aligns function signatures with https://docs.rs/object_store/latest/object_store/registry/trait.ObjectStoreRegistry.html so that the code may eventually be upstreamed into obstore
- Adds examples and improve docstrings
- Parses URLs on a per-segment basis (similar to `object_store`'s implementation) so that `s3://bucket1` does not get confused with `s3://bucket`.
- Requires all URLs registered to have a valid scheme (e.g., `file://` or `s3://`)
- Updates parsers to use the ObjectStoreRegistry

PR Checklist

- [x] Closes #629
- [x] Closes #656
- [x] Tests added
- [ ] Tests passing
- [x] Full type hint coverage
- [ ] Changes are documented in `docs/releases.rst` (not needed since ObjectStoreRegistry hasn't been released as part of the public API).


virtualizarr/manifests/__init__.py
```diff
@@ -4,11 +4,13 @@
 from virtualizarr.manifests.array import ManifestArray  # type: ignore # noqa
 from virtualizarr.manifests.group import ManifestGroup  # type: ignore # noqa
 from virtualizarr.manifests.manifest import ChunkEntry, ChunkManifest  # type: ignore # noqa
-from virtualizarr.manifests.store import ManifestStore, ObjectStoreRegistry  # type: ignore # noqa
+from virtualizarr.manifests.store import ManifestStore  # type: ignore # noqa
+from virtualizarr.manifests.registry import ObjectStoreRegistry
```
maxrjones: @TomNicholas should `ObjectStoreRegistry` exist outside the `manifests` namespace?

TomNicholas: Yes I think so. It doesn't actually have anything to do with a manifest. Maybe we should move `ObjectStoreRegistry` under `virtualizarr.store`? Or `.storage`? `ManifestStore` could also go there, but I can see an argument for that being in either `.manifests` or `.store` 🤷‍♂️

TomNicholas: Or `.registry` is another valid option I guess

virtualizarr/manifests/registry.py
```diff
@@ -0,0 +1,236 @@
...
+Url: TypeAlias = str
+Path: TypeAlias = str
+
+UrlKey = namedtuple("UrlKey", ["scheme", "netloc"])
+
+
+def get_url_key(url: Url) -> UrlKey:
+    """
+    Generate the UrlKey containing a url's scheme and authority/netloc that is used a the
+    primary key's in a [ObjectStoreRegsitry.map][virtualizarr.manifests.ObjectStoreRegistry.map]
```
TomNicholas: typo

```suggestion
    primary key's in a [ObjectStoreRegistry.map][virtualizarr.manifests.ObjectStoreRegistry.map]
```

virtualizarr/manifests/registry.py
```diff
@@ -0,0 +1,236 @@
...
+
+    Raises
+    ------
+    ValueError
+        If provided Url does not contain a scheme based on [urllib.parse.urlparse][]
+    """
+    parsed = urlparse(url)
+    if not parsed.scheme:
+        raise ValueError(
+            f"Urls are expected to contain a scheme, received {url} which parsed to {parsed}"
```
TomNicholas: Can we make this a little more helpful by giving an example of a valid scheme?

virtualizarr/parsers/netcdf3.py
```diff
@@ -68,7 +68,7 @@ def __call__(
             skip_variables=self.skip_variables,
             fs_root=Path.cwd().as_uri(),
         )
-
-        registry = ObjectStoreRegistry({get_store_prefix(file_url): object_store})
+        registry = ObjectStoreRegistry()
+        registry.register(file_url, object_store)
```
TomNicholas: Is it no longer possible to just initialize the registry with the mapping you want? Is the whole `.register` API design actually the best choice for virtualizarr or are we just copying a decision made in obstore? (and in rust??)

virtualizarr/manifests/registry.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+    return UrlKey(parsed.scheme, parsed.netloc)
+
+
+class PathEntry:
```
TomNicholas: I'm not totally sure I understand why this recursive(?) data structure is necessary. Couldn't we achieve similar behaviour just by having a flat dict of `str` paths and first matching on only the first segment, then if that fails matching on `/segment1/segment2` etc.? i.e. using `.startswith()` or something like that?

virtualizarr/manifests/registry.py
```diff
@@ -0,0 +1,236 @@
...
+        NamedTuple containing the Url's scheme and authority/netloc
+
+    Raises
+    ------
+    ValueError
+        If provided Url does not contain a scheme based on [urllib.parse.urlparse][]
+    """
+    parsed = urlparse(url)
+    if not parsed.scheme:
+        raise ValueError(
```
TomNicholas: So here we are requiring a scheme - do we also want to require a prefic too?

TomNicholas (commented):

virtualizarr/manifests/registry.py
maxrjones: I ported this module from Rust - https://docs.rs/object_store/latest/src/object_store/registry.rs.html#137 and went with their design. I expect their primary motivations for the nested structure is `O(path_depth)` vs `O(total_registrations)` lookup performance and de-duplication of common prefixes.

virtualizarr/parsers/netcdf3.py
maxrjones: > or are we just copying a decision made in obstore? (and in rust??)

I am currently copying the designs in Rust's object_store, excluding DefaultObjectStoreRegistry and thread safety. It would be simple to add an `__init__` method that takes `dict[url, ObjectStore]`, we would just be committing to at least sub-classing `obstore`'s implementation (if/when it exists).

virtualizarr/manifests/registry.py
maxrjones: I'm inclined to at a minimum leave prefix restrictions to a separate PR since it's easily separable from these changes and would open up per-scheme restrictions (e.g., I don't think it makes sense to require a netloc or path/prefix for memory stores).

virtualizarr/manifests/registry.py
maxrjones: Done in [`c204ce8` (#660)](https://github.com/zarr-developers/VirtualiZarr/pull/660/commits/c204ce8cce49a5bfe6844658483610a8ac856906)

virtualizarr/parsers/kerchunk/json.py
```diff
@@ -70,16 +76,25 @@ def __call__(
...
+                    registry.register(f"{self.fs_root}{path}", object_store)
+                elif not parsed.scheme:
+                    raise ValueError(
+                        f"Detected relative path `{path}`. Parser must have `fs_root` set for "
+                        f"handling relative path. Parser has `fs_root = {self.fs_root}`."
+                    )
+                else:
+                    registry.register(path, object_store)
         else:
             registry = self.store_registry
```
maxrjones: FYI these changes are not needed if we also adopt https://github.com/zarr-developers/VirtualiZarr/pull/662

virtualizarr/parsers/kerchunk/parquet.py
```diff
@@ -90,8 +91,25 @@ def __call__(
...
+            parsed = urlparse(path)
+            if self.fs_root and not parsed.scheme:
+                registry.register(f"{self.fs_root}{path}", object_store)
+            elif not parsed.scheme:
+                raise ValueError(
+                    f"Detected relative path `{path}`. Parser must have `fs_root` set for "
+                    f"handling relative path. Parser has `fs_root = {self.fs_root}`."
+                )
+            else:
+                registry.register(path, object_store)
```
maxrjones: these changes are not needed if we also adopt https://github.com/zarr-developers/VirtualiZarr/pull/662

virtualizarr/tests/test_parsers/test_kerchunk.py
```diff
@@ -150,7 +153,10 @@ def test_handle_relative_paths(refs_file_factory):
     refs_file = refs_file_factory(chunks={"a/0.0": ["test1.nc", 6144, 48]})
```
maxrjones, line 147: the changes to `test_handle_relative_paths` are temporary if https://github.com/zarr-developers/VirtualiZarr/pull/662/ is also adopted

virtualizarr/manifests/registry.py
```diff
@@ -0,0 +1,236 @@
...
+            f"Urls are expected to contain a scheme (e.g., `file://` or `s://`), received {url} which parsed to {parsed}"
```
TomNicholas: typo
```suggestion
            f"Urls are expected to contain a scheme (e.g., `file://` or `s3://`), received {url} which parsed to {parsed}"
```

virtualizarr/registry.py
```diff
@@ -0,0 +1,239 @@
...
+        ```python exec="on" source="above" session="registry-examples"
+        from obstore.store import S3Store
+        from virtualizarr.registry import ObjectStoreRegistry
+
+        reg = ObjectStoreRegistry()
+        orig_store = S3Store(bucket="my-bucket-1", prefix="orig-path")
+        reg.register("s3://my-bucket-1", orig_store)
+
+        new_store = S3Store(bucket="my-bucket-1", prefix="updated-path")
+        store = reg.register("s3://my-bucket-1", new_store)
```
TomNicholas: We discussed yesterday that this API goes against a common python convention, that functions should either modify in-place or have a return value, but not both.

References:
- https://github.com/pydata/xarray/issues/9196
- https://stackoverflow.com/questions/58269413/why-in-principle-an-in-place-modifying-method-should-return-none

virtualizarr/parsers/netcdf3.py
maxrjones: we're no longer following the upstream implementation

docs/api.md
```diff
@@ -52,7 +52,13 @@ See the page on data structures for more information.
 ::: virtualizarr.manifests.ManifestArray
 ::: virtualizarr.manifests.ManifestGroup
 ::: virtualizarr.manifests.ManifestStore
-::: virtualizarr.manifests.ObjectStoreRegistry
+
+#### Registry
+
+::: virtualizarr.registry.Url
+[Urls][virtualizarr.registry.Url] should be parseable by [urllib.parse.urlparse][].
+::: virtualizarr.registry.UrlKey
```
TomNicholas: Why is the `UrlKey` public? It doesn't seem to be used in the signatures of any other public interfaces?

maxrjones: good call, this isn't needed in the public docs - removed in [`dae1a5e` (#660)](https://github.com/zarr-developers/VirtualiZarr/pull/660/commits/dae1a5e42c9f604b608c3338e08c425d9bedb4c5)

PR #672, opened 2025-07-13
Handle scalar array and vars named c in ManifestStore

### LINKED ISSUE #530: Generate a test case for ManifestStore with scalar arrays
This issue is to track the TODO in https://github.com/zarr-developers/VirtualiZarr/pull/523#discussion_r2023159603. Zarr will use the `<array>/c` key for accessing scalar arrays, rather than `<array>/c/0` as one might expect. Currently, we will raise a NotImplementedError but we should generate a test case and properly support scalar arrays.

### LINKED ISSUE #671: ManifestStore fails to parse chunk indexes for array keys ending with the letter c.
https://github.com/zarr-developers/VirtualiZarr/pull/523 introduced logic to handle ManifestArrays that contain a single scalar which uses the Zarr convention of using the "c" key for handling scalars.  It appears that this inadvertently treats all keys ending with "c" as scalars.  See https://github.com/zarr-developers/VirtualiZarr/commit/b5d764c224c55abcf770592fe3adf53253fce7f2 for an MVE.  @maxrjones I don't fully understand the key parsing logic here but if you have some time to circle up Monday I can take a shot a PR for this.

maxrjones:
Improves the extraction of chunk indexes from the node name/key in the Zarr hierarchy to support scalar values and variables named `c`.

- [x] Closes #530 
- [x] Closes #671
- [x] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from collections.abc import AsyncGenerator, Iterable, Mapping
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, TypeAlias
@@ -110,15 +111,24 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int, 
         Raised if the key ends with "c", indicating a scalar array, which is not yet supported.
 
     """
-    if key.endswith("c"):
-        # Scalar arrays hold the data in the "c" key
-        raise NotImplementedError(
-            "Scalar arrays are not yet supported by ManifestStore"
-        )
-    parts = key.split(
-        "c/"
-    )  # TODO: Open an issue upstream about the Zarr spec indicating this should be f"c{chunk_key_encoding}" rather than always "c/"
-    return tuple(int(ind) for ind in parts[1].split(chunk_key_encoding))
+    # Handle empty or root path
+    if not key or key == "/" or key.endswith("zarr.json"):
+        return ()
+
+    # Remove leading slash if present
+    if key.startswith("/"):
+        key = key[1:]
+
+    # Handle scalar value
+    if key.endswith("/c"):
+        return (0,)
+
+    # Look for f"/c{chunk_key_encoding"}" followed by digits and more /digits
+    match = re.search(
+        f"(?:^|/)c{chunk_key_encoding}(\d+(?:{chunk_key_encoding}\d+)*)", key
+    ).group()
+    match = match.removeprefix("/").removeprefix(f"c{chunk_key_encoding}")
+    return tuple(int(ind) for ind in match.split(chunk_key_encoding))
 
 
 class ObjectStoreRegistry:
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -125,10 +125,16 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int, 
 
     # Look for f"/c{chunk_key_encoding"}" followed by digits and more /digits
     match = re.search(
-        f"(?:^|/)c{chunk_key_encoding}(\d+(?:{chunk_key_encoding}\d+)*)", key
-    ).group()
-    match = match.removeprefix("/").removeprefix(f"c{chunk_key_encoding}")
-    return tuple(int(ind) for ind in match.split(chunk_key_encoding))
+        rf"(?:^|/)c{chunk_key_encoding}(\d+(?:{chunk_key_encoding}\d+)*)", key
+    )
+    if not match:
+        raise ValueError(
+            "Key {key} with chunk_key_encoding {chunk_key_encoding} did not match the expected pattern for nodes in the Zarr hierarchy."
+        )
+    chunk_component = (
+        match.group().removeprefix("/").removeprefix(f"c{chunk_key_encoding}")
+    )
+    return tuple(int(ind) for ind in chunk_component.split(chunk_key_encoding))
 
 
 class ObjectStoreRegistry:
--- a/virtualizarr/tests/test_manifests/test_store.py
+++ b/virtualizarr/tests/test_manifests/test_store.py
@@ -22,7 +22,7 @@ from virtualizarr.manifests import (
     ManifestStore,
     ObjectStoreRegistry,
 )
-from virtualizarr.manifests.store import get_store_prefix
+from virtualizarr.manifests.store import get_store_prefix, parse_manifest_index
 from virtualizarr.manifests.utils import create_v3_array_metadata
 from virtualizarr.tests import (
     requires_hdf5plugin,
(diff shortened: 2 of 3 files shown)


virtualizarr/manifests/store.py
```diff
@@ -110,15 +111,26 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
         Raised if the key ends with "c", indicating a scalar array, which is not yet supported.
 
     """
-    if key.endswith("c"):
-        # Scalar arrays hold the data in the "c" key
-        raise NotImplementedError(
-            "Scalar arrays are not yet supported by ManifestStore"
+    # Handle empty, root path, or scalar array
+    if not key or key == "/" or key.endswith("zarr.json") or key.endswith("/c"):
+        return ()
```
TomNicholas: There's a lot going on here. IIUC:

- `"zarr.json"` should never be encountered here - it's checked for by the calling function, so if it's found here that indicates a logic bug. So should that case raise?
- `"/c"` is the scalar case, which returns an empty tuple because that's the shape of a scalar array.
- `not key or key == "/"` confuses me - what is the behaviour of zarr meant to be when passed an empty path? Why isn't that just an invalid key?

Whatever the answer is it would be good to explain each case in comments in the code.

Also I think the docstring of this function has become out of date with its implementation - it no longer seems to do anything with the variable name...

virtualizarr/manifests/store.py
```diff
@@ -110,15 +111,26 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
+
+    # Remove leading slash if present
+    if key.startswith("/"):
+        key = key[1:]
+
+    # Look for f"/c{chunk_key_encoding"}" followed by digits and more /digits
+    match = re.search(
+        rf"(?:^|/)c{chunk_key_encoding}(\d+(?:{chunk_key_encoding}\d+)*)", key
+    )
```
TomNicholas: Over [here](https://github.com/zarr-developers/VirtualiZarr/blob/e6306a85f1f9985403d546d7c85777a59142d902/virtualizarr/manifests/manifest.py#L10) we already have a regex for matching chunk keys - it would be good to re-use that if possible.

TomNicholas: Also is it possible to compile it once and re-use it?

virtualizarr/manifests/store.py
```diff
@@ -110,15 +111,26 @@ def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int,
...
-        # Scalar arrays hold the data in the "c" key
-        raise NotImplementedError(
-            "Scalar arrays are not yet supported by ManifestStore"
+    # Handle empty, root path, or scalar array
+    if not key or key == "/" or key.endswith("zarr.json") or key.endswith("/c"):
+        return ()
+
+    # Remove leading slash if present
+    if key.startswith("/"):
+        key = key[1:]
```
TomNicholas:
```suggestion
    key = key.removeprefix("/")
```

virtualizarr/tests/test_manifests/test_store.py
```diff
@@ -35,6 +35,30 @@
...
+        (("foo/bar/c.1.23.45", "."), (1, 23, 45)),
+        (("c/bar/c.1.23.45", "."), (1, 23, 45)),
+        (("c/c/c.1.23.45", "."), (1, 23, 45)),
+        (("c1.2/bar/c.1.23.45", "."), (1, 23, 45)),
+        (("c1.2/abc/c.1.23.45", "."), (1, 23, 45)),
+        (("c1.2/abc/c", "."), (0,)),
+        (("c1.2/abc/c.0", "."), (0,)),
+        (("c1.2/abc/c/0", "/"), (0,)),
+    ],
+)
```
TomNicholas: Good idea - there are lots of sneaky possibilities here. Could even use a hypothesis test, but that's probably overkill.

virtualizarr/tests/test_manifests/test_store.py
```diff
@@ -227,12 +251,12 @@ async def test_pickling(self, local_store):
...
         # Check new store works
         observed = await local_store.get(
-            "foo/c/0.0", prototype=default_buffer_prototype()
+            "foo/c.0.0", prototype=default_buffer_prototype()
         )
         assert observed.to_bytes() == b"\x01\x02\x03\x04"
         # Check old store works
         observed = await new_store.get(
-            "foo/c/0.0", prototype=default_buffer_prototype()
+            "foo/c.0.0", prototype=default_buffer_prototype()
```
TomNicholas, line 299: None of these tests check the `"foo/c"` scalar case - we should probably add a scalar array to the memorystore and test that directly, instead of relying on our HDF parser test.

maxrjones: I added a test in [`def61d5` (#672)](https://github.com/zarr-developers/VirtualiZarr/pull/672/commits/def61d5ab3477c1aa50b03221f4d0555d7845900)

virtualizarr/manifests/store.py
maxrjones: It's not as simple as just using that version since I want to handle either `chunk_key_encoding`, but I consolidated the functionality in [`5223be9` (#672)](https://github.com/zarr-developers/VirtualiZarr/pull/672/commits/5223be9be5dbcbd527f83fcba65fdbb585a557be)

virtualizarr/manifests/store.py
maxrjones: I simplified the parsing in [`c1bc16e` (#672)](https://github.com/zarr-developers/VirtualiZarr/pull/672/commits/c1bc16e40cefa44760aa22a2be922a8b21d669ec) and [`c1bc16e` (#672)](https://github.com/zarr-developers/VirtualiZarr/pull/672/commits/c1bc16e40cefa44760aa22a2be922a8b21d669ec). I updated the docstring in [`35c8a14` (#672)](https://github.com/zarr-developers/VirtualiZarr/pull/672/commits/35c8a142dc88c202eaf9289a1ec1cb5fa8bdd1bc).

virtualizarr/manifests/store.py
```diff
@@ -88,10 +90,9 @@ def get_zarr_metadata(manifest_group: ManifestGroup, key: str) -> Buffer:
 
 def parse_manifest_index(key: str, chunk_key_encoding: str = ".") -> tuple[int, ...]:
```
chuckwondo: @maxrjones, to resolve the mypy error on line 120, you can change this line to be the following, along with adding an import for `Literal`:

```
def parse_manifest_index(key: str, chunk_key_encoding: Literal[".", "/"] = ".") -> tuple[int, ...]:
```

virtualizarr/parsers/hdf/hdf.py
```diff
@@ -175,6 +175,12 @@ def _dataset_chunk_manifest(
     if dataset.chunks is None:
         if dsid.get_offset() is None:
             chunk_manifest = ChunkManifest(entries={}, shape=dataset.shape)
+        elif dataset.shape == ():
+            chunk_manifest = ChunkManifest.from_arrays(
+                paths=np.array(filepath, dtype=np.dtypes.StringDType),
```
chuckwondo: Perhaps the type annotation for `ChunkManifest.from_arrays` could be improved, as I believe that's the basis for the mypy error here, but I'd have to dig into things and understand why it is annotated as it is to know if there might be a better way to do it.

For now, I'd say you should just add a `# type: ignore` comment on this line.

virtualizarr/tests/test_manifests/test_store.py
```diff
@@ -83,8 +122,28 @@ def _generate_manifest_store(
         fill_value=0,
     )
     manifest_array = ManifestArray(metadata=array_metadata, chunkmanifest=manifest)
+    scalar_chunk_manifest = ChunkManifest.from_arrays(
+        paths=np.array(f"{prefix}/{filepath}", dtype=np.dtypes.StringDType),
```
chuckwondo: Similar to my other comment, I suggest you just add a `# type: ignore` comment on this line.

virtualizarr/parsers/hdf/hdf.py
maxrjones: Thanks, @chuckwondo!

virtualizarr/manifests/utils.py
```diff
@@ -15,6 +15,28 @@
...
+
+    Parameters
+    ----------
+    chunk_key_encoding
+        The chunk key separator used in the Zarr store.
+
+    Returns
+    -------
+    Regular expression for a chunk key index
+    """
```
TomNicholas:
```suggestion
def construct_chunk_pattern(chunk_key_encoding: Literal[".", "/"]) -> str:
    """
    Produces a pattern for finding a chunk indices from key within a Zarr store using [re.match][] or [re.search][].

    Parameters
    ----------
    chunk_key_encoding
        The chunk key separator used in the Zarr store.

    Returns
    -------
    String representation of regular expression for a chunk key index
    """
```

TomNicholas (approved):

TomNicholas:
The sudden failures are caused by the release of zarr-python 3.10 today - see #676

PR #675, opened 2025-07-13
Register shorter vz accessor

### LINKED ISSUE #241: A better name for the ".virtualize" accessor?
I'm not super happy with the `vds.virtualize.to_<format>` accessor name. It's confusing - really the "virtualize" accessor doesn't "virtualize" anything, the references are already virtual, it just allows you to write them out somewhere.

A better name might be `.serialize.to_<format>` or even just `.write.to_<format>` (as these methods correspond to virtualizarr "writers"). The problem with those names is that they don't really indicate that you're about to enter a virtualizarr-specific namespace. But then again if you attempt to call `ds.virtualize.to_<format>()` on a non-virtual dataset you will get a clear error anyway.

See also #239. This is also a manifestation of #171.

maxrjones:
This adds a `vz` accessor registration but doesn't remove the old `virtualize` accessor to not break code.

- [ ] Closes #241 
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


virtualizarr/accessor.py
```diff
@@ -60,6 +60,7 @@ def all_datatree_variables(root: xr.DataTree) -> Generator[xr.Variable, None, No
         queue.extend(node.children.values())
 
 
+@xr.register_dataset_accessor("vz")
 @xr.register_dataset_accessor("virtualize")
 class VirtualiZarrDatasetAccessor:
```
TomNicholas: It might be nice to add a deprecation warning. Here's one way to do that:

```python
class _VirtualiZarrDatasetAccessor:
    # all the actual code
    ...

@xr.register_dataset_accessor("vz")
class VirtualiZarrDatasetAccessor(_VirtualiZarrDatasetAccessor)
    pass

@xr.register_dataset_accessor("virtualize")
class DeprecatedVirtualiZarrDatasetAccessor(_VirtualiZarrDatasetAccessor)
    def __init__(self, ds: xr.Dataset):
        super().__init__()
        warnings.warn(DeprecationWarning, "please use .vz")
```
This way if/when we eventually remove `.virtualize` entirely, the type returned by the accessor doesn't change.

TomNicholas (approved):

PR #677, opened 2025-07-15
Zarr data types refactor compat round 2

TomNicholas:
Follow-up to #618, aiming to close #676
- [x] Closes #676
- [ ] ~~Tests added~~
- [x] Tests passing
- [x] Full type hint coverage
- [x] Changes are documented in `docs/releases.rst`
- [ ] ~~New functions/methods are listed in `api.rst`~~
- [ ] ~~New functionality has documentation~~

PR #680, opened 2025-07-16
Normalize kerchunk reference to true json before comparisons

### LINKED ISSUE #679: Breaking change to Kerchunk references spec output
**The [Kerchunk references spec](https://fsspec.github.io/kerchunk/spec.html) has a weird property, and we might want to write out kerchunk references from VirtualiZarr differently to avoid it.**

The [Kerchunk v0 spec](https://fsspec.github.io/kerchunk/spec.html#version-0) describes how keys and values (for example zarr (v2) keys) are to be laid out within a `dict`. The [Kerchunk v1 spec](https://fsspec.github.io/kerchunk/spec.html#version-1) wraps this up into a larger spec with another layer of `dict`. The problem is in the v0 part.

The Kerchunk v0 spec allows for two different ways to format any dict values that themselves have multiple key-value pairs: 
- as more `dict` objects, so the whole thing is truly JSON-serializable, like this:
    ```
    {
      ".zgroup": {"zarr_format": 2},
      ".zattrs": {"Conventions": "UGRID-0.9.0\n"},
      "x/.zattrs": {"_ARRAY_DIMENSIONS": ["node"]},
      "x/.zarray": {"chunks": [9228245], "compressor": null, "dtype": "<f8"},
      "x/0": ["s3://bucket/path/file.nc", 294094376, 73825960]
    }
    ```
- that same information with the first layer of values serialized to strings, like this:
    ```
    {
      ".zgroup": "{\n    \"zarr_format\": 2\n}",
      ".zattrs": "{\n    \"Conventions\": \"UGRID-0.9.0\n\"}",
      "x/.zattrs": "{\n    \"_ARRAY_DIMENSIONS\": [\n        \"node\"\n ...",
      "x/.zarray": "{\n    \"chunks\": [\n        9228245\n    ],\n    \"compressor\": null,\n    \"dtype\": \"<f8\",\n  ...",
     "x/0": ["s3://bucket/path/file.nc", 294094376, 73825960]
    }
    ```

Currently VirtualiZarr writes out the latter form, and [some of our tests expect that](https://github.com/zarr-developers/VirtualiZarr/blob/43cff8eeb6ec56b41dcd2462ab1f423abeb610de/virtualizarr/tests/test_writers/test_kerchunk.py#L37). I think we (I) arbitrarily chose this form, not even fully understanding the Kerchunk spec at the time.

Now though, I can see that these is a problem with this partially-string-serialized form: **the order of the keys in the string-serialized parts matter**.

As @d-v-b points out in https://github.com/zarr-developers/zarr-python/issues/3254, order is not supposed to matter in JSON. But two programs aiming to produce this spec format by first creating intermediate JSON objects and serializing them (i.e. treating them as true JSON) aren't guaranteed to end up with exactly compatible serialized string results, because the order of the keys in the string parts can end up different. This has now happened in VirtualiZarr (part of #677) due to upstream changes in Zarr's metadata classes, which reasonably assume order doesn't matter (see https://github.com/zarr-developers/zarr-python/issues/3254).

This means the change in zarr-python, which by rights shouldn't matter, has technically caused a breaking change in the format that VirtualiZarr's `.to_kerchunk` writes out, which is only breaking because of an ambiguity in the Kerchunk references spec.

We have four choices I can see:

1. Try to get zarr-python to preserve order of metadata keys (I don't think we should do that because JSON doesn't require that).
2. Keep writing the same way in VirtualiZarr, adjust our tests slightly, and hence ignore the fact our output has changed. This may be okay - if everyone else using these references (i.e. likely just fsspec/kerchunk) is first deserializing them to `dicts` piecemeal, then the problem washes out.
3. Manually alter the order of keys within the serialized string part before returning. This would be a weird hacky thing to do but would avoid altering the output and avoid changing tests. It would effectively be treating key order as an important part of the Kerchunk v0 spec to be respected.
4. Change the format that VirtualiZarr's `.to_kerchunk` writes out to use the true JSON version of the Kerchunk v0 spec (i.e. fully-nested dicts). This is the "correct" solution from a data model perspective, but is also breaking, and I don't know if changing it will surface any other bugs.

I'm inclined to do (3) for now and keep this issue open in case we need to do (4) later (or in case kerchunk/fsspec needs to make any changes to handle the ambiguity).

cc @martindurant in case you have thoughts

TomNicholas:
- [x] Closes #679
- [x] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage
- [ ] ~~Changes are documented in `docs/releases.rst`~~

### DIFF
--- a/virtualizarr/tests/test_writers/test_kerchunk.py
+++ b/virtualizarr/tests/test_writers/test_kerchunk.py
@@ -8,7 +8,45 @@ from zarr.core.metadata.v2 import ArrayV2Metadata
 from conftest import ARRAYBYTES_CODEC
 from virtualizarr.manifests import ChunkManifest, ManifestArray
 from virtualizarr.tests import requires_fastparquet, requires_kerchunk
-from virtualizarr.utils import convert_v3_to_v2_metadata
+from virtualizarr.utils import convert_v3_to_v2_metadata, kerchunk_refs_as_json, JSON
+
+
+def test_deserialize_to_json():
+    refs = {
+        "version": 1,
+        "refs": {
+            ".zgroup": '{"zarr_format":2}',
+            ".zattrs": "{}",
+            "a/.zarray": '{"shape":[2,3],"chunks":[2,3],"fill_value":0,"order":"C","filters":null,"dimension_separator":".","compressor":null,"attributes":{},"zarr_format":2,"dtype":"<i8"}',
+            "a/.zattrs": '{"_ARRAY_DIMENSIONS":["x","y"]}',
+            "a/0.0": ["/test.nc", 6144, 48],
+        },
+    }
+    json_expected: JSON = {
+        "version": 1,
+        "refs": {
+            ".zgroup": {"zarr_format": 2},
+            ".zattrs": {},
+            "a/.zarray": {
+                "shape": [2,3],
+                "chunks": [2,3],
+                "fill_value": 0,
+                "order": "C" ,
+                "filters": None,
+                "dimension_separator": ".",
+                "compressor": None,
+                "attributes": {},
+                "zarr_format": 2,
+                "dtype": "<i8",
+            },
+            "a/.zattrs": {
+                "_ARRAY_DIMENSIONS": ["x", "y"],
+            },
+            "a/0.0": ["/test.nc", 6144, 48],
+        },
+    }
+    refs_as_json = kerchunk_refs_as_json(refs)
+    assert refs_as_json == json_expected
 
 
 @requires_kerchunk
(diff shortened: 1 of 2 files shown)


TomNicholas:
Everything fails because everything is broken on `develop` thanks to zarr's recent release, not because this PR breaks things. Though it's difficult to be certain until we fix it (which requires #677, which requires this PR). But the tests altered in this PR do pass.

TomNicholas:
I just fixed the typing issues locally - they are unsurprising because I had not told the type hints that this isn't just any old JSON structure, it's specific to Kerchunk.

maxrjones (approved):

PR #682, opened 2025-07-16
Temporarily restrict Zarr versions

maxrjones:
We can remove this if https://github.com/zarr-developers/VirtualiZarr/pull/677 lands before we're ready to release V2.0.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
     "numcodecs>=0.15.1",
     "ujson",
     "packaging",
-    "zarr>=3.0.8",
+    "zarr>=3.0.8,!=3.1.0",
     "obstore>=0.5.1",
 ]
 


TomNicholas:
Oh good idea. Should we pin to less than zarr 3.1.0 though? Zarr isn't about to undo the things that broke VZ.

maxrjones:
> Oh good idea. Should we pin to less than zarr 3.1.0 though? Zarr isn't about to undo the things that broke VZ.

If we did this we'd have to remove Zarr from the upstream tests so that there are no conflicts between dependency configurations. I guess we could decide based on the probability of https://github.com/zarr-developers/VirtualiZarr/pull/677 being included in the release. How much work do you think the compatibility updates will require?

PR #668, opened 2025-07-10
Add ManifestStore support for empty chunks.

sharkinsspatial:
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

Add support for ManifestStore to return None when reading "empty" chunks.

Closes #16

Issue #394, opened 2025-01-28
Refactor tests around expected properties


TomNicholas:
Discussions with @maxrjones made me realise that we're thinking about testing wrong. In particular, with #124 in place we will have well-defined **expected properties** that we can and should test against. Let me explain.

@maxrjones asked whether implementing loading via a `ManifestStore`/`FunctionalStore` as described by @ayushnag in https://github.com/zarr-developers/VirtualiZarr/issues/124 would be a problematic mixing of concerns between creating virtual references and also loading data. But actually it's the opposite: this is exactly what we should be doing. 

The reason is that _if we cannot correctly load data via a `ManifestStore`, we cannot correctly load it from an `IcechunkStore` either_. Loading from an in-memory `ManifestArray` via zarr-python reading a `ManifestStore` and loading from an in-memory/on-disk/S3 `IcechunkStore` via zarr-python performs exactly the same set of decoding steps. Contrast that to how we currently load variables via a separate xarray backend (e.g. h5netcdf) - that does it's own decoding logic that isn't encoded in the zarr model.

To check that we are able to correctly load variables, the property to test is

```python
loaded_ds_from_manifests = open_virtual_dataset('file.nc', backend='netcdf', loadable_variables='all')

loaded_ds_from_backend = xr.open_dataset('file.nc', engine='h5netcdf').load()
xrt.assert_identical(loaded_ds_from_manifests, loaded_ds_from_backend)
```

i.e., we check our creation of manifests by loading from it and comparing against loading using a pre-existing xarray backend for that filetype. @dcherian told me once it's a good thing to have multiple different ways to load data from the same file format, as it allows cross-checking. This property just encodes that. We could write a similar property for any other virtualizable file format that has an xarray backend, and each test like this should live in `virtualizarr/tests/test_readers/test_hdf.py` etc.

Similarly we could also load from an in-memory icechunk store instead, which would make the property:

```python
vds = open_virtual_dataset('file.nc', backend='netcdf')
vds.virtualize.to_icechunk(icechunkstore)
loaded_ds_from_icechunk = xr.open_zarr(icechunkstore).load()

loaded_ds_from_backend = xr.open_dataset('file.nc', engine='h5netcdf').load()
xrt.assert_identical(loaded_ds_from_manifests, loaded_ds_from_backend)
```

This now looks a lot like our existing round-trip tests, inclusive of #376 (FYI @jsignell).

[...]

PR #683, opened 2025-07-16
Do not allow exceptions in ManifestStore

maxrjones:
Zarr python allows some exceptions to pass because nodes in the hierarchy that contain only fill values can be represented by missing files. We have a different way of representing missing files after https://github.com/zarr-developers/VirtualiZarr/pull/668, so we should never be expecting to have a failed get request due to a FileNotFoundError. I don't think we should expect the other errors used in regular Zarr store implementations.

TODO: 
- [x] Fix failing DMR++ tests, which likely relate to the test data not actually existing.
- [x] Fix `TestToVirtualXarray` tests
- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -32,9 +32,6 @@ if TYPE_CHECKING:
 __all__ = ["ManifestStore"]
 
 
-_ALLOWED_EXCEPTIONS: tuple[type[Exception], ...] = ()
-
-
 @dataclass
 class StoreRequest:
     """Dataclass for matching a key to the store instance"""
@@ -267,15 +264,12 @@ class ManifestStore(Store):
         )
 
         # Actually get the bytes
-        try:
-            bytes = await store.get_range_async(
-                path_in_store,
-                start=byte_range.start,
-                end=byte_range.end,
-            )
-            return prototype.buffer.from_bytes(bytes)  # type: ignore[arg-type]
-        except _ALLOWED_EXCEPTIONS:
-            return None
+        bytes = await store.get_range_async(
+            path_in_store,
+            start=byte_range.start,
+            end=byte_range.end,
+        )
+        return prototype.buffer.from_bytes(bytes)  # type: ignore[arg-type]
 
     async def get_partial_values(
         self,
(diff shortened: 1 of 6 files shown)


virtualizarr/manifests/store.py
```diff
@@ -32,11 +32,7 @@
 __all__ = ["ManifestStore"]
 
 
-_ALLOWED_EXCEPTIONS: tuple[type[Exception], ...] = (
-    FileNotFoundError,
-    IsADirectoryError,
-    NotADirectoryError,
-)
+_ALLOWED_EXCEPTIONS: tuple[type[Exception], ...] = ()
```
TomNicholas: Can we not just delete this now?

maxrjones:
@TomNicholas the tests required major changes because we can no longer point to fake files (which would previously just return NaNs). I got 3/4 of the way through the DMR++ tests and then decided to open https://github.com/zarr-developers/VirtualiZarr/issues/685 regarding the relative file tests and remove some unit tests that seemed redundant with a better approach of property tests (https://github.com/zarr-developers/VirtualiZarr/issues/394). In summary, I'd appreciate your review with the understanding that we can put more work into improving the DMRPP tests after the release.

virtualizarr/tests/test_manifests/test_array.py
```diff
@@ -138,9 +138,9 @@ def test_broadcast_existing_axis(self, manifest_array):
         assert expanded.shape == (3, 2)
         assert expanded.chunks == (1, 2)
         assert expanded.manifest.dict() == {
-            "0.0": {"path": "file:///foo.0.0.nc", "offset": 0, "length": 5},
-            "1.0": {"path": "file:///foo.0.0.nc", "offset": 0, "length": 5},
-            "2.0": {"path": "file:///foo.0.0.nc", "offset": 0, "length": 5},
+            "0.0": {"path": "file:///foo.0.0.nc", "offset": 0, "length": 8},
+            "1.0": {"path": "file:///foo.0.0.nc", "offset": 0, "length": 8},
+            "2.0": {"path": "file:///foo.0.0.nc", "offset": 0, "length": 8},
```
TomNicholas, line 143: Why did the length of these need to change?

maxrjones: Because the chunks are emulating two 4 byte numbers.

TomNicholas: But it doesn't actually matter for these tests right? You're just making it neater.

PR #673, opened 2025-07-13
Migrate to Icechunk v1.0, exclude broken Xarray compatibility

maxrjones:
Following https://icechunk.io/en/latest/migration/#virtual-datasets

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage
- [ ] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`
- [ ] New functionality has documentation

PR #687, opened 2025-07-17
Ensure virtual chunk container points to pytest temporary directory

TomNicholas:
I find that since v1.0.0, Icechunk tests fail locally for me (on MacOS), as they are unable to find quite the right path to use for the virtual chunk container. Instead of guessing, I looked up [how pytest actually chooses the temporary directory](https://docs.pytest.org/en/stable/how-to/tmp_path.html#temporary-directory-location-and-retention), and just made a single virtual chunk container pointing at that directory.

This is really a follow-up to #673, to try and ensure it works on all machines.

### DIFF
--- a/virtualizarr/tests/test_writers/test_icechunk.py
+++ b/virtualizarr/tests/test_writers/test_icechunk.py
@@ -28,8 +28,8 @@ if TYPE_CHECKING:
     )
 
 
-# Set pytest temporary data into a known location in what should be a cross-platform way. See https://docs.pytest.org/en/stable/how-to/tmp_path.html#temporary-directory-location-and-retention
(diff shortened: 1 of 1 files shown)


virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -35,21 +42,19 @@ def icechunk_storage(tmp_path: Path) -> "Storage":
 
 @pytest.fixture(scope="function")
 def icechunk_repo(icechunk_storage: "Storage", tmp_path: Path) -> "Repository":
-    config = icechunk.RepositoryConfig.default()
 
-    url_prefixes = ["file:///private/var/folders/70", "file:///tmp/"]
```
TomNicholas, line 40: I don't know what this `/70` bit was, but that's not the correct path on my machine.

PR #696, opened 2025-07-18
Support HDF over HTTPS

maxrjones:
The HDF and DMRPP files previously didn't work when using an `HTTPStore` that was mounted using a prefix, which is almost always the case when using `obstore.from_url`. This PR fixes that by using the more robust prefix parsing code inside `ObjectStoreRegistry` rather than a separate lighter weight version in `ObstoreReader`. It also documents the inputs for `ObstoreReader` so users know to not include the prefix.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_parsers/test_hdf/test_hdf.py
+++ b/virtualizarr/tests/test_parsers/test_hdf/test_hdf.py
@@ -2,9 +2,11 @@ import h5py  # type: ignore
 import numpy as np
 import pytest
 import xarray as xr
+from obstore.store import from_url
 
 from virtualizarr import open_virtual_dataset
 from virtualizarr.parsers import HDFParser
+from virtualizarr.registry import ObjectStoreRegistry
 from virtualizarr.tests import (
     requires_hdf5plugin,
     requires_imagecodecs,
@@ -195,3 +197,13 @@ def test_subgroup_variable_names(
         parser=parser,
     ) as vds:
         assert list(vds.dims) == ["dim_0"]
+
+
+def test_netcdf_over_https():
+    url = "https://oceania.generic-mapping-tools.org/cache/Iceland.nc"
+    store = from_url(url)
+    registry = ObjectStoreRegistry({url: store})
+    parser = HDFParser()
+    with parser(file_url=url, registry=registry) as ms:
+        ds = xr.open_zarr(ms, zarr_format=3, consolidated=False).load()
+        np.testing.assert_allclose(ds["z"].min().to_numpy(), -1857)
--- a/virtualizarr/tests/test_parsers/test_hdf/test_hdf.py
+++ b/virtualizarr/tests/test_parsers/test_hdf/test_hdf.py
@@ -10,6 +10,7 @@ from virtualizarr.registry import ObjectStoreRegistry
 from virtualizarr.tests import (
     requires_hdf5plugin,
     requires_imagecodecs,
+    requires_network,
 )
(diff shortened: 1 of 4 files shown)


virtualizarr/tests/test_parsers/test_hdf/test_hdf.py
```diff
@@ -195,3 +198,14 @@ def test_subgroup_variable_names(
...
+
+
+@requires_network
+def test_netcdf_over_https():
+    url = "https://oceania.generic-mapping-tools.org/cache/Iceland.nc"
+    store = from_url(url)
+    registry = ObjectStoreRegistry({url: store})
+    parser = HDFParser()
+    with parser(file_url=url, registry=registry) as ms:
+        ds = xr.open_zarr(ms, zarr_format=3, consolidated=False).load()
```
TomNicholas: This context manager is interesting. What is it doing? Wouldn't our context manager manager @chuckwondo want it on the `open_zarr` call instead?

virtualizarr/utils.py
```diff
@@ -30,25 +29,22 @@
...
-    parsed = urlparse(path)
-    if hasattr(store, "prefix") and store.prefix:
-        filepath = os.path.basename(parsed.path)
-    else:
-        filepath = parsed.path
-
-    return filepath
-
-
 class ObstoreReader:
```
TomNicholas, line 32: Note for narrative docs: either make this public API or show people this pattern, so they know to use it within their own parsers.

maxrjones: I think this should eventually live outside VirtualiZarr (perhaps in obspec-utils) so I hesitate to make it part of the public API

virtualizarr/utils.py
```diff
@@ -30,25 +29,22 @@
...
     _reader: ReadableFile
 
     def __init__(self, store: ObjectStore, path: str) -> None:
-        import obstore as obs
-
-        filepath = remove_prefix(store, path)
-        self._reader = obs.open_reader(store, filepath)
+        """
+        Create an obstore file reader that implements the read, readall, seek, and tell methods, which
+        can be used in libraries that file-like objects.
```
chuckwondo: nit: is this a grammatical error? are you missing the word "expect" (or "require", or similar), as in "libraries that _expect_ ..."?

maxrjones: yeah, thanks I missed the word expect

virtualizarr/tests/test_parsers/test_hdf/test_hdf.py
chuckwondo: Good catch! My nagging is working 😄 

I believe we want this, although, is the call to `load` necessary?

```python
    with (
        parser(file_url=url, registry=registry) as ms,
        xr.open_zarr(ms, zarr_format=3, consolidated=False).load() as ds,
    ):
```

virtualizarr/utils.py
chuckwondo: Perhaps for a separate PR, whether this is moved to obspec-utils or not, for completeness, I suggest this class also include a `close` method (even though [obstore.ReadableFile.close](https://developmentseed.org/obstore/latest/api/file/#obstore.ReadableFile.close) is currently a no-op, perhaps it won't always be so), as well as be made into a context manager, so it can properly be used in a `with` statement.

TomNicholas: Both good points.

PR #697, opened 2025-07-18
Add docstring for open_virtual_dataset and remove unused argument

maxrjones:
- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/xarray.py
+++ b/virtualizarr/xarray.py
@@ -44,11 +44,11 @@ def open_virtual_dataset(
     indexes: Mapping[str, xr.Index] | None = None,
 ) -> xr.Dataset:
     """
-    Open a data file, virtual dataset, Zarr store, etc. as an [xarray.Dataset][] wrapping virtualized zarr arrays.
+    Open an archival data source as an [xarray.Dataset][] wrapping virtualized zarr arrays.
 
-    No data variables will be loaded unless specified in the ``loadable_variables`` kwarg (in which case they will be xarray lazily indexed arrays).
+    No data variables will be loaded unless specified in the ``loadable_variables`` kwarg (in which case they will open as lazily indexed arrays using xarray's standard lazy indexing classes).
 
-    Xarray indexes can optionally be created (the default behaviour). To avoid creating any xarray indexes pass ``indexes={}``.
+    Xarray indexes can optionally be created (the default behaviour is to create indexes for any 1D coordinate variables). To avoid creating any xarray indexes pass ``indexes={}``.
 
     Parameters
     ----------
--- a/virtualizarr/xarray.py
+++ b/virtualizarr/xarray.py
@@ -53,7 +53,7 @@ def open_virtual_dataset(
     Parameters
     ----------
     file_url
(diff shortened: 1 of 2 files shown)


virtualizarr/xarray.py
```diff
@@ -41,9 +41,52 @@ def open_virtual_dataset(
     drop_variables: Iterable[str] | None = None,
     loadable_variables: Iterable[str] | None = None,
     decode_times: bool | None = None,
-    cftime_variables: Iterable[str] | None = None,
```
maxrjones, line 44: @TomNicholas what do you think about removing this unused parameter?

TomNicholas: Get it out of here. As we're breaking some part of the API we should just do a clean break of everything old in one go.

virtualizarr/xarray.py
```diff
@@ -41,9 +41,52 @@ def open_virtual_dataset(
...
     indexes: Mapping[str, xr.Index] | None = None,
 ) -> xr.Dataset:
+    """
+    Open a data file, virtual dataset, Zarr store, etc. as an [xarray.Dataset][] wrapping virtualized zarr arrays.
```
TomNicholas: Maybe this? I definitely wouldn't say "open a virtual dataset ... as a xr.Dataset".

```suggestion
    Open an archival data source as an [xarray.Dataset][] wrapping virtualized zarr arrays.
```

virtualizarr/xarray.py
```diff
@@ -41,9 +41,52 @@ def open_virtual_dataset(
...
+
+    No data variables will be loaded unless specified in the ``loadable_variables`` kwarg (in which case they will be xarray lazily indexed arrays).
```
TomNicholas:
```suggestion
    No data variables will be loaded unless specified in the ``loadable_variables`` kwarg (in which case they will open as lazily indexed arrays using xarray's standard lazy indexing classes).
```

virtualizarr/xarray.py
```diff
@@ -41,9 +41,52 @@ def open_virtual_dataset(
...
+
+    Xarray indexes can optionally be created (the default behaviour). To avoid creating any xarray indexes pass ``indexes={}``.
```
TomNicholas:
```suggestion
    Xarray indexes can optionally be created (the default behaviour is to create indexes for any 1D coordinate variables). To avoid creating any xarray indexes pass ``indexes={}``.
```

virtualizarr/xarray.py
```diff
@@ -41,9 +41,52 @@ def open_virtual_dataset(
...
+
+    Parameters
+    ----------
+    file_url
```
TomNicholas, line 55: As per our other conversation perhaps this should just be `url`?

TomNicholas (approved):

virtualizarr/xarray.py
maxrjones: I'll standardize as `url` across the docs in #700

PR #700, opened 2025-07-18
Standardize open_virtual_*dataset/parser signature to use "url"

maxrjones:
This is a breaking change to have consistent parameter names across the library for V2.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -268,12 +268,12 @@ def netcdf4_virtual_dataset(netcdf4_file):
     from virtualizarr.parsers import HDFParser
     from virtualizarr.tests.utils import obstore_local
 
-    store = obstore_local(file_url=netcdf4_file)
+    store = obstore_local(url=netcdf4_file)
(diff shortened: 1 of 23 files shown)


maxrjones:
> If we're going to change this then shouldn't it technically be URIs, not URLs?

I think we support URLs but not all URIs, but admittedly the distinction trips me up sometimes. If you trust claude 🙈  -https://claude.ai/share/af38c494-eb3a-4400-b753-46bca9d9ca0b

TomNicholas:
I stand corrected! (And that seems consistent with a non-LLM check.) So I guess we should say URL everywhere instead of file or URI.

TomNicholas:
And to link it back to the original issue - it maybe shouldn't be `file_url`, as the url could be e.g. a zarr store directory?

Issue #690, opened 2025-07-18
Update how-to-guide/usage for new API


maxrjones:
https://virtualizarr.readthedocs.io/en/latest/usage.html needs to be updated for the new API. Similar to #689, it would be best to use a real examples that are executed on build by starting the code block with:

```raw
```python exec="on" session="usage"
```

the session can be shared between examples on the page. the above setting does not show the source code. The following setting will show the source code:
```raw
```python exec="on" source="above" session="usage" html="true"
```

PR #701, opened 2025-07-18
Updating usage doc to v2

norlandrhagen:
Closes: https://github.com/zarr-developers/VirtualiZarr/issues/690

To Do:
- [x] open_virtual_mfdataset globbing fails
- [x] Cleanup docs deps.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -1,47 +1,52 @@
 # Usage
 
-This page explains how to use VirtualiZarr. To understand how the functionality actually works, read the page on [Data Structures](data_structures.md).
+This page explains how to use VirtualiZarr. Review the [Data Structures](data_structures.md) documentation if you want to understand the conceptual models underpinning VirtualiZarr.
 
 ## Opening files as virtual datasets
 
-VirtualiZarr is for manipulating "virtual" references to pre-existing data stored on disk in a variety of formats, by representing it in terms of the [Zarr data model](https://zarr-specs.readthedocs.io/en/latest/specs.html) of chunked N-dimensional arrays.
+VirtualiZarr is for creating and manipulating "virtual" references to pre-existing data stored in the cloud or on disk in a variety of formats, by representing it in terms of the [Zarr data model](https://zarr-specs.readthedocs.io/en/latest/specs.html) of chunked N-dimensional arrays.
 
-If we have a pre-existing netCDF file on disk:
-
-```python exec="on" session="usage" source="material-block" result="code"
-import xarray as xr
-
-# create an example pre-existing netCDF4 file
-ds = xr.tutorial.open_dataset('air_temperature')
-ds.to_netcdf('air.nc')
-print(ds)
-```
-
-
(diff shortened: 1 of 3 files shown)


docs/usage.md
```diff
@@ -99,114 +92,56 @@ vds.vz.nbytes
...
+```python exec="on" session="usage" source="material-block" result="code"
 vds = open_virtual_dataset(
-    'air.nc',
-    object_store=LocalStore,
-    parser=HDFParser(),
-    loadable_variables=['air', 'time'],
+    url=url,
+    registry=registry,
+    parser=parser,
+    loadable_variables=['time'],
```
maxrjones, line 133: @TomNicholas do you think we still need the section on "Loading CF-encoded time variables", now that we use the default loading of xarray (which includes 1D time coordinates) and open_zarr (which defaults to `decode_times=True`)

docs/usage.md
```diff
@@ -442,33 +385,3 @@ vds = open_virtual_dataset(
...
-renamed_vds = vds.vz.rename_paths(local_to_s3_url)
-renamed_vds['air'].data.manifest.dict()
-```
-
-```
-{'0.0.0': {'path': 'http://s3.amazonaws.com/my_bucket/air.nc', 'offset': 15419, 'length': 7738000}}
-```
-
-!!! note
-    Icechunk does not yet support [chunk references with HTTP URLs](https://github.com/earth-mover/icechunk/issues/526) (only local files and S3 URIs), so the ability to rename paths can be useful for renaming paths into a form Icechunk accepts.
```
maxrjones, line 474: FYI this section was not removed, just moved up.

docs/usage.md
```diff
@@ -399,30 +297,75 @@ This is especially useful for datasets that grow over time.
...
+
+```python
+combined_vds.vz.to_kerchunk('output/combined.json', format='json')
+```
+
+These zarr-like references can now be interpreted by [fsspec](https://github.com/fsspec/filesystem_spec), using kerchunk's built-in xarray backend (kerchunk must be installed to use `engine='kerchunk'`).
+
+```python
+combined_ds = xr.open_dataset('output/combined.json', engine="kerchunk")
+print(combined_ds)
```
maxrjones, line 324: For the dataset used in this usage guide, it is not advisable to write it to a Kerchunk JSON because one of the two bytesbytescodecs will get cut off. I am tempted to just remove this section, as mentioned in slack. We could later add a section that loads a Kerchunk JSON from object storage, which is a more recommended use case.

docs/usage.md
TomNicholas: It's far less likely to come up now, but technically we do still need this. Someone could have a variable that is not one that will be loaded by default (because it's a data variable, or a >1D coordinate variable, or it is a 1D coordinate variable that has a different name from its dimension).

maxrjones:
I'm going to approve and merge this even though ideally there will be likely follow-up work on the comments above because it's an important improvement.

maxrjones (approved):

PR #713, opened 2025-07-20
Include all V2 codecs as filters

maxrjones:
This PR applies @d-v-b's suggestion from https://github.com/zarr-developers/numcodecs/issues/767#issuecomment-3094441284 to use the filters argument in ArrayV2Metadata to store all codecs, to avoid a regression where we truncate the number of BytesBytesCodecs.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/utils.py
+++ b/virtualizarr/utils.py
@@ -7,10 +7,9 @@ import json
 from typing import TYPE_CHECKING, Any, Iterable, Mapping, Optional, Sequence, Union
 
 import obstore as obs
-from zarr.abc.codec import ArrayArrayCodec, BytesBytesCodec
 from zarr.core.metadata import ArrayV2Metadata, ArrayV3Metadata
 
-from virtualizarr.codecs import extract_codecs, get_codec_config
+from virtualizarr.codecs import get_codec_config
 from virtualizarr.types.kerchunk import KerchunkStoreRefs
 
 # taken from zarr.core.common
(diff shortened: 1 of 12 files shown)


virtualizarr/utils.py
```diff
@@ -132,32 +131,31 @@ def convert_v3_to_v2_metadata(
...
+                f"Expected a valid Zarr V2 or V3 codec dict, got {codec_config}"
+            )
 
+    # TODO: Find a more robust way to exclude any bytes codecs
+    # TODO: Test round-tripping big endian since that is stored in the bytes codec in V3; it should be included in data type instead for V2
+    v2_codecs = [
+        _to_v2_codec(get_codec_config(codec))
+        for codec in v3_metadata.codecs
+        if codec.__class__.__name__ != "BytesCodec"
+    ]
```
maxrjones: @d-v-b are all `ArrayBytesCodecs` irrelevant for Zarr V2 where we could exclude based on whether they are a `ArrayBytesCodec` instance?

TomNicholas (approved):
Nice.

 I do think this NISAR test should never have been merged - a unit test that downloads 0.2GB of data is ridiculous. Then we should not need any concept of slow tests. But in the interests of expediency we can wait to swap that out in a different PR, as it's not an urgent issue.

PR #720, opened 2025-07-21
Do not pass an empty filters list to ArrayV2Metadata

### LINKED ISSUE #695: Empty filters list in JSON reference file produces warnings
Hi Virtualizarr Team, 

I've been creating virtual data set (VDS) reference files in JSON format for a while now, and this is the first time I ran into this warning. 

I created the VDS fine, and was able to open and use it (notebook [here](https://github.com/DeanHenze/cloud_earthdata/blob/main/virtualizarr/vds_cases_notebooks/vds_ECCO_L4_TEMP_SALINITY_05DEG_DAILY_V4R4.ipynb)). But when opening, I get the following message:

```
[/opt/coiled/env/lib/python3.13/site-packages/zarr/core/metadata/v2.py:192](https://cluster-zzkxl.dask.host/opt/coiled/env/lib/python3.13/site-packages/zarr/core/metadata/v2.py#line=191): UserWarning: Found an empty list of filters in the array metadata document. This is contrary to the Zarr V2 specification, and will cause an error in the future. Use None (or Null in a JSON document) instead of an empty list of filters.
  warnings.warn(msg, UserWarning, stacklevel=1)
```

To my knowledge this isn't something I can specify in the calls to `open_virtual_dataset()`, e.g. I cannot specify "use Null rather than empty list for filters". Or I didn't see it in the API. So it would this be something that virtualizarr should do automatically?

maxrjones:
The filters argument should be either a populated list or `None`, not an empty list.

- [ ] Closes #695
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/utils.py
+++ b/virtualizarr/utils.py
@@ -133,8 +133,8 @@ def convert_v3_to_v2_metadata(
         The metadata object in v2 format.
     """
 
-    # TODO: Find a more robust way to exclude any bytes codecs
-    # TODO: Test round-tripping big endian since that is stored in the bytes codec in V3; it should be included in data type instead for V2
+    # TODO: Check that all ArrayBytesCodecs should in fact be excluded for V2 metadata storage.
+    # TODO: Test round-tripping big endian since that is stored in the bytes codec in V3; it should be included in data type instead for V2.
     v2_codecs = [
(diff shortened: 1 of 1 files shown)


TomNicholas:
This seems fine to me but I'm not very familiar with this part of the spec.

PR #731, opened 2025-07-23
Add comment on skip_signature=True in the s3 store example

jbusecke:
I am refactoring some of my code to virtualize NASA data to v2 at the moment, and just spent a bunch of minutes debugging a permission denied error due to me just copy/pasting the code in the example. 

Wondering if it would be instructive for users to know that this option is specific to the data used.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -39,7 +39,7 @@ for reading the original data, but some parsers may accept an empty [ObjectStore
     store = S3Store(
         bucket=bucket,
         region="us-west-2",
-        skip_signature=True #required for this specific example data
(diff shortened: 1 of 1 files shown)


docs/migration_guide.md
```diff
@@ -39,7 +39,7 @@ for reading the original data, but some parsers may accept an empty [ObjectStore
     store = S3Store(
         bucket=bucket,
         region="us-west-2",
-        skip_signature=True
+        skip_signature=True #required for this specific example data
```
maxrjones:
```suggestion
        skip_signature=True # required for this specific example data because the data is in a public bucket, so the S3Store shouldn't fetch and use credentials.
```

PR #735, opened 2025-07-24
[testing] Swaps out path for xarray anon s3 test

norlandrhagen:
- [x] Tests added
- [ ] Tests passing


TomNicholas:
Is the `slow-test` actually being run in the CI?

PR #736, opened 2025-07-25
Run slow tests in ci based on test-slow label

maxrjones:
Addresses https://github.com/zarr-developers/VirtualiZarr/pull/735#issuecomment-3115317181

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
         environment: [test-py311, test-py312, min-deps, minio]
     steps:
       - uses: actions/checkout@v4
-      - uses: prefix-dev/setup-pixi@v0.8.3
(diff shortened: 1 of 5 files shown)


maxrjones:
See https://github.com/zarr-developers/zarr-python/pull/3264#issuecomment-3115427902 for an explanation of the failing upstream tests

maxrjones:
> I'm not sure why we need `slow-tests` at all. We already have the tests requiring a network connection specifically marked - IIUC the "slow" ones are basically those requiring a network connection which download a big file. I would prefer that we just change those tests to not download a big file. Am I missing something?

I think there's utility in end-to-end tests in addition to unit/regression tests, but we could move these outside the virtualizarr repo.

Issue #650, opened 2025-07-01
Mkdocs alternatate tabs


norlandrhagen:
In the migration guide and other docs such as Icechunk you can enable alternative tabs in mkdocs. This allows you the reader to easily swtich between multiple code block options. 

ex. from the Icechunk docs:

<img width="918" alt="Image" src="https://github.com/user-attachments/assets/8759b24e-adcc-4bba-b3d2-bf77f33506ea" />
 

This is enabled with this added to the `mkdocs.yaml` file

```md
-  pymdownx.tabbed:
      alternate_style: true
```

This could be a nice bit to add throughout our docs where we outline multiple options. (local vs s3 etc.)

PR #741, opened 2025-07-28
Adds Mkdocs alternatate tabs to show multiple store configurations on Usage.md

norlandrhagen:
- [x] https://github.com/zarr-developers/VirtualiZarr/issues/650

Wondering your thoughts on this @TomNicholas and @maxrjones. Do we want to show tabs for multiple store configs and should they be on the usage page or somewhere else in the docs?

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -53,26 +53,33 @@ for reading the original data, but some parsers may accept an empty [ObjectStore
 
 === "Local Store"
 
-    ```python
-    from pathlib import Path
+```python
 
-    from obstore.store import LocalStore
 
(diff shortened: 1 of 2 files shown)


TomNicholas:
I like this! I think this and a separate page for obstore gotchas would both be useful.

norlandrhagen:
> I like this! I think this and a separate page for obstore gotchas would both be useful.

Sounds good! Should we do that in a separate PR or include it here?

TomNicholas:
The obstore gotchas can be a separate PR.

TomNicholas (approved):

PR #745, opened 2025-07-29
Don't write empty chunks to icechunk

### LINKED ISSUE #740: Trying to append to icechunk fails
Hi,

I've been working trying to virtualize GOES data (see #728) and have hit another snag. Not sure if this is an Icechunk bug, or Virtualizarr, so happy to move it to the Icechunk repo if desired. 

Using the current `main` branch of Virtualizarr,qnd icechunk 1.0.3 running the following code results in an error when appending the second virtual dataset to the Icechunk store. This happens with other files too. On the other hand, when I use `open_virtual_mfdataset` on these files and write it to an icechunk store at once, it works, but not when doing the files one after another.

```python
import icechunk
from icechunk import local_filesystem_storage
from virtualizarr.xarray import open_virtual_dataset
from obstore.store import from_url

path1 = "s3://noaa-goes19/ABI-L2-MCMIPC/2025/001/00/OR_ABI-L2-MCMIPC-M6_G19_s20250010001173_e20250010003557_c20250010004071.nc"
path2 = "s3://noaa-goes19/ABI-L2-MCMIPC/2025/001/00/OR_ABI-L2-MCMIPC-M6_G19_s20250010006173_e20250010008546_c20250010009059.nc"
parser = HDFParser()
bucket = "s3://noaa-goes19"
store = from_url(bucket, skip_signature=True)
registry = ObjectStoreRegistry({bucket: store})

storage = local_filesystem_storage("vds_test.icechunk")
repo = icechunk.Repository.open_or_create(storage)
session = repo.writable_session("main")
vds = open_virtual_dataset(path1, parser=parser, registry=registry, loadable_variables=["t", "x", "y"])
vds = vds.expand_dims("t")
print(vds)
vds.vz.to_icechunk(session.store)
session.commit("Add first file")
vds2 = open_virtual_dataset(path2, parser=parser, registry=registry, loadable_variables=["t", "x", "y"])
vds2 = vds2.expand_dims("t")
print(vds2)
session = repo.writable_session("main")
vds2.vz.to_icechunk(session.store, append_dim="t")
session.commit("Add second file")
```

This results in the following error

```python
  File "/Users/jacob/Development/planetary-datasets/dags/assets/icechunk_prover.py", line 33, in <module>
    snapshot_id = session.commit("Add second file")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jacob/Development/planetary-datasets/.pixi/envs/default/lib/python3.12/site-packages/icechunk/session.py", line 226, in commit
    return self._session.commit(
           ^^^^^^^^^^^^^^^^^^^^^
icechunk.IcechunkError:   x session error: failed to create manifest from chunk stream
  | 
  | context:
  |    0: icechunk::session::_commit
  |            with Add second file rewrite_manifests=false
  |              at icechunk/src/session.rs:987
  |    1: icechunk::session::commit
  |            with Add second file
  |              at icechunk/src/session.rs:950
  | 
```

TomNicholas:
This should fix the problem exposed in #740

- [x] Closes #740
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -8,6 +8,8 @@
 
 ### Bug fixes
 
+- Fixed bug caused by writing empty virtual chunks to Icechunk ([#745](https://github.com/zarr-developers/VirtualiZarr/pull/745)).
+  By [Tom Nicholas](https://github.com/TomNicholas).
 - Rewrote the internals of `ManifestArray.__getitem__` to ensure it actually obeys the array API standard under myriad edge cases ([#734](https://github.com/zarr-developers/VirtualiZarr/pull/734)).
   By [Tom Nicholas](https://github.com/TomNicholas).
 
--- a/virtualizarr/tests/test_writers/test_icechunk.py
+++ b/virtualizarr/tests/test_writers/test_icechunk.py
@@ -890,9 +890,7 @@ def test_write_empty_chunk(
     # regression test for https://github.com/zarr-developers/VirtualiZarr/issues/740
 
     # ManifestArray containing empty chunk
-    manifest = ChunkManifest(
-        {"0": {"path": "", "offset": 0, "length": 0}}
-    )
+    manifest = ChunkManifest({"0": {"path": "", "offset": 0, "length": 0}})
     metadata = array_v3_metadata(
         shape=(5,),
         chunks=(5,),
@@ -906,7 +904,9 @@ def test_write_empty_chunk(
     vds.vz.to_icechunk(icechunk_filestore)
 
     # when opened they should be treated as fill_value
-    roundtrip = xr.open_zarr(icechunk_filestore, zarr_format=3, consolidated=False, chunks={})
+    roundtrip = xr.open_zarr(
+        icechunk_filestore, zarr_format=3, consolidated=False, chunks={}
+    )
     expected_values = np.full(shape=(5,), fill_value=10, dtype=np.dtype("int32"))
     expected = xr.Variable(data=expected_values, dims=["x"])
     xrt.assert_identical(roundtrip["a"].variable, expected)
--- a/virtualizarr/tests/test_writers/test_icechunk.py
+++ b/virtualizarr/tests/test_writers/test_icechunk.py
@@ -881,3 +881,32 @@ class TestAppend:
 
 
 # TODO test with S3 / minio
+
+
+def test_write_empty_chunk(
+    icechunk_filestore: "IcechunkStore",
+    array_v3_metadata,
+):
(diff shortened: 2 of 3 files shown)


maxrjones (approved):

TomNicholas:
I agree that Zarr conflating uninitialized chunks with chunks containing only fill values is really annoying. But there is nothing we can do here  unless the Zarr format gains a way to differentiate?

maxrjones:
> I agree that Zarr conflating uninitialized chunks with chunks containing only fill values is really annoying. But there is nothing we can do here  unless the Zarr format gains a way to differentiate?

Does icechunk have a better way to differentiate? I tried reading through the IC spec but could not tell how that's handled. It seems like pure fill value chunks could fit as a specific type of chunk ref though, even before there's a pure Zarr solution.

d-v-b:
> I agree that Zarr conflating uninitialized chunks with chunks containing only fill values is really annoying. But there is nothing we can do here unless the Zarr format gains a way to differentiate?

How would this work exactly? It's pretty important that creating a Zarr array only requires writing a single metadata document. This entails that the _entire array_ (based on just 1 metadata document) must be consistently readable, even though no data has been written. This means that there must be some specification for what's inside the array, and that can only be contained in the metadata document. That value must be consistent with the data type for the array. This leads pretty directly to the `fill_value` metadata field, and the behavior we have today. I'm not sure how else you could achieve this, given the constraints I mentioned

dcherian:
> uninitialized chunks with chunks containing only fill values

FWIW in netCDF this is the historical meaning of `_FillValue`: https://docs.unidata.ucar.edu/nug/2.0-draft/nug_conventions.html#FillValue.

> How would this work exactly? It's pretty important that creating a Zarr array only requires writing a single metadata document. This entails that the entire array (based on just 1 metadata document) must be consistently readable, even though no data has been written.

I agree. You could return uninitialized memory so `np.empty` instead of `np.full`.  Historically "uninitialized" could mean "just allocated" (like NoFill mode in netCDF or `np.empty`). This behaviour is useful *at write-time* when you know that you will write every byte of uninitialized memory, so overwriting with a fill value isn't necessary. But I dont see how this translates to Zarr since we assemble chunks in memory, then flush it out.

dcherian:
> In Zarr, the point of fill_value is to indicate that data has not been written. 

AFAICT some substantial fraction of the Zarr community uses it to indicate "valid but background value", hence the historical choice of **0** as default fill value (!!!!)

PR #765, opened 2025-08-09
Unpin xarray

### LINKED ISSUE #688: Regression in xarray + icechunk test - now opening data via dask
As of xarray `2025.7.0` some of our [icechunk roundtrip tests fail](https://github.com/zarr-developers/VirtualiZarr/actions/runs/16345796636/job/46179016222#step:5:1713), because they open data as dask arrays instead of lazy xarray arrays. The simplest failing test is [`virtualizarr/tests/test_writers/test_icechunk.py::test_set_single_virtual_ref_with_encoding`](https://github.com/zarr-developers/VirtualiZarr/blob/1a00c95daa2e44e1ba843d96b94ba1e5e013efb2/virtualizarr/tests/test_writers/test_icechunk.py#L168).

I'm having trouble simplifying the issue. It seems to require all of these factors to be involved:
- Some decoding to happen (the similar `test_icechunk.py::test_set_single_virtual_ref_without_encoding` test passes)
- Icechunk to be involved (replacing Icechunk with a zarr MemoryStore makes it pass)
- Virtual references to be involved (rewriting the test to write native chunks instead of virtual chunks makes it pass)

For the moment we have pinned our xarray dependency to avoid this (in #673), but we do need to fix it.

My suspicion is that something in xarray has effectively changed the default behaviour of the `chunks` kwarg to `open_dataset`, but it somehow only triggers in a very specific situation involving encoding.

My next step is going to be `git bisect`-ing xarray.

### LINKED ISSUE #738: Unpin Xarray?
I see that Xarray is pinned to `==2025..6.0`, could that be relaxed?

https://github.com/zarr-developers/VirtualiZarr/blob/8230c365ebb4f44ad9393d508e0ea89b976c5c74/pyproject.toml#L23

maxrjones:
This PR addresses the failures on xarray>=2025.7.1. The tests are failing because https://github.com/pydata/xarray/commit/3679a5df71ccaf3a37a62bcd483870d5d1ae9da2 changes the contents of NetCDF files written by xarray. I addressed these failures in two ways:
- For icechunk tests, use synthetic data rather than hard-coded byte ranges for chunks inside an example NetCDF4 file
- For DMR++ tests, just update the byte ranges and xfail on older versions of xarray

- [ ] Closes #688 
- [ ] Closes #738
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage
* Always strip preceding / from path in ManifestStore

* Fix docs error

### DIFF
--- a/virtualizarr/tests/test_writers/test_icechunk.py
+++ b/virtualizarr/tests/test_writers/test_icechunk.py
@@ -881,3 +881,32 @@ class TestAppend:
 
 
 # TODO test with S3 / minio
+
+
+def test_write_empty_chunk(
+    icechunk_filestore: "IcechunkStore",
+    array_v3_metadata,
+):
+    # regression test for https://github.com/zarr-developers/VirtualiZarr/issues/740
+
+    # ManifestArray containing empty chunk
+    manifest = ChunkManifest({"0": {"path": "", "offset": 0, "length": 0}})
+    metadata = array_v3_metadata(
+        shape=(5,),
+        chunks=(5,),
+        data_type=np.dtype("int32"),
+        fill_value=10,
+    )
+    marr = ManifestArray(chunkmanifest=manifest, metadata=metadata)
+    vds = xr.Dataset({"a": ("x", marr)})
+
+    # empty chunks should never be written
+    vds.vz.to_icechunk(icechunk_filestore)
+
+    # when opened they should be treated as fill_value
+    roundtrip = xr.open_zarr(
+        icechunk_filestore, zarr_format=3, consolidated=False, chunks={}
+    )
+    expected_values = np.full(shape=(5,), fill_value=10, dtype=np.dtype("int32"))
+    expected = xr.Variable(data=expected_values, dims=["x"])
+    xrt.assert_identical(roundtrip["a"].variable, expected)
--- a/virtualizarr/tests/test_writers/test_icechunk.py
+++ b/virtualizarr/tests/test_writers/test_icechunk.py
@@ -3,19 +3,22 @@ from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Optional
 
+import numcodecs
 import numpy as np
 import numpy.testing as npt
+import obstore
 import pytest
 import xarray as xr
 import xarray.testing as xrt
 import zarr
-from zarr.core.buffer import default_buffer_prototype
+from zarr.codecs import BytesCodec
 from zarr.core.metadata import ArrayV3Metadata
+from zarr.dtype import parse_data_type
 
+from virtualizarr.codecs import zarr_codec_config_to_v3
 from virtualizarr.manifests import ChunkManifest, ManifestArray
 from virtualizarr.tests.utils import PYTEST_TMP_DIRECTORY_URL_PREFIX
 from virtualizarr.writers.icechunk import generate_chunk_key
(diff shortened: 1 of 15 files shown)


virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -51,6 +54,158 @@ def icechunk_repo(icechunk_storage: "Storage", tmp_path: Path) -> "Repository":
     )
 
 
+@pytest.fixture()
+def synthetic_vds(tmpdir: Path):
```
TomNicholas: You realise that we could replace the use of netCDF files in a huge number of our other tests with these??? For now let's just put these fixtures outside of `test_icechunk.py` so we can consolidate them later, because nothing about these is specific to icechunk.

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -112,179 +266,60 @@ def test_write_new_virtual_variable(
...
-        ) as actual_ds:
-            # Because we encode attributes, attributes may differ, for example
-            # actual_range for expected_ds.air is array([185.16, 322.1 ], dtype=float32)
-            # but encoded it is [185.16000366210935, 322.1000061035156]
-            xrt.assert_allclose(actual_ds, expected_ds)
+        ) as ds,
+    ):
+        np.testing.assert_equal(
+            ds["foo"].data, arr * 2
+        )  # Multiple original array by 2 to match scale factor
```
TomNicholas: Wait what is this? This is to make up for having dropped encoding?

TomNicholas: I don't understand why the old test didn't need this correction but the new test does

maxrjones:
```suggestion
        # We wrote a numpy array to a file and added encoding={"scale_factor": 2} to the
        # metadata. So, we expect the array loaded by xarray to be twice the magnitude of
        # the original numpy array if writing and applying the encoding is working properly.
        np.testing.assert_equal(
            ds["foo"].data, arr * 2
        )
```

maxrjones: Does https://github.com/zarr-developers/VirtualiZarr/pull/765/files#r2274474121 make more sense?

maxrjones: no array manipulation happened in the old version of the tests because both the expected and observed arrays were opened using xarray

TomNicholas: Ohh nice thank you!

TomNicholas: Riiight okay, thanks for the clarification.

virtualizarr/tests/test_writers/test_icechunk.py
TomNicholas: I did this in https://github.com/zarr-developers/VirtualiZarr/pull/765/commits/dee3cdedec99c390d7dc71cfd6d9b1973982517c

TomNicholas: and https://github.com/zarr-developers/VirtualiZarr/pull/765/commits/b86cfce73217ec90d659f9f36ad8161aa185c99e

TomNicholas (approved):
I don't know what's going on with pixi in the docs build but I think this is good to merge now.

PR #766, opened 2025-08-09
Include serializer from ManifestStore in Icechunk virtual references

### LINKED ISSUE #750: NetCDF3Parser workflow returning gibberish (while equivalent HDFParser workflow works fine)
I have a local NetCDF3 file that I opened as a virtual dataset using the virtualizarr=2 `NetCDF3Parser`, saved references to parquet, and then re-opened and accessed via the references.   

The workflow throws no errors, but the data after opening the references is gibberish.    

If I convert the NetCDF3 file to NetCDF4 and use the `HDFParser` it works fine. 

https://nbviewer.org/gist/rsignell/5d95c210e8e2b78d85269893cd89a85c

The files I used in this workflow can be obtained from:
https://usgs.osn.mghpcc.org/esip/rsignell/testing/vzarr2/test_netcdf3.nc
https://usgs.osn.mghpcc.org/esip/rsignell/testing/vzarr2/test_netcdf4.nc

Does this look like user error or a bug?

### LINKED ISSUE #762: ArrayBytesCodec not included in Icechunk serialization
Currently we do not include the `ArrayBytesCodec`s  in the codec pipeline when creating the Icechunk array https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/writers/icechunk.py#L342-L352.  This can be remedied as suggested by @maxrjones and @d-v-b by including the `serializer` arg.  We also don't currently have any full roundtrip tests that demonstrate serializing an array that has an `ArrayBytesCodec` to Icechunk so that should be included as well.  I have a PR for this but it appears to be encountering an issue with Icechunk decoding that I will detail in another issue.

maxrjones:
I think we need much more work around our handling of codecs (which would get much easier with https://github.com/zarr-developers/zarr-python/pull/3276), but this is a more immediate fix to support parsers that specify any serializer other than the default (e.g., big-endian bytes).

This makes your NetCDF3 example work @rsignell 

- [ ] Closes #762
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -179,12 +179,14 @@ class ManifestStore(Store):
             key, marr.metadata.chunk_key_encoding.separator
         )
 
+        if manifest._paths.shape == (1,) and chunk_indexes == ():
+            # TODO: Kerchunk parsers (e.g., NetCDF3) are returning 1 dimensional arrays for scalar arrays, this should be addressed at the source of the issue rather than as a workaround
+            chunk_indexes = (0,)
         path = manifest._paths[chunk_indexes]
         if path == "":
             return None
         offset = manifest._offsets[chunk_indexes]
         length = manifest._lengths[chunk_indexes]
-
         # Get the configured object store instance that matches the path
         store, path_after_prefix = self._registry.resolve(path)
         if not store:
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -179,9 +179,6 @@ class ManifestStore(Store):
             key, marr.metadata.chunk_key_encoding.separator
         )
 
-        if manifest._paths.shape == (1,) and chunk_indexes == ():
-            # TODO: Kerchunk parsers (e.g., NetCDF3) are returning 1 dimensional arrays for scalar arrays, this should be addressed at the source of the issue rather than as a workaround
-            chunk_indexes = (0,)
         path = manifest._paths[chunk_indexes]
         if path == "":
             return None
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -8,7 +8,7 @@
 
 ### Bug fixes
 
-- Fix handling of big-endian data in Icechunk by making sure that non-default zarr serializers are included in the zarr array metadata [#766](https://github.com/zarr-developers/VirtualiZarr/issues/766)). By [Max Jones](https://github.com/maxrjones)
+- Fix handling of big-endian data in Icechunk by making sure that non-default zarr serializers are included in the zarr array metadata [#766](https://github.com/zarr-developers/VirtualiZarr/issues/766). By [Max Jones](https://github.com/maxrjones)
 
 ### Documentation
 
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -8,6 +8,8 @@
 
 ### Bug fixes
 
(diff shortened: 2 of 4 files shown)


maxrjones:
> @maxrjones I ran my workflow and still got gibberish: [nbviewer.org/gist/rsignell/32f563cfbe83c20ac01aba00190e5912](https://nbviewer.org/gist/rsignell/32f563cfbe83c20ac01aba00190e5912) Did I need to update some other stuff in addition to virtualizarr? Or perhaps more likely I did something else wrong. 😕

Yeah, this fix is only for Icechunk. Kerchunk uses Zarr specification 2, which will require separate fix or adapting our Kerchunk writer to use  Zarr format 3. I'd be tempted to go with the second option because converting from Zarr V3 to V2 for Kerchunk adds a lot of surface area for potential bugs, but that would be a separate PR.

rsignell:
@maxrjones cool, I wanted to use icechunk anyway!   And as you said, it *does* solve my use case:
https://nbviewer.org/gist/rsignell/480cf6ac5142ce2b828199bcc601a93e

virtualizarr/manifests/store.py
```diff
@@ -179,12 +179,14 @@ async def get(
             key, marr.metadata.chunk_key_encoding.separator
         )
 
+        if manifest._paths.shape == (1,) and chunk_indexes == ():
+            # TODO: Kerchunk parsers (e.g., NetCDF3) are returning 1 dimensional arrays for scalar arrays, this should be addressed at the source of the issue rather than as a workaround
+            chunk_indexes = (0,)
```
TomNicholas: Isn't this going to cause issues when reading from references that were created correctly? And isn't this bugfix unrelated to the serializer change?

TomNicholas (commented):
So to check my understanding, the reason this failed was because the default array->bytes codec (i.e. the "serializer") in zarr v3 terms was for little endian? Which is fine until you use it to decode big-endian data, when it will give you gibberish?

virtualizarr/manifests/store.py
maxrjones: I opened a separate issue and removed this change - https://github.com/zarr-developers/VirtualiZarr/issues/772. It was necessary for Rich's example to work, which is why it snuck in.

maxrjones:
> So to check my understanding, the reason this failed was because the default array->bytes codec (i.e. the "serializer") in zarr v3 terms was for little endian? Which is fine until you use it to decode big-endian data, when it will give you gibberish?

That's right.

We still likely have bugs because a user can define their own codec that doesn't subclass from Zarr python. In that case, their custom codec will be dropped by https://github.com/zarr-developers/VirtualiZarr/blob/f3149d6464fa2e88c01c71d03170993f93bb3e8c/virtualizarr/codecs.py#L49-L63 if they use create_v3_array_metadata. This may be the cause of https://github.com/zarr-developers/VirtualiZarr/issues/770.

I expect a lot of our codec handling will get much easier with @d-v-b's refactor in Zarr-Python, which is why I have been proposing small patch fixes rather than more fundamental changes.

TomNicholas:
> their custom codec will be dropped by

Could we at least raise in that scenario?

TomNicholas (approved):
Add a changelog entry and this is good to go, thank you!

maxrjones:
> > their custom codec will be dropped by
> 
> Could we at least raise in that scenario?

yeah we definitely should raise in that scenario.

Issue #768, opened 2025-08-11
Serialize to_kerchunk using Zarr format 3


maxrjones:
@mpiannucci pointed out that Kerchunk's reference specification is agnostic to the Zarr format. I think we should use Zarr format 3 rather than Zarr format 2 in the kerchunk writer to reduce the potential for bugs impacting the metadata conversation. This would also require updating the Kerchunk parser to parser Zarr format 3 references.

PR #769, opened 2025-08-11
Extract endianness from Bytes codec in V2 metadata conversion

### LINKED ISSUE #750: NetCDF3Parser workflow returning gibberish (while equivalent HDFParser workflow works fine)

maxrjones:
This PR complements #766 by adding to_kerchunk support (whereas #766 added to_icechunk support). For a more robust long-term solution that would eliminate the need for separate metadata pipelines between kerchunk and icechunk, see issue #768.

@rsignell I tested your workflow works using this PR (see https://github.com/maxrjones/test-scripts/blob/main/test-virtualizarr-netcdf3/test-virtualizarr.py).

- [x] Closes #750
- [x] Tests added
- [x] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -20,6 +20,8 @@
   By [Tom Nicholas](https://github.com/TomNicholas).
 - Fix handling of big-endian data in Icechunk by making sure that non-default zarr serializers are included in the zarr array metadata [#766](https://github.com/zarr-developers/VirtualiZarr/issues/766).
   By [Max Jones](https://github.com/maxrjones)
+- Fix handling of big-endian data in Kerchunk references [#769](https://github.com/zarr-developers/VirtualiZarr/issues/769).
+  By [Max Jones](https://github.com/maxrjones)
 
 ### Documentation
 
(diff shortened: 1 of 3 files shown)


rsignell:
Yep, this works for my use case!
image

PR #774, opened 2025-08-13
Raise if missing virtual chunk containers

### LINKED ISSUE #763: Decoding of virtual chunks from Icechunk fails when creating an array with an ArrayBytesCodec serializer.
While testing the inclusion of the `serializer` during Icechunk array creation as detailed in https://github.com/zarr-developers/VirtualiZarr/issues/762 I'm running into what appears to potentially be an upstream encoding or decoding issue with Icechunk.  Using a modified `write_virtual_variable_to_icechunk` I'm seeing `_FillValue` data returned rather than the values generated by the dummy `ArrayBytesCodec` used for testing.  The following notebook (along with the required change to the icechunk writer) can be used to reproduce
https://gist.github.com/sharkinsspatial/63fac8901bf58b9b9ba18853fd01a9d5

TomNicholas:
Closes #763 using suggestion in https://github.com/zarr-developers/VirtualiZarr/issues/763#issuecomment-3180234781

- [x] Should close #763
- [x] Tests added
- [ ] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_writers/test_icechunk.py
+++ b/virtualizarr/tests/test_writers/test_icechunk.py
@@ -24,6 +24,7 @@ if TYPE_CHECKING:
         IcechunkStore,
         Repository,
         Storage,
+        Diff,
     )
 
 
@@ -373,9 +374,23 @@ def test_validate_containers(
     # assert that no uncommitted changes have been written to Icechunk session
     # Idea is that session has not been "polluted" with half-written changes
     session = icechunk_filestore.session
-    diff = session.status
-    # TODO I don't think this will work without adding __bool__ to icechunk's Diff class
-    assert not diff
+    diff = session.status()
+    assert diff_is_empty(diff), diff
+
+
+def diff_is_empty(diff: "Diff") -> bool:
+    # TODO would be nicer to implement __bool__ on icechunk's Diff class
+    return not any(
+        [
+            bool(diff.deleted_arrays),
+            bool(diff.deleted_groups),
+            bool(diff.new_arrays),
+            bool(diff.new_groups),
+            bool(diff.updated_arrays),
+            bool(diff.updated_chunks),
+            bool(diff.updated_groups),
+        ]
+    )
 
 
 def test_checksum(
--- a/virtualizarr/tests/test_writers/test_icechunk.py
+++ b/virtualizarr/tests/test_writers/test_icechunk.py
@@ -338,6 +338,46 @@ def test_write_loadable_variable(
         npt.assert_equal(pressure_array, expected_array)
 
 
+def test_validate_containers(
+    icechunk_filestore: "IcechunkStore",
+    array_v3_metadata,
+    tmpdir: Path,
+) -> None:
+    # create some references referring to data that doesn't have a corresponding virtual chunk container
+    manifest = ChunkManifest(
+        {"0.0": {"path": "s3://bucket/path/file.nc", "offset": 0, "length": 100}}
+    )
+    metadata = array_v3_metadata(
+        shape=(3, 4),
+        chunks=(3, 4),
+        codecs=None,
+    )
+    ma = ManifestArray(
+        chunkmanifest=manifest,
+        metadata=metadata,
+    )
+    vds = xr.Dataset(
+        {
+            "foo": (["x", "y"], ma),
+            # include some non-virtual data too
+            "bar": (["x", "y"], np.ones((3, 4))),
+        },
+    )
+
+    # assert that an error is raised when attempting to write to icechunk
+    with pytest.raises(
+        ValueError, match="No Virtual Chunk Container set which matches prefix"
+    ):
+        vds.vz.to_icechunk(icechunk_filestore)
+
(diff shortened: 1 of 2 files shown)


virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -338,6 +339,60 @@ def test_write_loadable_variable(
...
+        [
+            bool(diff.deleted_arrays),
+            bool(diff.deleted_groups),
+            bool(diff.new_arrays),
+            bool(diff.new_groups),
+            bool(diff.updated_arrays),
+            bool(diff.updated_chunks),
+            bool(diff.updated_groups),
+        ]
+    )
```
TomNicholas: @paraseba if `Diff` had this as a new method

```python
class Diff:
    @property
    def __bool__(self) -> bool:
        """Returns true if there are any changes whatsoever."""
        return any(
            [
                bool(diff.deleted_arrays),
                bool(diff.deleted_groups),
                bool(diff.new_arrays),
                bool(diff.new_groups),
                bool(diff.updated_arrays),
                bool(diff.updated_chunks),
                bool(diff.updated_groups),
            ]
        )
```

then I could simplify my check to just

```suggestion
    assert not diff, diff
```

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -338,6 +339,60 @@ def test_write_loadable_variable(
...
+    with pytest.raises(
+        ValueError, match="No Virtual Chunk Container set which supports prefix"
+    ):
+        vds.vz.to_icechunk(icechunk_filestore)
+
+    # assert that no uncommitted changes have been written to Icechunk session
+    # Idea is that session has not been "polluted" with half-written changes
+    session = icechunk_filestore.session
+    diff = session.status()
+    assert diff_is_empty(diff), diff
```
dcherian:
```suggestion
    assert not session.has_uncommitted_changes()
```

virtualizarr/tests/test_writers/test_icechunk.py
TomNicholas: Request tracked upstream in https://github.com/earth-mover/icechunk/issues/1165

virtualizarr/writers/icechunk.py
```diff
@@ -172,12 +177,25 @@ def virtual_datatree_to_icechunk(
...
+        return StorePath(store, path="" if at_root else subtree.relative_to(vdt))
+
+    # can't just use a dict because StorePath is not hashable
+    paths_and_virtual_datasets = [
+        (get_store_path(subtree, vdt), node_to_vds(subtree)) for subtree in vdt.subtree
+    ]
+    virtual_datasets = [pair[1] for pair in paths_and_virtual_datasets]
 
-        store_path = StorePath(store, path="" if at_root else tree.relative_to(vdt))
+    validate_virtual_chunk_containers(store.session.config, virtual_datasets)
```
TomNicholas: This line requires an unreleased change in icechunk: https://github.com/earth-mover/icechunk/pull/1164

TomNicholas: Otherwise you can't do `session.config`

virtualizarr/writers/icechunk.py
```diff
@@ -188,6 +217,49 @@ def virtual_datatree_to_icechunk(
         )
 
 
+def validate_virtual_chunk_containers(
```
TomNicholas, line 227: I kinda feel like I should be able to just call some public Icechunk API here to do this instead of doing it myself. Like there is currently no way to validate a virtual reference without then immediately setting it.

TomNicholas: Raised https://github.com/earth-mover/icechunk/issues/1167#issue-3319420773 upstream to track this idea.

TomNicholas:
The upstream tests are now passing, which means that oncee Icechunk is released (see https://github.com/zarr-developers/VirtualiZarr/pull/774#discussion_r2273929806) this can be merged (and then this PR would need to pin to that new IC release).

virtualizarr/writers/icechunk.py
```diff
@@ -188,6 +217,54 @@ def virtual_datatree_to_icechunk(
...
+        for var in dataset.variables.values()
+        if isinstance(var.data, ManifestArray)
+    ]
+
+    # get the prefixes of all virtual chunk containers
+    if config.virtual_chunk_containers is None:
+        # TODO for some reason Icechunk returns None instead of an empty dict if there are zero containers
+        supported_prefixes = set()
+    else:
+        supported_prefixes = set(config.virtual_chunk_containers.keys())
```
TomNicholas: This is due to https://github.com/earth-mover/icechunk/issues/1168

TomNicholas:
> Does this mean we're bumping the minimum supported IC version?

We would need to yes.

TomNicholas:
> since we're using ``True`` as the default.

We were already using ``True`` as the default - the breaking change is that we're actually enforcing it!

But I agree that this effectively constitutes a breaking change - I will add it now.

PR #787, opened 2025-08-26
account for `endian` being possibly `None`

keewis:
For some reason a dataset I've been working on produces `BytesCodec` instances where `endian` is `None`, which causes the translation from v3 to v2 metadata to fail.

No tests because I didn't check thoroughly on where these should be.

(I've seen that you're planning to migrate to v3 entirely, in which case this won't be necessary anymore. It's such a simple fix, though, that this might still be worth it)

- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/utils.py
+++ b/virtualizarr/utils.py
@@ -145,7 +145,7 @@ def convert_v3_to_v2_metadata(
     # but other codec pipelines could store endianness elsewhere.
     big_endian = any(
         isinstance(codec, ArrayBytesCodec)
-        and hasattr(codec, "endian")
+        and getattr(codec, "endian", None) is not None
         and codec.endian.value == "big"
         for codec in v3_metadata.codecs
     )


keewis:
> are you able to share this dataset?

Not sure, I'll have to find a public version of it.

> what dtype is this for?

these are single-byte dtypes (in my case, `int8`). The encoding in the virtual dataset is indeed empty, but it still creates a `BytesCodec` instance with `endian=None`. The repr of `v3_metadata` is:
```
(BytesCodec(endian=None), Shuffle(codec_name='numcodecs.shuffle', codec_config={'elementsize': 1}), Zlib(codec_name='numcodecs.zlib', codec_config={'level': 1}))
```
which I assume is the default for this dtype?

keewis:
in digging through the zarr code, I noticed that
```python
BytesCodec.from_dict({"name": "bytes", "configuration": {}})
BytesCodec.from_dict({"name": "bytes"})
```
both get the default system endian because that's the default for `__init__`, and it is only when explicitly passing `None` that that appears:
```python
BytesCodec.from_dict({"name": "bytes", "configuration": {"endian": None}})
```
(but that would be against the spec, right?)

Also:
```python
from zarr.codecs import BytesCodec

codec = BytesCodec(endian=None)
roundtripped = BytesCodec.from_dict(codec.to_dict())

assert codec.endian == roundtripped.endian
```
fails for the same reason.

@d-v-b, any idea on what the intended behavior is?

keewis:
the failing RTD build looks unrelated (it fails while creating the env)

betolink:
We just ran into this and was about to open the same PR, thanks for working on this @keewis. It'll be great if this gets merged and we get a new release since this is a blocker for all earthaccess virtualizarr workflows. cc @maxrjones

PR #790, opened 2025-08-28
feat: nested group in `ManifestStore` + `HDFParser`

### LINKED ISSUE #664: `hdf` parser limited to level-0 `h5py.Dataset`
I think there is a limitation to the `hdf5` parser where it only looks for "shallow" groups i.e., groups with only one level of dataset.

Some example data:
```bash
wget https://datasets.cellxgene.cziscience.com/4cc9a6b9-ae3a-4084-b3f0-f19578eb30ac.h5ad -O sc_dataset.h5ad
```

And then:
```python
import virtualizarr

store = virtualizarr.parsers.HDFParser()("./sc_dataset.h5ad", object_store=obstore.store.LocalStore('./'))
zarr.open_group(store, mode="r", zarr_format=3) # evidently I need to specify v3 otherwise this crashes without finding an group?
```
outputs
```
<Group ManifestStore(group=
ManifestGroup(
    arrays={},
    groups={},
    metadata=GroupMetadata(attributes={'encoding-type': 'anndata', 'encoding-version': '0.1.0'}, zarr_format=3, consolidated_metadata=None, node_type='group'),
)
, stores=<virtualizarr.manifests.store.ObjectStoreRegistry object at 0x15b4e2ab0>)>
```
which lacks the nested group structure:
```python
import h5py

with h5py.File("./sc_dataset.h5ad") as f:
    print(list(f.keys()))
```
```
['X', 'layers', 'obs', 'obsm', 'obsp', 'raw', 'uns', 'var', 'varm', 'varp']
```

I think the culprit is here

https://github.com/zarr-developers/VirtualiZarr/blob/ce44d1fbd579a0883f10ce35e1bb3247777f2643/virtualizarr/parsers/hdf/hdf.py#L94-L126

This function is called in `HDFParser` but only handles level-0 datasets and not nested groups.

Maybe this is intentional, not sure! Thanks for the library, very cool!

ilan-gold:
Some potential TODOs:

1. More complex `h5` objects from currently parsed file formats that fit the nested paradigm and would otherwise have not worked
2. ~~Along the lines of 1., maybe bring in `anndata` as an optional dep to test the behavior of this more deeply~~ I think this raises separate issues because `anndata` has semantics around object dtype that would need to be handled by a parser, I think 
3. More pathological examples and/or other places where this feature might not work.  I'm not too familiar with this library/space so would be very appreciative of some investigative directions! Here are a few potential ones:

- `open_virtual_dataset` seems to return an `xarray.Dataset` but I think in the case where things are nested, it should return a `DataTree` maybe? or error out? `construct_virtual_dataset` and thus `ManifestStore.to_virtual_dataset` suffer the same issue
- Can the non-hdf5 formats handle nested structures, like `netCDF` or `fits`?

- [x] Closes #664
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -7,6 +7,9 @@
 - Allow nested-groups inside `ManifestStore` and `ManifestGroup` objects and update `HDFParser` to be able to create nested `zarr.Group` objects.
   ([#790](https://github.com/zarr-developers/VirtualiZarr/pull/790)).
   By [Ilan Gold](https://github.com/ilan-gold)
+- `ZarrParser` now handles Zarr V2 and V3 array parsing.
+  ([#565](https://github.com/zarr-developers/VirtualiZarr/pull/822)).
+  By [Neil Schroeder](https://github.com/neilSchroeder)
 
 ### Breaking changes
 
--- a/virtualizarr/parsers/zarr.py
+++ b/virtualizarr/parsers/zarr.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
 import asyncio
+from abc import ABC, abstractmethod
 from collections.abc import Iterable
-from pathlib import Path  # noqa
-from typing import TYPE_CHECKING, Any, Hashable
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
 
 import zarr
 from zarr.api.asynchronous import open_group as open_group_async
+from zarr.core.group import GroupMetadata
 from zarr.core.metadata import ArrayV3Metadata
 from zarr.storage import ObjectStore
 
@@ -16,7 +18,7 @@ from virtualizarr.manifests import (
     ManifestGroup,
     ManifestStore,
 )
-from virtualizarr.manifests.manifest import validate_and_normalize_path_to_uri  # noqa
+from virtualizarr.manifests.manifest import validate_and_normalize_path_to_uri
 from virtualizarr.registry import ObjectStoreRegistry
 from virtualizarr.vendor.zarr.core.common import _concurrent_map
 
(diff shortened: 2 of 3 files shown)


TomNicholas:
> `open_virtual_dataset` seems to return an `xarray.Dataset` but I think in the case where things are nested, it should return a `DataTree` maybe? or error out? `construct_virtual_dataset` and thus `ManifestStore.to_virtual_dataset` suffer the same issue

Sorry I think I should have linked to #84. What we want is to add `open_virtual_datatree` and `ManifestStore.to_virtual_datatree`, and have `.to_virtual_dataset/tree` implemented on a `ManifestStore` class which supports nested groups (that's what this PR adds, which is great).

> Can the non-hdf5 formats handle nested structures, like netCDF or fits?

NetCDF4 is HDF5 in a trenchcoat 🙂 NetCDF3 cannot handle groups. TIFF can. FITS IDK.

virtualizarr/manifests/group.py
```diff
@@ -40,13 +40,7 @@ def __init__(
...
 
         _arrays: Mapping[str, ManifestArray] = {} if arrays is None else arrays
-
-        if groups:
-            # TODO add support for nested groups
-            raise NotImplementedError
-        else:
-            _groups: Mapping[str, ManifestGroup] = {} if groups is None else groups
-
+        _groups: Mapping[str, ManifestGroup] = {} if groups is None else groups
```
TomNicholas, line 43: I'm not sure if #796 should be dealt with first...

ilan-gold:
Yes, I just assumed the PR was "approved" because it was put in v3.0.0, milestone, but I'm ready to respond to any requested changes :)

virtualizarr/manifests/store.py
```diff
@@ -97,6 +98,16 @@ def parse_manifest_index(
     return tuple(int(ind) for ind in chunk_component.split(chunk_key_encoding))
 
 
+def get_deepest_group_or_array(
+    node: ManifestGroup, key: list[str]
```
chuckwondo: I suggest changing the `key` type to either `Sequence` or `Iterable` instead of restricting to `list`.

virtualizarr/tests/test_parsers/test_hdf/test_hdf.py
```diff
@@ -129,12 +130,29 @@ def test_variable_with_dimensions(self, chunked_dimensions_netcdf4_url):
...
 
+    def test_nested_groups_are_detected(self, nested_group_hdf5_url):
+        manifest_store = manifest_store_from_hdf_url(nested_group_hdf5_url)
+        assert len(manifest_store._group["group"]["nested_group"].arrays) == 1
+
+    def test_nested_data(self, nested_group_hdf5_url):
+        manifest_store = manifest_store_from_hdf_url(nested_group_hdf5_url)
+        z = zarr.open_group(manifest_store, mode="r", zarr_format=3)
+
+        with h5py.File(nested_group_hdf5_url.split("file://")[-1], mode="r") as f:
```
chuckwondo: I'm getting nit-picky, but this is perhaps more readable:

```suggestion
        with h5py.File(nested_group_hdf5_url.removeprefix("file://"), mode="r") as f:
```

virtualizarr/manifests/store.py
ilan-gold: Yup, great call, better permissive on inputs

virtualizarr/xarray.py
```diff
@@ -333,6 +333,7 @@ def construct_virtual_dataset(
 
     """
 
+    # TODO: Remove private API `._group`
```
TomNicholas, line 336: Are you planning to do this in a follow-up?

PR #791, opened 2025-08-28
Do not validate references for "empty" chunks.

sharkinsspatial:
- [x] Tests passing

Currently, serializing a `ManifestArray` containing empty chunks to Icechunk requires explicitly passing `validate_containers=False` to `to_icechunk` which may be counterintuitive for some users.  ref https://github.com/zarr-developers/VirtualiZarr/issues/740.  This PR changes the current behavior to only perform reference validation for references with a `path`.

### DIFF
--- a/virtualizarr/tests/test_writers/test_icechunk.py
+++ b/virtualizarr/tests/test_writers/test_icechunk.py
@@ -731,7 +731,7 @@ def test_write_empty_chunk(
     vds = xr.Dataset({"a": ("x", marr)})
 
     # empty chunks should never be written
-    vds.vz.to_icechunk(icechunk_filestore, validate_containers=False)
+    vds.vz.to_icechunk(icechunk_filestore)
 
     # when opened they should be treated as fill_value
     roundtrip = xr.open_zarr(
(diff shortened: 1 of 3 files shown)


sharkinsspatial:
@TomNicholas Changelog updated 👍.

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -731,7 +731,7 @@ def test_write_empty_chunk(
     vds = xr.Dataset({"a": ("x", marr)})
 
     # empty chunks should never be written
-    vds.vz.to_icechunk(icechunk_filestore, validate_containers=False)
+    vds.vz.to_icechunk(icechunk_filestore)
```
maxrjones, line 734: Don't you still want validate_containers=False here?

TomNicholas: No. We are testing that even when explicitly validating the references, empty references are still allowed through.

TomNicholas: And the default is `True`, so we don't want to override that.

PR #795, opened 2025-09-05
Fix examples' links

malmans2:
Looks like #644 broke the examples' links in the documentation. This PR fixes them.

Maybe we should move the examples into the docs folder so the documentation can use local paths?


malmans2:
Ah, I just realised some examples aren't just notebooks or stand-alone files, which is probably why local paths are not used

Issue #804, opened 2025-09-24
TypeError: Cannot interpret 'Float32(endianness='little')' as a data type


rsignell:
My virtualizarr append-to-icechunk notebook was working with `virtualizarr=2.1.1` but is broken with `virtualizarr=2.1.2`:
https://gist.github.com/rsignell/553ac0f3025413288d314897c8fa4ca1

In cell [23] it errors out with `TypeError: Cannot interpret 'Float32(endianness='little')' as a data type `

Is this something I need to change or is it a bug?

rsignell:
@keewis, might this be related to your endianness change in 2.1.2?

rsignell:
```
(notebook) jovyan@jupyter-rsignell4:~$ conda list zarr
# packages in environment at /srv/conda/envs/notebook:
#
# Name                    Version                   Build  Channel
virtualizarr              2.1.2              pyhd8ed1ab_0    conda-forge
zarr                      3.1.3              pyhcf101f3_0    conda-forge
```

PR #760, opened 2025-08-08
Remove unnecessary dtype conversion in icechunk writer

popododo0720:
- [x] Closes #758
- [ ] Tests added (not needed for this simple change)
- [ ] Tests passing
- [ ] Full type hint coverage (no type changes)
- [ ] Changes are documented in `docs/releases.rst` (minor internal change)
- [ ] New functions/methods are listed in `api.rst` (no new functions)
- [ ] New functionality has documentation (no new functionality)

## Description
Removed unnecessary `.to_native_dtype()` conversion in icechunk writer as suggested by @d-v-b in #758. 
Now passing `metadata.data_type` directly.

## Changes
- Simplified dtype parameter in `VirtualiZarr/virtualizarr/writers/icechunk.py` line 347

This is my first contribution!

PR #805, opened 2025-09-24
Revert "Remove unnecessary dtype conversion in icechunk writer"

TomNicholas:
Reverts zarr-developers/VirtualiZarr#760, as per #804


maxrjones (approved):
I think we should merge this to fix https://github.com/zarr-developers/VirtualiZarr/issues/804 until we understand better why the Zarr dtypes aren't working as expected.

TomNicholas:
👍 go for it

PR #806, opened 2025-09-24
Parse zarr v2

### LINKED ISSUE #565: Virtualize Native Zarr V2 format

TomNicholas:
@norlandrhagen this is in theory all that would be needed. But when I actually run it I hit some bugs. e.g.

```bash
TypeError: Cannot convert object None with type <class 'NoneType'> to a scalar compatible with the data type Int16(endianness='little').
```

- [ ] Closes #565
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_parsers/test_zarr.py
+++ b/virtualizarr/tests/test_parsers/test_zarr.py
@@ -12,11 +12,7 @@ from virtualizarr.registry import ObjectStoreRegistry
 @pytest.mark.parametrize(
     "zarr_store",
     [
-        pytest.param(
-            2,
-            id="Zarr V2",
-            marks=pytest.mark.skip(reason="Zarr V2 not currently supported."),
-        ),
+        pytest.param(2, id="Zarr V2"),
         pytest.param(3, id="Zarr V3"),
     ],
     indirect=True,


virtualizarr/parsers/zarr.py
```diff
@@ -84,9 +84,10 @@ def get_metadata(zarr_array: zarr.AsyncArray[Any]) -> ArrayV3Metadata:
     zarr_format = zarr_array.metadata.zarr_format
 
     if zarr_format == 2:
-        # TODO: Once we want to support V2, we will have to deconstruct the
-        # zarr_array codecs etc. and reconstruct them with create_v3_array_metadata
-        raise NotImplementedError("Reading Zarr V2 currently not supported.")
+        # only import this if needed because it's private zarr internals which are likely unstable
+        from zarr.metadata.migrate_v3 import _convert_array_metadata
+
+        return _convert_array_metadata(zarr_array.metadata)
```
TomNicholas, line 90: fyi @d-v-b I think this is the piece of API from https://github.com/zarr-developers/zarr-python/pull/3257 that we would ideally like to be exposed as public and stable in zarr-python.

d-v-b:
> TypeError: Cannot convert object None with type <class 'NoneType'> to a scalar compatible with the data type Int16(endianness='little').

That's a symptom of converting a V2 array which had `fill_value: null` to v3, where the fill value can't be null. You can handle this case by using the default fill value, which is associated with the data type via `default_scalar` method

virtualizarr/parsers/zarr.py
d-v-b: that function will probably need some time to stabilize but in general :100: to making it public

neilSchroeder:
Hey Tom, I'm interested in adding some changes, but I'm a little confused about how to go about it in this scenario. Should I open a new PR with the changes in my own fork or contribute to your fork?

Issue #811, opened 2025-10-09
ZarrParser get_metadata fails for LocalStore Zarr with dtype uint8


eternity8:
I am new to VirtualiZarr, tried loading a Zarr file from my local machine as a virtual zarr with the ZarrParser:

`
data = open_virtual_dataset(
    url=file_url, parser=ZarrParser("my_file.zarr"), registry=registry
)
`

However, the open operation fails due to one of the columns in the zarr having type uint8.

The failure occurs in the get_metadata function due to a key error on the default fill value lookup:
https://github.com/zarr-developers/VirtualiZarr/blob/a2b65c1f48e29c0e51751a29e0b32f09be660bad/virtualizarr/parsers/zarr.py#L82

The default fill values are hardcoded as a dictionary:

https://github.com/zarr-developers/VirtualiZarr/blob/a2b65c1f48e29c0e51751a29e0b32f09be660bad/virtualizarr/parsers/zarr.py#L25-L33
but does not include the 'u' key, (np.dtype("uint8").kind ) 

Is this a simple fix of adding more keys to the hardcoded default fill value list? Or are unsigned ints unsupported at this stage?

PR #812, opened 2025-10-09
Remove `ZARR_DEFAULT_FILL_VALUE` lookup

norlandrhagen:
Towards https://github.com/zarr-developers/VirtualiZarr/issues/811.

In the zarr-parser I used the `ZARR_DEFAULT_FILL_VALUE` lookup dict, but the `fill_value` was never used / reassigned in the `ArrayV3Metadata`. I think it's only purpose was causing bugs like #811. 

- [x] Closes #xxxx
- [x] Tests added
- [x] Tests passing


norlandrhagen:
Added! I think the mypy and rtd errors are unrelated.

TomNicholas (approved):
Thanks!

PR #822, opened 2025-10-31
Parse zarr v2

### LINKED ISSUE #565: Virtualize Native Zarr V2 format

neilSchroeder:
# Checklist

- [x] Closes #565 
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

# How `zarr.py` Handles Zarr V2 Stores
`ZarrParser` should now support both Zarr V2 and V3 stores by normalizing V2 stores to appear as V3. This approach ensures that all parsers produce V3-compatible outputs, and confines modifications to `zarr.py`.

## V2 → V3 Normalization Strategy
The parser performs a two-part normalization:

### 1. Chunk Key Mapping (`get_chunk_mapping_prefix`)
For V2 arrays:

- Chunk files are stored directly under the array path: `array_name/0`, `array_name/0.1.2`
- Metadata files (.zarray, .zattrs, etc.) are filtered out
- Chunk coordinates are normalized to dot-separated format: "0.1.2"
- File paths in the manifest point to the actual V2 chunk locations
- Manifest keys contain only chunk coordinates (no path structure)

### 2. Metadata Conversion (`get_metadata()`)
After converting V2 metadata to V3 using `_convert_array_metadata`, we have to replace the `chunk_key_encoding`. 

- The automatic converter preserves `V2ChunkKeyEncoding` in the V3 metadata
- When `zarr`/`xarray` sees `V2ChunkKeyEncoding`, it requests chunks using V2-style paths: `array/0`
- With `DefaultChunkKeyEncoding`, `zarr` requests chunks using V3-style paths: `array/c/0`
- `ManifestStore.get()` expects V3-style paths and uses `parse_manifest_index()` to extract chunk coordinates
- `parse_manifest_index()` requires the `/c/` component to correctly parse the path

## Additional metadata handling

- None fill values: Converted to appropriate dtype defaults
- Dimension names: Extracted from `_ARRAY_DIMENSIONS` attribute or generated as `{array_name}_dim_{i}`
- All other metadata: Converted using `zarr`'s standard V2→V3 migration utilities

## Implementation Notes

I'm not convinced I've done a particularly elegant implementation here, but adding another class for V2 parsing didn't seem like it would be particularly extensible. Very happy to hear thoughts on perhaps a better implementation.

@TomNicholas thank you very much for your feedback, it definitely helped me wrap my head around the right approach to take here.

Edit: I've done a bit of re-design to use a strategy pattern for dispatching to parsing v2 and v3 arrays. This should make future integrations of zarr array version parsing a lot more maintainable. This is also just a lot easier to read than my original implementation. Tests and documentation are also up to date.

### DIFF
--- a/virtualizarr/parsers/zarr.py
+++ b/virtualizarr/parsers/zarr.py
@@ -226,7 +226,11 @@ class ZarrV2Strategy(ZarrVersionStrategy):
         # Replace V2ChunkKeyEncoding with V3 DefaultChunkKeyEncoding
         # The automatic conversion preserves V2's encoding, causing zarr to use V2-style
         # paths (array/0) instead of V3-style (array/c/0). This ensures V3 semantics.
-        if "attributes" in v3_dict and "_ARRAY_DIMENSIONS" in v3_dict["attributes"]:
+        if (
+            "attributes" in v3_dict
+            and isinstance(v3_dict["attributes"], dict)
+            and "_ARRAY_DIMENSIONS" in v3_dict["attributes"]
+        ):
             del v3_dict["attributes"]["_ARRAY_DIMENSIONS"]
             v3_metadata = ArrayV3Metadata.from_dict(v3_dict)
         v3_dict["chunk_key_encoding"] = {"name": "default", "separator": "."}
@@ -245,9 +249,14 @@ class ZarrV3Strategy(ZarrVersionStrategy):
         # Check for sharding - not yet supported
         from zarr.codecs import ShardingCodec
 
-        if any(
-            isinstance(codec, ShardingCodec) for codec in zarr_array.metadata.codecs
-        ):
+        # Type narrowing: V3 strategy only handles V3 arrays with V3 metadata
+        metadata = zarr_array.metadata
+        if not isinstance(metadata, ArrayV3Metadata):
+            raise TypeError(
+                f"Expected ArrayV3Metadata in V3 strategy, got {type(metadata)}"
+            )
+
+        if any(isinstance(codec, ShardingCodec) for codec in metadata.codecs):
             raise NotImplementedError(
                 "Zarr V3 arrays with sharding are not yet supported. "
                 "Sharding stores multiple chunks in a single storage object with non-zero offsets, "


virtualizarr/manifests/store.py
TomNicholas, line 1: Hey @neilSchroeder ! This is awesome that you're working on this!

However, I don't think you should need to make any changes to this file in order to add this feature.

IIUC you're altering the key-parsing logic inside `ManifestStore` to accommodate V2-like keys. But that's not the right place for that logic. The `ManifestStore` is explicitly a V3 store - it's key logic should not need to be changed. Instead, you need to alter the `ZarrParser` (or add a separate `ZarrV2Parser`) to do that mapping. The output of the Parser should not have any trace of what file format was parsed, zarr or otherwise. So you might need some v2-key-handling logic like this inside your parser, but not in `manifests/store.py`. Does that make sense? Am I understanding properly?

Changing this logic in the `ManifestStore` is also what's causing other, unrelated, tests to fail.

neilSchroeder: Yeah this totally makes sense. Thanks for the feedback! I wasn't exactly sure where some of these fixes needed to go so I took a pretty naive test-based-development approach. 

I'll think about this a little more carefully and see if I can't come up with a more elegant solution that avoids manipulating the stores.

TomNicholas: Great - shout if you have any questions / aren't sure about anything!

neilSchroeder:
@TomNicholas I think it's ready for a review.

virtualizarr/parsers/zarr.py
```diff
@@ -26,64 +26,272 @@
...
+            chunk_key = "0"
+            object_key = f"{prefix}{chunk_key}"
+            size = await zarr_array.store.getsize(object_key)
+            actual_path = join_url(path, object_key)
+            return {"0": {"path": actual_path, "offset": 0, "length": size}}
+
+        # List all keys under the array prefix, filtering out metadata files
+        prefix_keys = [(x,) async for x in zarr_array.store.list_prefix(prefix)]
+        if not prefix_keys:
+            return {}
```
TomNicholas, line 164: There's an important subtlety here that's worth leaving a comment about: The reason we don't just generate the names of the keys from the chunk grid is because zarr chunks are allowed to be missing, and they are allowed to be missing in our VZ manifests too. It might actually be worth adding a test for this case - the case that there are some chunks but some are missing - VZ should return the `fill_value` for any missing chunks.

virtualizarr/parsers/zarr.py
```diff
@@ -26,64 +26,272 @@
...
+
+        # Normalize to dot-separated coordinates (V2 can use either '/' or '.')
+        _dict_keys = [coord.replace("/", ".") for coord in chunk_coords]
+        _paths = [join_url(path, k) for k in chunk_keys]
+        _offsets = [0] * len(_lengths)
+
+        return {
+            key: {"path": p, "offset": offset, "length": length}
+            for key, p, offset, length in zip(_dict_keys, _paths, _offsets, _lengths)
+        }
```
TomNicholas: No need to create a potentially millions-long list of zeros.

```suggestion
        return {
            key: {"path": p, "offset": 0, "length": length}
            for key, p, offset, length in zip(_dict_keys, _paths, _offsets, _lengths)
        }
```

virtualizarr/parsers/zarr.py
```diff
@@ -26,64 +26,272 @@
...
 
-async def build_chunk_manifest(zarr_array: ZarrArrayType, path: str) -> ChunkManifest:
-    """Build a ChunkManifest from a dictionary"""
-    chunk_map = await get_chunk_mapping_prefix(zarr_array=zarr_array, path=path)
-    return ChunkManifest(chunk_map)
+    def get_metadata(self, zarr_array: ZarrArrayType) -> ArrayV3Metadata:
+        """Return V3 metadata as-is (no conversion needed)."""
+        return zarr_array.metadata  # type: ignore[return-value]
```
TomNicholas, line 285: An important thing that we should really already be doing is raising if the zarr V3 array contains [sharding](https://zarr.readthedocs.io/en/stable/user-guide/performance.html#sharding), because we don't yet support that (and if we don't raise then our `offset=0` means we will silently return corrupted results 😬 ).

virtualizarr/parsers/zarr.py
```diff
@@ -26,64 +26,272 @@
...
+
+    async def get_chunk_mapping(
+        self, zarr_array: ZarrArrayType, path: str
+    ) -> dict[str, dict[str, Any]]:
+        """Get mapping of chunk coordinates to storage locations."""
+        ...
+
+    def get_metadata(self, zarr_array: ZarrArrayType) -> ArrayV3Metadata:
+        """Get V3 metadata for the array (converting if necessary)."""
+        ...
```
TomNicholas: Using a protocol here is overkill given that the `ZarrVersionStrategy` is not public API. I appreciate the effort to consolidate code though, so perhaps this could just be an ABC?

virtualizarr/parsers/zarr.py
```diff
@@ -26,64 +26,272 @@
...
+        V3 uses paths like: array_name/c/0/1/2
+        """
+        name = getattr(zarr_array, "name", "") or ""
+        name = name.lstrip("/")
+
+        # Handle scalar arrays
+        if zarr_array.shape == ():
+            prefix = f"{name}/c" if name else "c"
+            size = await zarr_array.store.getsize(prefix)
+            return {"c": {"path": join_url(path, prefix), "offset": 0, "length": size}}
```
TomNicholas: There's quite a lot of logic that is basically the same between the two strategies, for example this scalar handling. It would be nice if it was factored out into little common utility functions (I bet Claude could do this easily).

virtualizarr/parsers/zarr.py
```diff
@@ -113,40 +315,97 @@ async def _construct_manifest_group(
...
 
     manifest_dict = {
         array.basename: result for array, result in zip(zarr_arrays, manifest_arrays)
     }
 
-    return ManifestGroup(manifest_dict, attributes=zarr_group.attrs)
+    manifest_group = ManifestGroup(manifest_dict, attributes=zarr_group.attrs)
+
+    # Set V3 group metadata to ensure zarr/xarray uses V3 semantics
+    from zarr.core.group import GroupMetadata
```
TomNicholas: Move import to top of file

virtualizarr/parsers/zarr.py
```diff
@@ -26,64 +26,272 @@
...
-        raise NotImplementedError("Zarr format is not recognized as v2 or v3.")
+        raise NotImplementedError(f"Zarr format {zarr_format} is not supported")
+
 
+async def build_chunk_manifest(zarr_array: ZarrArrayType, path: str) -> ChunkManifest:
+    """Build a ChunkManifest from chunk coordinate mappings."""
+    strategy = get_strategy(zarr_array)
+    chunk_map = await strategy.get_chunk_mapping(zarr_array, path)
+
+    if not chunk_map:
```
TomNicholas, line 327: Why would there ever not be a `chunk_map` here?

virtualizarr/tests/test_parsers/test_zarr.py
```diff
@@ -6,7 +6,8 @@
 from virtualizarr import open_virtual_dataset
 from virtualizarr.manifests import ManifestArray
 from virtualizarr.parsers import ZarrParser
-from virtualizarr.parsers.zarr import get_chunk_mapping_prefix, get_metadata
+
+# from virtualizarr.parsers.zarr import get_chunk_mapping_prefix, get_metadata
```
TomNicholas:
```suggestion
```

virtualizarr/tests/test_parsers/test_zarr.py
```diff
@@ -18,7 +19,7 @@
         pytest.param(
             2,
             id="Zarr V2",
-            marks=pytest.mark.skip(reason="Zarr V2 not currently supported."),
+            # marks=pytest.mark.skip(reason="Zarr V2 not currently supported."),
```
TomNicholas:
```suggestion
```

virtualizarr/parsers/zarr.py
```diff
@@ -113,40 +315,97 @@ async def _construct_manifest_group(
...
+    consistent handling across versions. This includes:
+
+    - Converting the metadata structure from V2 to V3 specification
+    - Handling `None` fill values with appropriate defaults
+    - Setting dimension names from `_ARRAY_DIMENSIONS` attributes or generating defaults
+    - Replacing V2ChunkKeyEncoding with V3's DefaultChunkKeyEncoding
+
+    The parser preserves array attributes and handles empty arrays (no chunks)
+    gracefully. Metadata files (.zarray, .zattrs, .zgroup, .zmetadata) are
+    automatically filtered out when listing chunks.
```
TomNicholas: These are implementation details that don't need to be in the public docs.
```suggestion
```

virtualizarr/parsers/zarr.py
```diff
@@ -156,24 +415,67 @@ def __call__(
...
+
+           - Converts V2 metadata to V3 if necessary
+           - Builds chunk manifests with file paths and byte ranges
+           - Preserves attributes and dimension names
+
+        6. Constructs a ManifestGroup with V3 metadata
+        7. Returns a ManifestStore wrapping the virtual representation
+
+        No data is copied during this process; the ManifestStore contains
+        only lightweight references to the original chunks.
```
TomNicholas: implementation details
```suggestion
```

virtualizarr/tests/test_parsers/test_zarr.py
```diff
@@ -97,33 +96,196 @@ def test_virtual_dataset_zarr_attrs(self, zarr_store):
...
+                # Check dimension names - handling V2 vs V3 difference
+                zarr_format = zg[array].metadata.zarr_format
+                if zarr_format == 2:
+                    # V2 stores dimensions in attributes
+                    expected_dims = expected.get("attributes", {}).get(
+                        "_ARRAY_DIMENSIONS", None
+                    )
+                    if expected_dims:
+                        assert expected_dims == list(vds[array].dims)
+                    # If no _ARRAY_DIMENSIONS, VirtualiZarr generates dimension names
```
TomNicholas: I don't think this is true. Xarray generates default dimension names upon `Variable` construction, but if you used the parser to create a `ManifestStore` pointing at a Zarr store that had no dimension names, I don't think the arrays of your ManifestStore would contain any dimension names.

virtualizarr/tests/test_parsers/test_zarr.py
```diff
@@ -97,33 +96,196 @@ def test_virtual_dataset_zarr_attrs(self, zarr_store):
...
+    """Test that scalar arrays produce correct chunk mappings for both V2 and V3."""
     import asyncio
 
-    chunk_map = asyncio.run(
-        get_chunk_mapping_prefix(
-            zarr_array=zarr_store_scalar, path=str(zarr_store_scalar.store_path)
-        )
+    from zarr.api.asynchronous import open_array
+
+    from virtualizarr.parsers.zarr import get_strategy
```
TomNicholas: move this to the top of the file as zarr is a required dependency anyway

virtualizarr/tests/test_parsers/test_zarr.py
TomNicholas, line 1: There is a an alternative way to test the parser which would be worth adding:

```python
zarr_store = ...
manifeststore = ZarrParser()(zarr_url)

expected = xr.open_zarr(zarr_store)
actual = xr.open_zarr(manifeststore)
xrt.assert_identical(actual, expected)
```

This will open and load the actual data, checking everything is the same. Which is ultimately what we care about.

virtualizarr/parsers/zarr.py
TomNicholas: If all the common logic was factored out, this V3Strategy class should be able to become extremely small, as it barely has to do anything.

virtualizarr/parsers/zarr.py
```diff
@@ -26,64 +26,272 @@
...
+        """
+        from zarr.core.metadata import ArrayV2Metadata
+        from zarr.metadata.migrate_v3 import _convert_array_metadata
+
+        v2_metadata = zarr_array.metadata
+        assert isinstance(v2_metadata, ArrayV2Metadata)
+
+        try:
+            v3_metadata = _convert_array_metadata(v2_metadata)
+        except TypeError as e:
```
TomNicholas: This exception catching seems far too general - can we instead explicitly look for the known problematic case before trying to convert the metadata? Or even better fix it upstream...

virtualizarr/parsers/zarr.py
neilSchroeder: I think I've added a test that covers what you're getting at, but it will need some review to make sure I've understood correctly. As for "worth leaving a comment": did you mean in your review here, or do you think it's important to actually add a comment in the file to raise awareness?

TomNicholas: > I think I've added a test that covers what you're getting at,

I see there are tests for the case that there are no chunks at all in the array, but I don't see a test for the case that there are some chunks but some are missing. Maybe that distinction is not important though.

> As for "worth leaving a comment": did you mean in your review here, or do you think it's important to actually add a comment in the file to raise awareness?

I meant add a comment to the file.

neilSchroeder: I just added the test and haven't pushed it yet. Sorry.

And I'll add a comment to the file.

virtualizarr/parsers/zarr.py
neilSchroeder: I think there are two global situations where this could happen:
1) An array whose metadata has been generated, but no data has been written
2) An array with shape (0,) or similar

For V2 arrays `get_chunk_mapping()` will return `{}` if `prefix_keys` is empty, and when `chunk_keys` is empty after removing metadata files. For V3 this could happen if there are no files found with a `c/` prefix. 

I'm a little unclear about order of operations and whether or not these scenarios are realistic. Do you think this should be removed? Or maybe handled differently?

TomNicholas: Actually ignore me, I think what you're done here is good. 

> I'm a little unclear about order of operations and whether or not these scenarios are realistic.

These scenarios are definitely plausible. 

> Or maybe handled differently?

There might be a way to refactor to have a few fewer levels of functions, but this is good.

neilSchroeder:
@TomNicholas what are the next steps here? Will this be merged whenever someone has time to do the next release? Do we need another reviewer?

PR #829, opened 2025-11-11
Raise informative error on Zarr V2 parsing with Zarr-Python<3.1.3

maxrjones:
This PR builds on #822 to raise a more informative error if using the parser with older versions of Zarr-Python, which do not contain the necessary V2->V3 metadata conversion functionality.

To be merged after #828 to keep changes isolated

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_parsers/test_zarr.py
+++ b/virtualizarr/tests/test_parsers/test_zarr.py
@@ -14,22 +14,36 @@ from virtualizarr.registry import ObjectStoreRegistry
 
 ZarrArrayType = zarr.AsyncArray | zarr.Array
 
-
-@pytest.mark.parametrize(
-    "zarr_store",
-    [
-        pytest.param(
-            2,
-            id="Zarr V2",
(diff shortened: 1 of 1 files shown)


virtualizarr/tests/test_parsers/test_zarr.py
```diff
@@ -106,7 +111,19 @@ def test_virtual_dataset_zarr_attrs(self, zarr_store):
...
+        pytest.param(
+            2,
+            marks=pytest.mark.skipif(
+                version.parse(zarr.__version__) < version.parse("3.1.3"),
+                reason="Zarr format 2 parsing requires zarr>=3.1.3",
+            ),
+        ),
+        3,
+    ],
+)
```
TomNicholas: Maybe factor this out as a variable, so that when we inevitable have to remove this, we only have to remove it in one place?

PR #830, opened 2025-11-12
Bump workflow actions versions

maxrjones:
Chore, motivated by a GH security advisory on one of the actions


chuckwondo (approved):
LGTM!

@maxrjones, to save us such manual effort in the future, I recommend adding a `.github/dependabot.yml` file to do this for us automatically on a periodic basis. For reference, see <https://github.com/MAAP-Project/gedi-subsetter/blob/main/.github/dependabot.yml>.

PR #831, opened 2025-11-12
chore: add support for Python 3.14

maxrjones:
This PR updates the tests to run against the minimum (3.11) and maximum (3.14) supported Python versions and updates the executor arguments for Python 3.14's new buffersize kwarg.


TomNicholas (approved):
Why does the pixi build fail?

Issue #834, opened 2025-11-12
Bug: scalar arrays fail to open with `ValueError: Invalid format for chunk key`


neilSchroeder:
### Description

When attempting to open a Zarr store (V2 or V3) containing scalar arrays (arrays with shape `()`), VirtualiZarr fails with:

```
ValueError: Invalid format for chunk key: 'c'
```

This occurs because the chunk key validation in `ChunkManifest.__init__` expects coordinate-based keys like `"0"` or `"0.1.2"`, but the code is sometimes passing through the storage key `"c"` directly instead of normalizing it to a coordinate.

### Environment

- VirtualiZarr version: 2.2.0
- Zarr version: 3.x
- Python version: 3.13

### Steps to Reproduce

```python
from obstore.store import from_url
from virtualizarr import open_virtual_dataset
from virtualizarr.parsers import ZarrParser
from virtualizarr.registry import ObjectStoreRegistry

ZARR_BUCKET = "s3://cadcat"
ZARR_SOURCE = "s3://cadcat/wrf/ucla/cesm2/historical/1hr/lwdnbc/d01"

store = from_url(ZARR_SOURCE, region="us-west-2", skip_signature=True)
registry = ObjectStoreRegistry({ZARR_SOURCE: store})

parser = ZarrParser()
vds = open_virtual_dataset(
    url=ZARR_SOURCE,
    parser=parser,
    registry=registry,
) # Fails with: ValueError: Invalid format for chunk key: 'c'
```

The same issue occurs with Zarr V3 stores.

### Root Cause

In `_handle_scalar_array()` (line 99), there's a problematic ternary that attempts to distinguish between V2 and V3 scalar chunk keys:

```python
return {
    "0" if scalar_key == "0" else "c": {  # Line 99
        "path": actual_path,
        "offset": 0,
        "length": size,
    }
}
```

**The problem:** This checks if `scalar_key` equals exactly `"0"`, but the actual key passed in includes the array name prefix:
- **V2**: `scalar_key = "temperature/0"` → condition is `False` → returns `"c"` 
- **V3**: `scalar_key = "temperature/c"` → condition is `False` → returns `"c"` 

The manifest key `"c"` then fails validation because `validate_chunk_keys()` expects numeric coordinate notation.

### Proposed Solution

[...]

PR #836, opened 2025-11-12
Allow storing scalar arrays under 'c' key

maxrjones:
This PR strengthens the key parsing in `ChunkManifest`, reusing/generalizing logic about the Zarr hierarchy from `ManifestStore`.

@neilSchroeder, I hope this will address the issues you were finding, which I think was a bug in the ChunkManifest logic rather than the Zarr parser. Sorry for the redundancy with your PR in https://github.com/zarr-developers/VirtualiZarr/pull/835, I just didn't think you should need to workaround this.
Fixes https://github.com/zarr-developers/VirtualiZarr/issues/834

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -1,5 +1,20 @@
 # Release notes
 
+## v2.2.1 (TBD)
+
+### New Features
+
+### Breaking changes
+
+### Bug fixes
+
+- Allow storing scalar arrays under 'c' key. ([#836](https://github.com/zarr-developers/VirtualiZarr/pull/836)).
+  By [Max Jones](https://github.com/maxrjones)
+
+### Documentation
+
+### Internal changes
+
 ## v2.2.0 (12th November 2025)
 
 ### New Features


virtualizarr/manifests/manifest.py
```diff
@@ -507,7 +503,10 @@ def validate_chunk_keys(chunk_keys: Iterable[ChunkKey]):
 
 def get_chunk_grid_shape(chunk_keys: Iterable[ChunkKey]) -> tuple[int, ...]:
     # find max chunk index along each dimension
-    zipped_indices = zip(*[split(key) for key in chunk_keys])
+    if list(chunk_keys) == ["c"]:
+        # Scalar array, cannot be split
+        return ()
+    zipped_indices = zip(*[parse_manifest_index(key) for key in chunk_keys])
```
chuckwondo: In case `chunk_keys` cannot be iterated over more than once:

```suggestion
    chunk_keys = tuple(chunk_keys)
    
    if chunk_keys == ("c",):
        # Scalar array, cannot be split
        return ()
    zipped_indices = zip(*map(parse_manifest_index, chunk_keys))
```

I'd also suggest adding at least 1 unit test for this case.

virtualizarr/manifests/manifest.py
```diff
@@ -1,11 +1,11 @@
 import re
 from collections.abc import ItemsView, Iterable, Iterator, KeysView, ValuesView
 from pathlib import PosixPath
-from typing import Any, Callable, NewType, Tuple, TypedDict, cast
+from typing import Any, Callable, NewType, TypedDict, cast
```
chuckwondo: Import `Callable` from `collections.abc`. It was added there in py3.9 and importing it from `typing` was deprecated.

maxrjones: thanks, done in [`3c2cba4` (#836)](https://github.com/zarr-developers/VirtualiZarr/pull/836/commits/3c2cba4e87fa8812982ed5c7f6b4dbd35b3997af)

TomNicholas (approved):

PR #837, opened 2025-11-13
Improve ManifestStore.list_dir for arrays and nested groups

maxrjones:
The listing functionality in ManifestStore previously assumed a maximum hierarchy depth of 2. Following the addition of [nested group support](https://github.com/zarr-developers/VirtualiZarr/pull/790), we need to expand the functionality.

I would greatly appreciate feedback if there's a way to make this logic less gnarly.
 
- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -10,6 +10,8 @@
 
 - Allow storing scalar arrays under 'c' key. ([#836](https://github.com/zarr-developers/VirtualiZarr/pull/836)).
   By [Max Jones](https://github.com/maxrjones)
+- Improve ManifestStore.list_dir for arrays and nested groups ([#837](https://github.com/zarr-developers/VirtualiZarr/pull/837))
+  By [Max Jones](https://github.com/maxrjones)
 
 ### Documentation
 


virtualizarr/manifests/store.py
```diff
@@ -98,14 +98,41 @@ def parse_manifest_index(
     return tuple(int(ind) for ind in chunk_component.split(chunk_key_encoding))
 
 
-def get_deepest_group_or_array(
-    node: ManifestGroup, key: Iterable[str]
-) -> ManifestGroup | ManifestArray:
-    for var in key:
+def _get_deepest_group_or_array(
+    node: ManifestGroup | ManifestArray, key: str
```
chuckwondo: It looks like `node` will always be a `ManifestGroup`, no?

```suggestion
    node: ManifestGroup, key: str
```

virtualizarr/manifests/store.py
```diff
@@ -98,14 +98,41 @@ def parse_manifest_index(
...
-    return node
+            node = node.arrays[var]
+        elif var in node.groups:
+            node = node.groups[var]
+        else:
+            # Can't traverse deeper - return last node and remainder
+            return node, "/".join(key_list[i:])
+
+    # All parts matched - return empty tuple for remainder
+    return node, ""
```
chuckwondo: Now I see why you annotated the `node` param as `ManifestGroup | ManifestArray`, but we don't want to do that because we only pass in a `ManifestGroup`.

To avoid the annotation issue, I suggest you simplify the implementation like so (which _also_ allows you to pass in an empty string for `key` so you don't have to check for the case like you do elsewhere, before calling this function):

```suggestion
    var, suffix = key.split("/", 1) if "/" in key else (key, "")

    if var in node.arrays:
        return node.arrays[var], suffix
    if var in node.groups:
        return _get_deepest_group_or_array(node.groups[var], suffix)

    # Can't traverse deeper - return last node and remainder
    return node, suffix
```

virtualizarr/manifests/store.py
```diff
@@ -285,10 +312,44 @@ def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
...
-        yield "zarr.json"
-        for k in self._group.arrays.keys():
-            yield k
+        # docstring
+        # Navigate to the target node
+        if not prefix:
+            node: ManifestArray | ManifestGroup = self._group
+            suffix = ""
+        else:
+            node, suffix = _get_deepest_group_or_array(self._group, prefix)
```
chuckwondo: If you accept my suggestion for `_get_deepest_group_or_array`, then you can simplify things here too:

```suggestion
        # docstring inherited
        # Navigate to the target node
        node, suffix = _get_deepest_group_or_array(self._group, prefix)
```

virtualizarr/manifests/store.py
```diff
@@ -285,10 +312,44 @@ def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
...
+            if (
+                separator := getattr(
+                    node.metadata.chunk_key_encoding, "separator", None
+                )
+                != "."
+            ):
+                raise NotImplementedError(
+                    f"Array listing only supports '.' as chunk key separator, "
+                    f"got {separator!r}"
+                )
```
chuckwondo: I think this is _slightly_ simpler:

```suggestion
        # TODO: Support listing when using other chunk_key_encodings
        elif (
            separator := getattr(node.metadata.chunk_key_encoding, "separator", None)
            != "."
        ):
            raise NotImplementedError(
                f"Array listing only supports '.' as chunk key separator, "
                f"got {separator!r}"
            )
        else:
 ```

chuckwondo (changes requested):

TomNicholas (approved):

PR #838, opened 2025-11-14
Feat: add top-level open_virtual_datatree

maxrjones:
This PR adds `open_virtual_datatree` for virtualizing higher-level hierarchical data structures.

Builds on #837 

TODO:
- [x] Unit tests
- [x] Test with other backends (e.g., HDF)

Checklist:
- [ ] Closes https://github.com/zarr-developers/VirtualiZarr/issues/84
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


virtualizarr/manifests/group.py
```diff
@@ -125,3 +125,37 @@ def to_virtual_dataset(self) -> xr.Dataset:
             coord_names=coord_names,
             attrs=attributes,
         )
+
+    def to_virtual_datasets(self) -> dict[str, xr.Dataset]:
```
chuckwondo: Consider naming this method `flatten`.

chuckwondo:
```suggestion
    def to_virtual_datasets(self) -> Mapping[str, xr.Dataset]:
```

virtualizarr/manifests/group.py
```diff
@@ -125,3 +125,37 @@ def to_virtual_dataset(self) -> xr.Dataset:
...
+            # Add the subgroup's datasets with proper path prefixes
+            for subpath, dataset in subgroup_datasets.items():
+                if subpath == "":
+                    # Direct child group
+                    full_path = group_name
+                else:
+                    # Nested subgroup
+                    full_path = f"{group_name}/{subpath}"
+                result[full_path] = dataset
+        return result
```
chuckwondo, line 152: Take it or leave it, but you can simplify to the following, if you like:

```suggestion
        return {"": self.to_virtual_dataset()} | {
            (group_name if subpath == "" else f"{group_name}/{subpath}"): dataset
            for group_name, subgroup in self.groups.items()
            for subpath, dataset in subgroup.to_virtual_datasets().items()
        }
```

virtualizarr/manifests/group.py
```diff
@@ -125,3 +125,37 @@ def to_virtual_dataset(self) -> xr.Dataset:
...
+
+    def to_virtual_datatree(self) -> xr.DataTree:
+        """
+        Create a "virtual" [xarray.DataTree][] containing the contents of one zarr group.
+
+        All variables in the returned DataTree will be "virtual", i.e. they will wrap ManifestArray objects.
+        """
+        datasets = self.to_virtual_datasets()
+        return xr.DataTree.from_dict(datasets)
```
chuckwondo: No biggie, but any reason you didn't just inline the call, like so?

```suggestion
        return xr.DataTree.from_dict(self.to_virtual_datasets())
```

virtualizarr/tests/test_xarray.py
```diff
@@ -655,6 +659,54 @@ def test_open_subgroup(
...
+            with open_datatree(
+                netcdf4_file_with_data_in_sibling_groups, engine="h5netcdf"
+            ) as dt:
+                vdt.isomorphic(dt)
+                assert list(vdt["/subgroup1"].variables) == ["foo"]
+                assert isinstance(vdt["/subgroup1"]["foo"].data, ManifestArray)
+                assert vdt["/subgroup1"]["foo"].shape == (3,)
+                assert list(vdt["/subgroup2"].variables) == ["bar"]
+                assert isinstance(vdt["/subgroup2"]["bar"].data, ManifestArray)
+                assert vdt["/subgroup2"]["bar"].shape == (2,)
```
chuckwondo: Consider combining context managers and inlining the parser:

```suggestion
        with (
            open_virtual_datatree(
                url=netcdf4_file_with_data_in_sibling_groups,
                registry=local_registry,
                parser=HDFParser(),
            ) as vdt,
            open_datatree(
                netcdf4_file_with_data_in_sibling_groups, engine="h5netcdf"
            ) as dt,
        ):
            vdt.isomorphic(dt)
            assert list(vdt["/subgroup1"].variables) == ["foo"]
            assert isinstance(vdt["/subgroup1"]["foo"].data, ManifestArray)
            assert vdt["/subgroup1"]["foo"].shape == (3,)
            assert list(vdt["/subgroup2"].variables) == ["bar"]
            assert isinstance(vdt["/subgroup2"]["bar"].data, ManifestArray)
            assert vdt["/subgroup2"]["bar"].shape == (2,)
```

virtualizarr/tests/test_xarray.py
```diff
@@ -655,6 +659,54 @@ def test_open_subgroup(
...
+        with open_virtual_datatree(
+            url=netcdf4_file_with_data_in_sibling_groups,
+            registry=local_registry,
+            parser=parser,
+            loadable_variables=["foo", "bar"],
+        ) as vdt:
+            with open_datatree(
+                netcdf4_file_with_data_in_sibling_groups, engine="h5netcdf"
+            ) as dt:
+                xr.testing.assert_allclose(vdt, dt)
```
chuckwondo: Another place to consider combining context managers and inlining the parser:

```suggestion
        with (
            open_virtual_datatree(
                url=netcdf4_file_with_data_in_sibling_groups,
                registry=local_registry,
                parser=HDFParser(),
                loadable_variables=["foo", "bar"],
            ) as vdt,
            open_datatree(
                netcdf4_file_with_data_in_sibling_groups, engine="h5netcdf"
            ) as dt,
        ):
            xr.testing.assert_allclose(vdt, dt)
```

virtualizarr/xarray.py
```diff
@@ -354,6 +415,44 @@ def construct_virtual_dataset(
...
+
+    partially_loaded_datasets = {}
+    for name, virtual_node in fully_virtual_datatree.subtree_with_keys:
+        loadable_node = fully_loadable_datatree[name]
+        node_dataset = replace_virtual_with_loadable_vars(
+            virtual_node.to_dataset(), loadable_node.to_dataset(), loadable_variables
+        )
+        node_dataset = node_dataset.drop_vars(list(drop_variables or ()))
+        partially_loaded_datasets[name] = node_dataset
+    return xr.DataTree.from_dict(partially_loaded_datasets)
```
chuckwondo, line 525: Another suggestion for (arguable) simplification (also another encouragement to embrace comprehensions):

```suggestion
    node = manifest_store._group[group] if group else manifest_store._group

    if isinstance(node, ManifestArray):
        node = ManifestGroup(arrays={group: node}, attributes={})

    fully_loadable_datatree = xr.open_datatree(
        manifest_store,  # type: ignore[arg-type]
        group=group,
        engine="zarr",
        consolidated=False,
        zarr_format=3,
        decode_times=decode_times,
    )

    partially_loaded_datasets = {
        name: replace_virtual_with_loadable_vars(
            virtual_node.to_dataset(),
            fully_loadable_datatree[name].to_dataset(),
            loadable_variables,
        ).drop_vars(list(drop_variables or ()))
        for name, virtual_node in node.to_virtual_datatree().subtree_with_keys
    }

    return xr.DataTree.from_dict(partially_loaded_datasets)
```

chuckwondo (approved):
@maxrjones, this looks great. Approved!

Feel free to ignore any or all of my suggested changes, which are only suggestions for some simplification, which are certainly arguable, and thus may not appear simpler to others. I leave them all to your discretion.

virtualizarr/manifests/group.py
maxrjones: Thanks for the review @chuckwondo! This is the only suggestion that I'm leaning towards leaving because I think the comments are helpful for understanding the logic

virtualizarr/xarray.py
```diff
@@ -35,6 +35,68 @@
...
+        - [virtualizarr.parsers.KerchunkJSONParser][] for re-opening Kerchunk JSONs.
+        - [virtualizarr.parsers.KerchunkParquetParser][] for re-opening Kerchunk Parquets.
+        - [virtualizarr.parsers.ZarrParser][] for virtualizing Zarr stores.
+        - [virtualizarr.parsers.ZarrParser][] for virtualizing Zarr stores.
+        - [virtual_tiff.VirtualTIFF][] for virtualizing TIFFs.
+
+    drop_variables
+        Variables in the data source to drop before returning.
+    loadable_variables
+        Variables in the data source to load as Dask/NumPy arrays instead of as virtual arrays.
```
TomNicholas: This seems under-specified. Can I load/drop specific variables from one group? If so how? Does it understand a path-like string?

maxrjones: I added more details and examples in [`9b91ad3` (#838)](https://github.com/zarr-developers/VirtualiZarr/pull/838/commits/9b91ad3b130db12ead32ad8e73d6a02dce32d3cc)

maxrjones: I should ask, do you approve of the behavior that's now described regarding loadable/droppable variables @TomNicholas? we should avoid breaking changes obviously, so this is an important component to get right

TomNicholas: I'm not a big fan of the current map-over-groups behaviour, because IMO it only makes sense for overviews-style-datatrees. Accepting path-like strings is more effort but should work. Unfortunately this whole API design problem is just a little bit messy 😞 

I think maybe this is an argument for implementing something like `vdt["var"].load()`...

maxrjones: Ok, so what if we only implement the default behavior in this PR and punt on adding loadable/drop_variables? It's the end of the project increment for me, so I'd like to get something in before I get pulled onto other work and splitting PRs is good practice anyways.

We'd need to decide on the default behavior between:
1. Same as xarray (load coordinates)
2. Don't load anything

TomNicholas: I'm okay with that, because then generalizing later wouldn't be a breaking change.

> We'd need to decide on the default behavior

Maybe you could keep both these options by keeping the kwargs, but raising a `NotImplementedError` if the user attempts to pass an a non-empty list.

TomNicholas (approved):

PR #844, opened 2025-12-19
Move ObjectStoreRegistry and Reader functionality to obspec_utils

maxrjones:
I think that `ObstoreReader` and `ObjectStoreRegistry` are useful outside of virtualizarr (e.g., for earthaccess) and should therefore be a separate library. I started [obspec-utils](https://github.com/virtual-zarr/obspec-utils) accordingly, in line with some past discussions.


TomNicholas:
This is technically a breaking change to the API, because the type of he ObjectStoreRegistry has changed. (Note that it literally changes the `Parser` protocol definition too). This justifies a minor version bump at least...

TomNicholas (approved):
Seems good to me

maxrjones:
> This is technically a breaking change to the API, because the type of he ObjectStoreRegistry has changed. (Note that it literally changes the `Parser` protocol definition too). This justifies a minor version bump at least...

I'll think on this a bit more. I forgot that a reason for obspec-utils was not only to have the code available as a lighter dependency but also to change the typing from obstore-based to obspec-based, which would be good to test before obspec-utils is a dependency of virtualizarr

TomNicholas (approved):
Shall we just merge this? It all looks great to me.

docs/api/parsers/hdf5.md
```diff
@@ -1,3 +1,4 @@
 # HDF5/NetCDF4
 
+::: virtualizarr.parsers.typing.ReaderFactory
```
TomNicholas:
```suggestion
```

I'm going to remove this because it seems like such a simple type hint that it doesn't need to be declared as public API.

PR #845, opened 2025-12-24
Fix `fill_value=null` zarr v2 to v3 conversion for string dtypes

### LINKED ISSUE #826: Can't concat virtual datasets without dropping CF variables
I've got some virtual datasets that I need to concatenate along the time dimension.  

Following the CF conventions, the dataset variables include `time_bnds` and the `grid_mapping` variable `rotated_pole`.

Creating a list of virtual datasets with Virtualizarr works fine.   

Here's what one virtual dataset in the list looks like:

Image

If I drop the `time_bnds` and `rotated_pole`  variables from each virtual dataset, I can concatenate with:
``` python
combined_ds = xr.concat(
    ds_list,
    dim="time",
    coords="minimal",
    compat="override",
    combine_attrs="override",
)
```
but if I don't drop them, I get:
``` python-traceback
File /srv/conda/envs/notebook/lib/python3.13/site-packages/xarray/core/duck_array_ops.py:261, in asarray(data, xp, dtype)
    258     return converted
    260 if xp is np or not hasattr(xp, "astype"):
--> 261     return converted.astype(dtype)
    262 else:
    263     return xp.astype(converted, dtype)

File /srv/conda/envs/notebook/lib/python3.13/site-packages/virtualizarr/manifests/array.py:209, in ManifestArray.astype(self, dtype, copy)
    207 """Cannot change the dtype, but needed because xarray will call this even when it's a no-op."""
    208 if dtype != self.dtype:
--> 209     raise NotImplementedError()
    210 else:
    211     return self

NotImplementedError: 
```

A reproducer notebook is here: https://gist.github.com/rsignell/88bd394bc76f30c6f5517da263490d9d

observingClouds:
Adds support to convert unset/`null` fill values from zarr format 2 metadata to format 3 for string types.

- [x] Closes #826  (at least the [dtype part](https://github.com/zarr-developers/VirtualiZarr/issues/826#issuecomment-3512308665))
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage
~~- [ ] New functions/methods are listed in `api.rst`~~
~~- [ ] New functionality has documentation~~


observingClouds:
@maxrjones and others please have a look now. I am not sure if I did the casting correctly to fix the mypy error, but all is passing now. I have also tested that my test is failing with the previous version.

virtualizarr/parsers/zarr.py
```diff
@@ -194,20 +195,13 @@ def get_metadata(self, zarr_array: ZarrArrayType) -> ArrayV3Metadata:
 
         if v2_metadata.fill_value is None:
             v2_dict = v2_metadata.to_dict()
-            v2_dict["fill_value"] = 0
+            v2_dtype = parse_dtype(cast(Any, v2_dict["dtype"]), zarr_format=2)
+            fill_value = v2_dtype.default_scalar()
+            v2_dict["fill_value"] = (
+                fill_value.item() if hasattr(fill_value, "item") else fill_value
+            )
```
maxrjones:
```suggestion
            v2_dict["fill_value"] = v2_dtype.to_json_scalar(fill_value, zarr_format=2)
```
This should make it more certain that the result is json serializable

virtualizarr/tests/test_parsers/test_zarr.py
```diff
@@ -264,7 +264,17 @@ async def get_meta():
...
-def test_v2_metadata_with_none_fill_value():
+@pytest.mark.parametrize(
+    "dtype",
+    [
+        "int32",
+        "uint8",
+        "float64",
+        "bool",
+        "U10",
+    ],
```
maxrjones, line 279:
```suggestion
        "U10",
        "datetime64[s]",
        "timedelta64[s]",
        "S10",
        "V10",
    ],
```
Some examples that would fail using `fill_value.item()`

maxrjones (changes requested):
Thanks @observingClouds 

I think the other fix from https://github.com/zarr-developers/VirtualiZarr/issues/826#issuecomment-3690340280 (`if not np.issubdtype(self.dtype, dtype):`) is also needed to close out #826. Do you want to include that fix in this PR?

maxrjones (approved):
thanks for this fix!

PR #847, opened 2026-01-05
Fix typing errors

maxrjones:
Addresses this failing workflow - https://github.com/zarr-developers/VirtualiZarr/actions/runs/20701631805

### DIFF
--- a/virtualizarr/manifests/array.py
+++ b/virtualizarr/manifests/array.py
@@ -148,7 +148,9 @@ class ManifestArray:
             return _isnan(self.shape)
         return NotImplemented
 
-    def __array__(self, dtype: np.typing.DTypeLike | None = None, copy: bool | None = None) -> np.ndarray:
+    def __array__(
+        self, dtype: np.typing.DTypeLike | None = None, copy: bool | None = None
+    ) -> np.ndarray:
         raise NotImplementedError(
             "ManifestArrays can't be converted into numpy arrays or pandas Index objects"
         )


virtualizarr/manifests/array_api.py
```diff
@@ -97,9 +97,14 @@ def concatenate(
     new_shape[axis] = new_length_along_concat_axis
 
     # do concatenation of entries in manifest
-    concatenated_paths = np.concatenate(
-        [arr.manifest._paths for arr in arrays],
-        axis=axis,
+    concatenated_paths = (
+        cast(  # `np.concatenate` is type hinted as if the output could have Any dtype
```
TomNicholas, line 101: I presume this is because of numpy's [type promotion rules](https://numpy.org/doc/2.3/reference/arrays.promotion.html).

TomNicholas (approved):

virtualizarr/manifests/array.py
```diff
@@ -148,7 +148,7 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs) -> Any:
             return _isnan(self.shape)
         return NotImplemented
 
-    def __array__(self, dtype: np.typing.DTypeLike = None) -> np.ndarray:
+    def __array__(self, dtype: None = None) -> np.ndarray:
```
chuckwondo: I believe the correct signature is this:

```suggestion
    def __array__(self, dtype: np.typing.DTypeLike | None = None, copy: bool | None = None) -> np.ndarray:
```

chuckwondo: Sorry, `DTypeLike` actually includes `None`, so no need for `| None` on the `dtype` parameter.

chuckwondo: Updated:

```
    def __array__(self, dtype: np.typing.DTypeLike = None, copy: bool | None = None) -> np.ndarray:
```

chuckwondo: Oh, wait. I see the issue: numpy <2.4 include `None` in `DTypeLike`, but numpy >=2.4 does _not_ include it, so my original suggestion is necessary (explicit union with `None`).

This seems to raise another problem. I just noticed that we do not commit `uv.lock`, which I discovered because I was _unable_ to reproduce the typing errors on `main`, which led me to check my environments version of numpy with the version listed in the CI output. Mine was 2.3.4, but CI shows 2.4.0. Once I ran `pixi update`, I got 2.4.0 and then running mypy produced the same errors.

Perhaps we should open a separate issue regarding committing `uv.lock` so that we can readily reproduce an environment with identical dependencies.

chuckwondo: @TomNicholas, any reason why `uv.lock` is _not_ committed?

chuckwondo: Sorry, my bad. We're using `pixi`. For some reason I had a random `uv.lock` file sitting in my clone.

However, I'm puzzled as to why pulling from `main` and completely rebuilding my `test` pixi env did _not_ reproduce the typing error. I had to explicitly run `pixi update` to get the matching numpy dependency that CI used to produce the error (no idea if other deps were mismatched with CI after updating, but at least numpy matched, which was the problem).

chuckwondo: Oh, perhaps because `pixi.lock` is not committed. Thus, my question remains (but for `pixi.lock`, rather than `uv.lock`).

maxrjones: Let's pull out the pixi lock discussion into a separate issue. That was my decision and I'd be glad to discuss the choice more. For this PR, do you think using the typing as you suggest is the appropriate resolution?

chuckwondo: Yes, let's have a separate discussion about `pixi.lock`.

For this PR, yes, I believe the appropriate change in array.py should be this one:

```python
    def __array__(self, dtype: np.typing.DTypeLike | None = None, copy: bool | None = None) -> np.ndarray:
```

(I'm glad the numpy maintainers pulled `None` out of the `DTypeLike` type definition.)

PR #855, opened 2026-01-24
Add example of virtualizing GOES using caching and request splitting

maxrjones:
This PR shows how to cache entire files before virtualizing.

Full explanation: for data formats that use b-link trees (like non-cloud-optimized version of HDF5), you'll probably want to cache the entire file up front. If you also want to load many variables, you'll want to cache the file's contents at the store level rather than at the parser/file reader level so that it's accessible by ManifestStore as well as the parser. This PR adds an example of how to do that using new features in [obspec-utils](https://obspec-utils.readthedocs.io/en/latest/). It relies on the sequence of PRs discussed in #844, which makes VirtualiZarr work with any ReadableStore using duck-type rather than only the stores implemented in obstore.

It reduces the amount of time to virtualize a single GOES-16 file on my laptop from ~47s to ~8s ⚡ 

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/examples/V2/goes_with_caching_stores.py
+++ b/examples/V2/goes_with_caching_stores.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "virtualizarr @ git+https://github.com/maxrjones/VirtualiZarr@obspec-utils-readablefile-protocol",
+#     "obstore",
+#     "obspec-utils",
+#     "xarray",
+#     "numpy",
+#     "h5py",
+# ]
+# ///
+"""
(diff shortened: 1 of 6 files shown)


examples/V2/goes_with_caching_stores.py
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "virtualizarr @ git+https://github.com/zarr-developers/VirtualiZarr@main",
```
TomNicholas:
```suggestion
#     "virtualizarr>=0.2.4",
```

PR #873, opened 2026-02-08
Implement to_icechunk(region=...)

### LINKED ISSUE #308: Build virtual Zarr store using Xarray's dataset.to_zarr(region="...") model
I'm wondering how we can build virtualization pipelines as a map rather than a map-reduce process. The map paradigm would follow the structure below from [Earthmover's excellent blog post on serverless datacube pipelines](https://earthmover.io/blog/serverless-datacube-pipeline), where the virtual dataset from each worker would get written directly to an Icechunk Virtual Store rather than being transferred back to the coordination node for a concatenation step. Similar to their post, the parallelization between workers during virtualization could leverage a serverless approach like lithops or coiled. Dask concurrency could speed up virtualization within each worker. I think the main feature needed for this to work is an analogue to `xr.Dataset.to_zarr(region="...")`, complementary to `xr.Dataset.to_zarr(append_dim="...")`(documented in https://docs.xarray.dev/en/latest/generated/xarray.Dataset.to_zarr.html)  (xref https://github.com/zarr-developers/VirtualiZarr/issues/21, https://github.com/zarr-developers/VirtualiZarr/pull/272).

Image

I tried to check that this feature request doesn't already exist, but apologies if I missed something.

vladidobro:
- [x] Closes #308 
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -548,8 +506,7 @@ def test_append_virtual_ref_with_encoding(
             )
 
     ## When appending to a virtual ref with encoding, it succeeds
-    @pytest.mark.asyncio
```
vladidobro, line 551: unrelated, but I believe this asyncio mark does not belong here

virtualizarr/writers/icechunk.py
```diff
@@ -52,8 +55,14 @@ def virtual_dataset_to_icechunk(
     append_dim
         Name of the dimension along which to append data. If provided, the dataset must
         have a dimension with this name.
+    region
+        Optional mapping from dimension names to either a) ``"auto"``, or b) integer
```
vladidobro, line 60: copied the first line from xarray docs

virtualizarr/writers/icechunk.py
```diff
@@ -94,6 +103,12 @@ def virtual_dataset_to_icechunk(
             f"append_dim: expected type Optional[str], but got type {type(append_dim)}"
         )
 
+    if not isinstance(region, (type(None), str, Mapping)):
```
vladidobro, line 107: this does not fully type check it, is that needed?

virtualizarr/writers/icechunk.py
```diff
@@ -153,7 +170,7 @@ def virtual_datatree_to_icechunk(
         originally defined. This saves disk space, but requires opening the
         full tree to load inherited coordinates.
     validate_containers
-        If ``True``, raise if any virtual chunks have a refer to locations that don't
+        If ``True``, raise if any virtual chunks refer to locations that don't
```
vladidobro, line 174: looks like a typo

virtualizarr/writers/icechunk.py
```diff
@@ -164,6 +181,8 @@ def virtual_datatree_to_icechunk(
         time, icechunk will raise an error at runtime when trying to read the virtual
         chunk. When not specified, icechunk will not check for modifications to the
         virtual chunks at runtime.
+    **kwargs
+        Additional keyword arguments to be passed to ``xarray.Dataset.vz.to_icechunk``.
```
vladidobro, line 186: this is not technically true, as DataTree.vz.to_icechunk does not call Dataset.vz.to_icechunk, but instead the private `write_virtual_dataset_to_icechunk_group`, but this explanation (inspired by xarray) seems better than referencing private functions

virtualizarr/writers/icechunk.py
```diff
@@ -284,14 +318,20 @@ def write_virtual_dataset_to_icechunk_group(
 
     # First write all the non-virtual variables
     if loadable_variables:
+        if append_dim is None and region is None:
+            mode = "a"
+        else:
+            # let xarray set it automatically, 'a' for append_dim and 'r+' for region
```
vladidobro, line 316: i think this best mimics what xarray is doing in pathological examples.
setting region should drop all dimension variables from the dataset, so they will no longer be in "loadable_variables"

virtualizarr/writers/icechunk.py
```diff
@@ -359,15 +400,16 @@ def get_axis(
 
 
 def check_compatible_arrays(
-    ma: "ManifestArray", existing_array: "Array", append_axis: int
+    ma: "ManifestArray", existing_array: "Array", append_axis: int | None
```
vladidobro: i want to reuse this function for region, when there is no append_axis

virtualizarr/writers/icechunk.py
```diff
@@ -434,37 +496,24 @@ def write_virtual_variable_to_icechunk(
...
-            f"append_axis {append_axis} is greater than the number of indices {len(index)}"
-        )
-
-    return [
-        ind + existing_num_chunks
-        if axis is append_axis and existing_num_chunks is not None
-        else ind
-        for axis, ind in enumerate(index)
-    ]
+    return [index + offset for index, offset in zip(index, chunk_index_offsets)]
```
vladidobro: it seemed better to reuse this for region too. append_dim is just a special case where
```
                                            append_axis
                                                 |
                                                 v
chunk_index_offsets = [0, 0, ..., 0, existing_num_chunks, 0, ..., 0, 0]
```

I guess like this it will be a bit faster too, with fewer code in the hottest path (though there is one more addition per index)

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -394,47 +393,6 @@ def test_checksum(
         npt.assert_equal(pressure_array, arr)
 
 
-def test_generate_chunk_key_no_offset():
```
vladidobro, line 397: i deleted tests for generate_chunk_key because of how trivial it is now.

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -710,6 +667,246 @@ def test_append_dim_not_in_dims_raises_error(
...
+                zarr_format=3,
+            ) as ds,
+        ):
+            is_written = ds["y"].isin(my_vds["y"]).broadcast_like(ds)
+            xrt.assert_allclose(
+                is_written,
+                ds["foo"] != 0,
+            )
+
+    @pytest.mark.xfail(
```
vladidobro: TODO: i think these are related to line https://github.com/zarr-developers/VirtualiZarr/pull/873/changes#diff-b818d0a8a21eff7a602deefe4e928ce63a2da7ca4c8ec9c4e8c23db97cb17729R466
where i set the default to slice(None)

I believe that is how xarray handles this situation, but have to check

virtualizarr/writers/icechunk.py
```diff
@@ -268,8 +288,22 @@ def write_virtual_dataset_to_icechunk_group(
...
+        xarray_store = XarrayZarrStore(
+            zarr_group=group,
+            append_dim=append_dim,
+            write_region=region,
+            consolidate_on_close=False,
+            close_store_on_close=False,
+            mode="r+",
+        )
+        vds = xarray_store._validate_and_autodetect_region(vds)
+        region = xarray_store._write_region
```
aaron-kaplan: I notice that this references "private" attributes of a class that's not from this package. Is that considered ok in VirtualiZarr?

vladidobro: Hi,
in the referenced issue,
https://github.com/zarr-developers/VirtualiZarr/issues/308#issuecomment-3862083835

@TomNicholas said
> Whether or not we vendor xarray kind of depends on how much of xarray we would have to vendor... Some of the internals are more stable than others. I would start out by not vendoring anything and we'll see how fragile that looks.

virtualizarr/tests/test_writers/test_icechunk.py
TomNicholas: you're right, feel free to remove it as part of this PR (and make the test function synchronous). I can only assume that the test function _used_ to call some async code. 🤷‍♂️

virtualizarr/writers/icechunk.py
TomNicholas: Best effort is fine, the `region` kwarg is pretty overloaded

virtualizarr/writers/icechunk.py
TomNicholas: Yeah it's not "okay", but arguably the lesser of two evils. VirtualiZarr as a whole is a slightly weird re-purposing of xarray.

If this is the only internal import from xarray then we have done better than I expected for this feature.

TomNicholas (commented):
This is looking good @vladidobro !

virtualizarr/tests/test_writers/test_icechunk.py
```diff
@@ -710,6 +667,310 @@ def test_append_dim_not_in_dims_raises_error(
...
+
+        my_vds = vds_grid[1][1]
+        my_vds = my_vds.assign_coords(x=my_vds["x"] - 1)
+        write_session = initialized_repo.writable_session("main")
+        with pytest.raises(ValueError, match="is not aligned to whole chunks"):
+            my_vds.vz.to_icechunk(
+                write_session.store, region={"y": "auto", "x": "auto"}
+            )
+
+    @pytest.mark.xfail(
```
vladidobro, line 937: should I leave this test here? turns out I was unable write a datatree with xarray anyway

TomNicholas: what happened? This seems separable though.

TomNicholas (approved):

vladidobro:
Thank you for the fixes and deploy!

PR #879, opened 2026-02-20
Support indexing when it removes a dimension of length 1

aaron-kaplan:
This is to support `xr.DataArray.squeeze`. It implements the following comment that was in manifests/indexing.py:

> \# TODO cover possibility of indexing into a length-1 dimension (which just removes that dimension)?

I'll raise a couple points as in-line comments.

todo:
- [ ] Convert the removed coordinate to a scalar attribute

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/indexing.py
+++ b/virtualizarr/manifests/indexing.py
@@ -2,8 +2,11 @@ from types import EllipsisType
 from typing import TYPE_CHECKING, TypeAlias, cast
 
 import numpy as np
+from zarr.core.chunk_grids import RegularChunkGrid
+from zarr.core.metadata.v3 import ArrayV3Metadata
 
 from virtualizarr.manifests.array_api import expand_dims
+from .manifest import ChunkManifest
 
 # indexer with only basic selectors, no new axes or ellipsis
 T_BasicIndexer_1d: TypeAlias = int | slice | np.ndarray
@@ -147,18 +150,18 @@ def apply_selection(
     for axis, (length, indexer_1d) in enumerate(
         zip(marr.shape, indexer_without_newaxes)
     ):
-        output_arr = apply_selection_1d(output_arr, indexer_1d, length)
+        output_arr = apply_selection_1d(output_arr, indexer_1d, axis, length)
 
     return output_arr
 
 
 def apply_selection_1d(
-    marr: "ManifestArray", indexer_1d: T_BasicIndexer_1d, length: int
+    marr: "ManifestArray", indexer_1d: T_BasicIndexer_1d, axis: int, length: int
 ) -> "ManifestArray":
     """
     Actually index the ManifestArray along 1 dimension.
 
-    Notice that none of these options actually do any indexing right now!
+    Notice that most of these options don't actually do any indexing right now!
     """
 
     if isinstance(indexer_1d, slice):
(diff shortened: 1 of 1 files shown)


virtualizarr/manifests/indexing.py
```diff
@@ -2,8 +2,11 @@
 from typing import TYPE_CHECKING, TypeAlias, cast
 
 import numpy as np
+from zarr.core.chunk_grids import RegularChunkGrid
+from zarr.core.metadata.v3 import ArrayV3Metadata
 
 from virtualizarr.manifests.array_api import expand_dims
+from .manifest import ChunkManifest
```
aaron-kaplan, line 9: Sibling imports avoid loading the parent package's __init__.py, which sometimes resolves circular import problems. As I write this I can't remember if this was actually necessary in this case. I'll check.

virtualizarr/manifests/indexing.py
```diff
@@ -147,18 +150,18 @@ def apply_selection(
...
     ):
-        output_arr = apply_selection_1d(output_arr, indexer_1d, length)
+        output_arr = apply_selection_1d(output_arr, indexer_1d, axis, length)
 
     return output_arr
 
 
 def apply_selection_1d(
-    marr: "ManifestArray", indexer_1d: T_BasicIndexer_1d, length: int
+    marr: "ManifestArray", indexer_1d: T_BasicIndexer_1d, axis: int, length: int
```
aaron-kaplan, line 159: This function signature wasn't dictated by xarray, was it? Seems like it's not implementable without the axis parameter.

virtualizarr/manifests/indexing.py
```diff
@@ -169,10 +172,27 @@ def apply_selection_1d(
...
+                    dimension_names=None, # TODO
+                )
+            )
+            manifest = ChunkManifest.from_arrays(
+                paths=np.squeeze(marr.manifest._paths, axis=axis),
+                offsets=np.squeeze(marr.manifest._offsets, axis=axis),
+                lengths=np.squeeze(marr.manifest._lengths, axis=axis),
+                validate_paths=False
+            )
+            marr = marr.__class__(metadata, manifest)
```
aaron-kaplan, line 191: I would rather say `ManifestArray` instead of `marr.__class__` for clarity, but importing array.py led to a circular import. Seems to me that this function should really be a method of `ManifestArray`, not split out into a separate file.

TomNicholas: The more conventional fix is to just import `ManifestArray` inside the function that needs it.

virtualizarr/manifests/indexing.py
TomNicholas: No, these are all implementation details, as xarray only cares about the signatures of certain methods on the `ManifestArray` class (for indexing `__getitem__` is what matters). So feel free to alter the the functions in this file as necessary. (That is one reason I put the indexing internals in this separate `indexing.py` file.)

virtualizarr/manifests/indexing.py
TomNicholas: It shouldn't be. If needed I think you can just import `ChunkManifest` inside `array_selection_1d` instead.

TomNicholas:
This should now have been included as part of #994, superceding this PR

PR #880, opened 2026-02-20
Fix dmrpp error handling

betolink:
I think a while ago we talked about eventually moving the parser to pydap as that would be the more native place to improve it and fix bugs, I think we should move in that direction and offer a thin wrapper for virtualization. However that integration is still TDB, this PR addresses some of the quirkiness of some DMRpp files mostly by adding a safe attribute getter function. Unit tests were added and now the parsers shows up as one of the valid parsers for VirtualiZarr. 

- [ ] Closes #xxxx
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

p.s. once #794 gets merged, we should be able to create and append Zarr stores from NASA earthdata using DMRpp + kerchunk really fast! 

cc @Mikejmnez @owenlittlejohns @TomNicholas @danielfromearth


betolink:
@Mikejmnez nice!! should we try to merge this and then migrate once the pydap parser is ready and gets released? we could also integrate this PR (the idea of the safe attr getter) into your branch and close this PR. The next PR here would be to use PyDAP. I'm open to both options.

betolink:
@Mikejmnez one note, Max PR is for when VZ already parsed these into Kerchunk, so then we can open that store and append without having to recreate it all from scratch. 

This PR is only logging "{var} could not be parsed because of {x}" (instead of just the parser blowing) and also enumerates phony dimensions, right now all they become "phony_dim" instead of "phony_dim_{n}". I'm excited to see your work and Max's in VZ, we'll be in a position to roundtrip chunkmanifest with inline values!

virtualizarr/parsers/dmrpp.py
```diff
@@ -389,8 +454,8 @@ def _parse_variable(self, var_tag: ET.Element) -> ManifestArray:
         # Dimension info
         dims: dict[str, int] = {}
         dimension_tags = self._find_dimension_tags(var_tag)
-        for dim in dimension_tags:
-            dims.update(self._parse_dim(dim))
+        for dim_index, dim in enumerate(dimension_tags):
```
Mikejmnez: @betolink It appears to me that this enumeration is being set at the variable level, correct? What happens if there are two different variables with unnamed dimensions? This bit suggests the counting resets to zero, right? Perhaps an edge case, but perhaps not?

TomNicholas (approved):
lemme know when you want me to merge

PR #890, opened 2026-02-25
Correct Scaling docs. 

jbusecke:
This PR will incorporate some of the findings from #889 since I am stalled on that one. 

My process here:
- [x] Add `parallel=ProcessPoolExecutor` as testing option to confirm that it generally fails (my assumption from the initial motivation of #889)
- [x] Adjust docs:
    - Remove recommendation of ProcessPoolExecutor
    - Warn that using ThreadPoolExecutor + HDFParser will basically result in serial processing
    - Suggest to use local lithops or dask until #889 is fixed.

### DIFF
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -100,8 +100,11 @@ It is the default executor.
 ### Threads or Processes
 
 One way to parallelize creating virtual references from a single machine is to across multiple threads or processes.
-For this you can use the [`ThreadPoolExecutor`][concurrent.futures.ThreadPoolExecutor] or [`ProcessPoolExecutor`][concurrent.futures.ProcessPoolExecutor] class from the [`concurrent.futures`][] module in the python standard library.
+For this you can use the [`ThreadPoolExecutor`][concurrent.futures.ThreadPoolExecutor] or ~[`ProcessPoolExecutor`][concurrent.futures.ProcessPoolExecutor]~[^1] class from the [`concurrent.futures`][] module in the python standard library.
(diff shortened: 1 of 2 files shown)


docs/scaling.md
```diff
@@ -100,9 +100,12 @@ It is the default executor.
 ### Threads or Processes
 
 One way to parallelize creating virtual references from a single machine is to across multiple threads or processes.
-For this you can use the [`ThreadPoolExecutor`][concurrent.futures.ThreadPoolExecutor] or [`ProcessPoolExecutor`][concurrent.futures.ProcessPoolExecutor] class from the [`concurrent.futures`][] module in the python standard library.
+For this you can use the [`ThreadPoolExecutor`][concurrent.futures.ThreadPoolExecutor] or ~[`ProcessPoolExecutor`][concurrent.futures.ProcessPoolExecutor]~[^1] class from the [`concurrent.futures`][] module in the python standard library.
 You simply pass the executor class directly via the `parallel` kwarg to [`open_virtual_mfdataset`][virtualizarr.open_virtual_mfdataset].
 
+!!! warning
+    [^1]: `ProcessPoolExecutor` does not currently work with VirtualiZarr. For progress on a fix, see [PR #889](https://github.com/zarr-developers/VirtualiZarr/pull/889).
```
TomNicholas: If it just doesn't work at all then we should remove any mention of it from the docs entirely.

TomNicholas (approved):

PR #892, opened 2026-02-26
Updates `zarr-parser` to use obstore `list_async` instead of concurrent_map

### LINKED ISSUE #891: Speed up ZarrParser using obstore and Arrow?
Our `ZarrParser` unfortunately has to list every chunk in the object store (see https://github.com/zarr-developers/VirtualiZarr/issues/850#issuecomment-3962495995). But I think we can make this a lot faster and less memory-intensive.

Currently we use a vendored function from zarr-python 

https://github.com/zarr-developers/VirtualiZarr/blob/785de91306f5b5ef5863d8a92c9c9e0751f7d98d/virtualizarr/vendor/zarr/core/common.py#L17

and use it to call `store.getsize` concurrently on all the (possible) keys of a zarr array.

https://github.com/zarr-developers/VirtualiZarr/blob/785de91306f5b5ef5863d8a92c9c9e0751f7d98d/virtualizarr/parsers/zarr.py#L118

Then the results go into a python dict

https://github.com/zarr-developers/VirtualiZarr/blob/785de91306f5b5ef5863d8a92c9c9e0751f7d98d/virtualizarr/parsers/zarr.py#L125

which becomes the chunk manifest for that array.

Instead what we could do is:
- use [`obstore.list`](https://developmentseed.org/obstore/v0.9.0/api/list/#obstore.list) to do the concurrent loop in rust instead of python
- pass `return_arrow=True` to get a stream of PyArrow RecordBatches back
- construct the python `ChunkManifest` object's numpy arrays[^1] directly from the Arrow arrays, minimizing memory copies (i.e. the opposite of what I did in #861)

I haven't benchmarked the current approach but I'm pretty sure this would be waaay faster.

[^1]: Unfortunately we can't just keep the manifests as arrow arrays the whole way through because of the potential need to concatenate manifests along arbitrary dims.

norlandrhagen:
- [x] Closes #891
- [x] Tests passing
- [x] Full type hint coverage

- [x] Swaps out the `_concurrent_map` in `build_chunk_mapping` with obstore's `list_async`.
- [x] Constructs the python ChunkManifest object's numpy arrays directly from the Arrow arrays. *
`*` There is still a conversion to a dict, so not quite. 
- [x] Bonus - removes the zarr vendor code.


virtualizarr/parsers/zarr.py
```diff
@@ -90,33 +93,43 @@ async def _handle_scalar_array(
...
+    # filter out metadata files
+    chunk_keys = [k for k in size_map if not k.split("/")[-1].startswith(".")]
+
     if not chunk_keys:
         return {}
 
-    lengths = await _concurrent_map(
-        [(k,) for k in chunk_keys], zarr_array.store.getsize
-    )
+    lengths = [size_map[k] for k in chunk_keys]
```
TomNicholas: I think we really want to work hard to avoid creating any python lists / dicts at all

TomNicholas: instead we want obstore -> arrow -> numpy

via https://arrow.apache.org/docs/python/numpy.html#arrow-to-numpy

TomNicholas: I think the hardest part of this is dealing with logic for missing keys - arrow might return these a nulls, but the to_numpy conversion doesn't support nulls?

Any operations we do should either be as pyarrow arrays or as numpy arrays, never as python collections

virtualizarr/parsers/zarr.py
```diff
@@ -90,33 +93,43 @@ async def _handle_scalar_array(
...
     -------
     dict
         Mapping of normalized chunk coordinates to storage locations.
     """
+
+    size_map: dict[str, int] = {}
+    stream = zarr_array.store.store.list_async(prefix=prefix, return_arrow=True)
+    async for batch in stream:
+        size_map.update(
+            zip(batch.column("path").to_pylist(), batch.column("size").to_pylist())
```
TomNicholas: is this zipping of pylists creating a python dict? we want to avoid that

virtualizarr/parsers/zarr.py
```diff
@@ -6,6 +6,7 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
 
+import numpy as np
```
TomNicholas, line 10: I would expect to see pyarrow being added as a dependency of the zarrparser too

virtualizarr/manifests/manifest.py
```diff
@@ -322,6 +327,55 @@ def from_arrays(
...
         return obj
 
+    @classmethod
+    def _from_arrow(
+        cls,
+        *,
+        chunk_keys: "pa.Array",
+        paths: "pa.Array",
+        sizes: "pa.Array",
+        chunk_grid_shape: tuple[int, ...],
```
TomNicholas: In general this should be 

```suggestion
        paths: "pa.StringArray",
        offsets: "pa.UInt64Array",
        lengths: "pa.UInt64Array",
```

because although in Zarr's case offsets are always zero it would be good not to assume that here.

virtualizarr/manifests/manifest.py
```diff
@@ -322,6 +327,55 @@ def from_arrays(
 
         return obj
 
+    @classmethod
+    def _from_arrow(
+        cls,
+        *,
+        chunk_keys: "pa.Array",
```
TomNicholas: I don't know that you need to pass this - maybe instead we should pass arrow arrays with nulls for unintialized chunks?

virtualizarr/parsers/zarr.py
```diff
@@ -90,41 +94,67 @@ async def _handle_scalar_array(
...
+    Tuple of (normalized_keys, full_paths, sizes) as PyArrow arrays, or None if no chunks found.
     """
-    if not chunk_keys:
-        return {}
+    import pyarrow as pa
+    import pyarrow.compute as pc
+
+    path_batches = []
+    size_batches = []
+    stream = zarr_array.store.store.list_async(prefix=prefix, return_arrow=True)
```
TomNicholas: Just grabbing the underlying obstore store is a interesting idea...

virtualizarr/manifests/manifest.py
```diff
@@ -322,6 +327,54 @@ def from_arrays(
...
+            .astype(np.dtypes.StringDType())
+        )
+        offsets_np = pc.if_else(
+            pc.is_null(offsets), pa.scalar(0, pa.uint64()), offsets
+        ).to_numpy(zero_copy_only=False)
+        lengths_np = pc.if_else(
+            pc.is_null(lengths), pa.scalar(0, pa.uint64()), lengths
+        ).to_numpy(zero_copy_only=False)
+
+        if shape is not None:
```
TomNicholas: What happens if shape is `None`? Should that even be allowed?

virtualizarr/manifests/manifest.py
```diff
@@ -322,6 +327,54 @@ def from_arrays(
...
+            pc.if_else(pc.is_null(paths), "", paths)
+            .to_numpy(zero_copy_only=False)
+            .astype(np.dtypes.StringDType())
+        )
+        offsets_np = pc.if_else(
+            pc.is_null(offsets), pa.scalar(0, pa.uint64()), offsets
+        ).to_numpy(zero_copy_only=False)
+        lengths_np = pc.if_else(
+            pc.is_null(lengths), pa.scalar(0, pa.uint64()), lengths
+        ).to_numpy(zero_copy_only=False)
```
TomNicholas: Lets split the arrow compute operations from the numpy conversions if only because it makes it easier to read.

virtualizarr/parsers/zarr.py
```diff
@@ -319,17 +366,84 @@ async def build_chunk_manifest(zarr_array: ZarrArrayType, path: str) -> ChunkMan
...
     strategy = get_strategy(zarr_array)
-    chunk_map = await strategy.get_chunk_mapping(zarr_array, path)
+    chunk_grid_shape = zarr_array._chunk_grid_shape
 
-    if not chunk_map:
-        if zarr_array.shape and zarr_array.chunks:
-            chunk_grid_shape = tuple(
-                math.ceil(s / c) for s, c in zip(zarr_array.shape, zarr_array.chunks)
-            )
+    # scalar arrays go through the dict path instead of the pure arrow bit
```
TomNicholas: It would be nice to not have to keep the whole old codepath around just for this special case...

virtualizarr/parsers/zarr.py
```diff
@@ -319,17 +366,84 @@ async def build_chunk_manifest(zarr_array: ZarrArrayType, path: str) -> ChunkMan
...
+
+    result = await _build_chunk_mapping(zarr_array, path, prefix)
+
+    if result is None:
+        return ChunkManifest({}, shape=chunk_grid_shape)
 
-    return ChunkManifest(chunk_map)
+    normalized_keys, full_paths, all_lengths = result
+
+    # Incoming: lots of LLM arrow mumbo jumbo for sparse arrays
```
TomNicholas: there's a lot going on here that I'm suspicious could be simplified

virtualizarr/tests/test_parsers/test_zarr.py
```diff
@@ -20,6 +20,7 @@
     get_strategy,
     join_url,
 )
+from virtualizarr.tests import requires_pyarrow
```
TomNicholas, line 23: It should also be possible to just call this function at the start of the file to skip the entire file if pyarrow isn't installed, then you don't need decorators on every test.

norlandrhagen: good call!

virtualizarr/parsers/zarr.py
norlandrhagen: Totally agree. I took a shot at trying to simplify it a bit. The handling of sparse arrays makes it a bit verbose.

virtualizarr/parsers/zarr.py
```diff
@@ -400,40 +393,21 @@ async def build_chunk_manifest(zarr_array: ZarrArrayType, path: str) -> ChunkMan
...
-        flat_positions = pa.repeat(pa.scalar(0, pa.int64()), len(normalized_keys))
-        for dim, dim_stride in enumerate(strides):
-            dim_indices = pc.list_slice(
-                pc.split_pattern(normalized_keys, pattern="."), dim, dim + 1
-            ).flatten()
-            flat_positions = pc.add(
-                flat_positions,
-                pc.multiply(pc.cast(dim_indices, pa.int64()), dim_stride),
-            )
+    split_keys = pc.split_pattern(normalized_keys, pattern=".")
```
TomNicholas: The chunk key encoding could also be `"/"` - we can probably read that from the `zarr.json` and use it here?

TomNicholas (approved):

TomNicholas (approved):
Yes, thank you so much @norlandrhagen !

virtualizarr/parsers/zarr.py
kylebarron: I think in this case you can assert that there are no nulls. I don't think this particular `list` function will ever create nulls in the arrow arrays.

virtualizarr/manifests/manifest.py
```diff
@@ -8,13 +10,16 @@
...
-from typing import Any, NewType, TypedDict, cast
+from typing import TYPE_CHECKING, Any, NewType, TypedDict, cast
 
 import numpy as np
 
 from virtualizarr.manifests.utils import construct_chunk_pattern, parse_manifest_index
 from virtualizarr.types import ChunkKey
 
+if TYPE_CHECKING:
+    import pyarrow as pa  # type: ignore[import-untyped,import-not-found]
```
kylebarron, line 21: I'd strongly suggest not tying to pyarrow

1. You should very easily be able to make your code generic and not tied to pyarrow
2. pyarrow doesn't have any internal type checking, so typing as `pa.StringArray` or `pa.Uint64Array` means absolutely nothing to the user (it might mean something to the developer)

pyproject.toml
```diff
@@ -50,6 +50,8 @@ hdf = [
     "imagecodecs-numcodecs==2024.6.1",
 ]
 
+zarr = ["arro3-core", "pyarrow"]
```
kylebarron, line 53: Feel free to disregard, but ideally you shouldn't need to depend on _both_ dependencies. Pyarrow is very large. I looked recently and it looks like it's gotten _even larger_

Image

50MB _compressed_ is huge.

norlandrhagen: Ah thanks for the feedback @kylebarron. Good to know about pyarrow.

virtualizarr/manifests/manifest.py
```diff
@@ -322,6 +327,55 @@ def from_arrays(
 
         return obj
 
+    @classmethod
+    def _from_arrow(
+        cls,
+        *,
+        paths: "pa.StringArray",
+        offsets: "pa.UInt64Array",
+        lengths: "pa.UInt64Array",
```
kylebarron, line 336: I'd suggest typing these as [`ArrowArrayExportable`](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html#protocol-typehints), and then using an arrow library of choice to import the data, such as passing input to `pyarrow.array` or `arro3.core.Array.from_arrow()`. 

Then this API will automatically support any arrow input, including polars, duckdb, arro3, etc https://github.com/apache/arrow/issues/39195#issuecomment-2245718008

virtualizarr/manifests/manifest.py
```diff
@@ -322,6 +327,55 @@ def from_arrays(
...
+        import pyarrow as pa  # type: ignore[import-untyped,import-not-found]
+        import pyarrow.compute as pc  # type: ignore[import-untyped,import-not-found]
+
+        arrow_paths = pc.if_else(pc.is_null(paths), "", paths)
+        arrow_offsets = pc.if_else(
+            pc.is_null(offsets), pa.scalar(0, pa.uint64()), offsets
+        )
+        arrow_lengths = pc.if_else(
+            pc.is_null(lengths), pa.scalar(0, pa.uint64()), lengths
+        )
```
kylebarron, line 365: Requiring a pyarrow dependency just for these three lines is not worth it IMO. Much better to just document that the users must remove any null values before passing in arguments.

kylebarron: And then you can probably use arro3-core for all your needs and save the big pyarrow dependency.

PR #899, opened 2026-03-05
Support functools.partial with lithops

chuckwondo:
This PR addresses some closely related issues with parallel executors:

- Lithops has an open bug that prevents use of `functools.partial`: https://github.com/lithops-cloud/lithops/issues/1428. This PR primarily provides a workaround for this.
- Additionally, for python <= 3.13, on linux, `ProcessPoolExecutor` defaults to mode "fork", which can be problematic. When calling `get_executor` with `ProcessPoolExecutor`, if the default mode is "fork", it uses "forkserver" instead, which is the default from python 3.14 onward.
- These fixes allowed removal of xfail pytest markers for ProcessPoolExecutor and lithops tests.
- Also fixed subtle bug in how we were passing arguments to lithops `FunctionExecutor.map`. This was not exposed because the erroneous way we were passing args works when the function passed to `map` expects only 1 argument, but when needing multiple arguments, the syntax was incorrect.

Acceptance criteria:

- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_parsers/test_dmrpp.py
+++ b/virtualizarr/tests/test_parsers/test_dmrpp.py
@@ -1,3 +1,4 @@
+import platform
 import textwrap
 from contextlib import nullcontext
 from pathlib import Path
@@ -486,6 +487,10 @@ def test_split_groups(hdf5_groups_file, group_path):
     assert result_tags == expected_tags
 
 
+@pytest.mark.xfail(
+    platform.system() == "Linux",
+    reason="Time values are improperly parsed, leading to pandas.errors.OutOfBoundsTimedelta errors",
+)
 @pytest.mark.parametrize(
     "group,warns",
     [
--- a/virtualizarr/tests/test_parsers/test_dmrpp.py
+++ b/virtualizarr/tests/test_parsers/test_dmrpp.py
@@ -489,7 +489,7 @@ def test_split_groups(hdf5_groups_file, group_path):
 
 @pytest.mark.xfail(
     platform.system() == "Linux",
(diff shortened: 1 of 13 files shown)


virtualizarr/tests/test_parsers/test_dmrpp.py
```diff
@@ -486,6 +487,10 @@ def test_split_groups(hdf5_groups_file, group_path):
     assert result_tags == expected_tags
 
 
+@pytest.mark.xfail(
+    platform.system() == "Linux",
+    reason="Time values are improperly parsed, leading to pandas.errors.OutOfBoundsTimedelta errors",
+)
```
TomNicholas, line 493: This is a real bug right? Surely we should raise an issue for this and fix it separately rather than xfailing?

TomNicholas (approved):
This is awesome thank you so much @chuckwondo !

virtualizarr/tests/test_parsers/test_dmrpp.py
chuckwondo: Yes, I believe this is a real bug, so I will open a new issue for this.

For this PR, I decided to mark this as xfail because I saw that this has been causing build failures elsewhere, but seemed to be ignored and merges to main were happening even with this failing. Also, this does _not_ fail on macOS, which is why I did not see the failure until it ran in CI. It passes on my MacBook.

TomNicholas: In that case I think xfailing and linking to the open github issue in the `reason` is appropriate.

PR #907, opened 2026-03-06
Update `kerchunk` writer `refs_to_dataframe` to deal with `ArrowStringArray` from new ZarrParser deps.

norlandrhagen:
# What I did
https://github.com/zarr-developers/VirtualiZarr/pull/892 added two new deps, `arro3-core` and `pyarrow` These broke the Kerchunk parquet tests:

`ValueError: Error converting column "path" to bytes using encoding UTF8. Original error: Unable to avoid copy while creating an array as requested.`

Adding a pandas context seemed to resolve this. `pd.option_context("future.infer_string", False)`

Acceptance criteria:

- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/virtualizarr/accessor.py
+++ b/virtualizarr/accessor.py
@@ -192,6 +192,7 @@ class _VirtualiZarrDatasetAccessor:
 
             return None
         elif format == "parquet":
+            import pandas as pd
             from kerchunk.df import refs_to_dataframe
 
             if isinstance(filepath, Path):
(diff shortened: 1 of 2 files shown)


TomNicholas (approved):
Nice catch!!

virtualizarr/accessor.py
```diff
@@ -192,21 +192,27 @@ def to_kerchunk(
 
             return None
         elif format == "parquet":
+            import pandas as pd
```
TomNicholas, line 195: Can we make pandas an explicit dependency of the kerchunk parquet writer? Currently we only get away with this because xarray is a required dependency which brings along pandas.

PR #909, opened 2026-03-09
Compile regular expressions

### LINKED ISSUE #901: Explore whether use of `re.compile` will reduce memory allocations
From a memray profile of https://gist.github.com/TomNicholas/9cc70049603f903c66c44e4963f39b20 `re` is allocating boatloads of memory (probably in extremely tiny increments). It's worth checking if using `re.compile` will reduce this.

```
🥇 Top 5 largest allocating locations (by size):
	- search:/Users/deepak/miniforge3/envs/icechunk/lib/python3.13/re/__init__.py:177 -> 25.421GB
	- match:/Users/deepak/miniforge3/envs/icechunk/lib/python3.13/re/__init__.py:167 -> 8.474GB
	- set_virtual_refs:/Users/deepak/repos/icechunk/icechunk-python/python/icechunk/store.py:323 -> 8.268GB
	- commit:/Users/deepak/repos/icechunk/icechunk-python/python/icechunk/session.py:400 -> 4.382GB
	- stack:/Users/deepak/miniforge3/envs/icechunk/lib/python3.13/site-packages/numpy/_core/shape_base.py:465 -> 2.194GB
	- ```

### PR DESCRIPTION
**Pull request by chuckwondo**
# What I did

Compile and cache regular expressions

Acceptance criteria:

- [x] Closes #901
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -12,7 +12,7 @@ from typing import Any, NewType, TypedDict, cast
 
 import numpy as np
 
-from virtualizarr.manifests.utils import construct_chunk_pattern, parse_manifest_index
+from virtualizarr.manifests.utils import compiled_chunk_pattern, parse_manifest_index
 from virtualizarr.types import ChunkKey
 
 # doesn't guarantee that writers actually handle these
@@ -27,9 +27,6 @@ VALID_URI_PREFIXES = {
     "http://",
     "https://",
 }
-
-
-_CHUNK_KEY = rf"^{construct_chunk_pattern('.')}"
 
 
 class ChunkEntry(TypedDict):
@@ -494,7 +491,7 @@ def validate_chunk_keys(chunk_keys: Iterable[ChunkKey]):
 
     # Check if all keys have the correct form
     for key in chunk_keys:
-        if not re.match(_CHUNK_KEY, key):
+        if not re.match(compiled_chunk_pattern("."), key):
             raise ValueError(f"Invalid format for chunk key: '{key}'")
 
     # Check if all keys have the same number of dimensions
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by TomNicholas — APPROVED** [review]
Thanks @chuckwondo !

File: virtualizarr/manifests/manifest.py [hunk 1]
```diff
@@ -494,7 +491,7 @@ def validate_chunk_keys(chunk_keys: Iterable[ChunkKey]):
 
     # Check if all keys have the correct form
     for key in chunk_keys:
-        if not re.match(_CHUNK_KEY, key):
+        if not re.match(compiled_chunk_pattern("."), key):
```
TomNicholas, line 494: This function can handle either possible separator, but this is the only place its used, and we hardcode separator here.

Do we actually need to be flexible over `separator`? I think it's an internal implementation detail of `ChunkManifest`...

chuckwondo: Good question. I was wondering about that, but don't have enough familiarity with the codebase to know what would be reasonable here. Any suggestion?

TomNicholas: In general we don't really know which separator was intended at construction time, as we don't currently have an argument which specifies it. So either:
- We add a separator argument argument to the `ChunkManifest` constructor
- We just compile both regexes, check against both, and raise if a key doesn't match either? 

Once the `ChunkManifest` object is constructed the choice of separator is effectively normalized away.

chuckwondo: I see. I started digging into it a bit, and it is perhaps best for a separate PR. Offhand, the required changes don't appear to be particularly large, but certainly seems like a separate issue.

I'm going to simply add some unit tests to address the code coverage warnings from above, and I'll open a separate issue for what you mention above.

TomNicholas: sounds good, feel free to merge this

PR #917, opened 2026-03-10
feat: configurable chunk separator

### LINKED ISSUE #911: Allow ChunkManifest to be constructed with either "." or "/" chunk key separator
Upon constructing a ChunkManifest, it is assumed that chunk keys are separated by `.`, and use of `/` in keys will fail validation. Using `/` should be supported. As mentioned in https://github.com/zarr-developers/VirtualiZarr/pull/909#discussion_r2907174899, we can either add a constructor parameter to allow specifying the separator explicitly, or simply attempt to validate against one separator, then against the second separator if validation against the first fails.

maxrjones:
# What I did

The chunk key encoding is an [extension point](https://zarr-specs.readthedocs.io/en/latest/v3/chunk-key-encodings/index.html), so this PR makes it configurable.
Acceptance criteria:

- [ ] Closes #911 
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -556,10 +556,10 @@ def validate_chunk_keys(
 ) -> None:
     if not chunk_keys:
         return
-
+    pattern = compiled_chunk_pattern(separator)
     # Check if all keys have the correct form
     for key in chunk_keys:
-        if not re.match(compiled_chunk_pattern(separator), key):
+        if not re.match(pattern, key):
             raise ValueError(f"Invalid format for chunk key: '{key}'")
 
     # Check if all keys have the same number of dimensions
--- a/virtualizarr/manifests/manifest.py
+++ b/virtualizarr/manifests/manifest.py
@@ -14,7 +14,11 @@ from typing import TYPE_CHECKING, Any, NewType, TypedDict, cast
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


TomNicholas (approved):
Nice! Let's add a release note for this one.

virtualizarr/manifests/manifest.py
```diff
@@ -534,39 +546,45 @@ def join(inds: Iterable[Any]) -> ChunkKey:
...
+def validate_chunk_keys(
+    chunk_keys: Iterable[ChunkKey], separator: ChunkKeySeparator = "."
+) -> None:
     if not chunk_keys:
         return
 
     # Check if all keys have the correct form
     for key in chunk_keys:
-        if not re.match(compiled_chunk_pattern("."), key):
+        if not re.match(compiled_chunk_pattern(separator), key):
```
chuckwondo: I don't know if this matters much, as I don't know how large `chunk_keys` might be, and `compiled_chunk_pattern` caches results, but it's still an unnecessary function call performed within the loop, so perhaps we can hoist it above the loop (I should have done so when I changed this code earlier), so something like this:

```python
pattern = compiled_chunk_pattern(separator)

for key in chunk_keys:
    if not re.match(pattern, key):
```

chuckwondo (approved):
Thanks for picking this up so quickly @maxrjones!

Looks great. I'm approving. I have only 1 minor suggestion, which you can take or leave.

PR #859, opened 2026-01-26
Define parser-specific Store protocols using obspec primitives

maxrjones:
Previously, VirtualiZarr imported ReadableStore from obspec_utils.protocols, which was overly broad relative to the methods used by parsers. This PR updates the parsers to directly define their required protocols using obspec.

As a consequence, obspec is added as a direct dependency (rather than via obspec_utils). I have also bumped the minimum obspec_utils version due to the ParallelStoreBug in v0.7.0 referenced in https://github.com/zarr-developers/VirtualiZarr/issues/858.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage
- [ ] Changes are documented in `docs/releases.rst`
- [ ] New functions/methods are listed in `api.rst`
- [ ] New functionality has documentation

PR #919, opened 2026-03-10
chore: add test for ZarrParser on bucket without list permissions

maxrjones:
# What I did

@TomNicholas mentioned some recent Zarr parser changes at a call today (#892), and I was curious about the implications for buckets without list permissions (e.g., HTTPS stores). This PR just adds a test that xfails.

As an aside, this is where https://github.com/zarr-developers/VirtualiZarr/pull/859 would come in handy because parsers could define their own required methods.

Acceptance criteria:

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_parsers/test_zarr.py
+++ b/virtualizarr/tests/test_parsers/test_zarr.py
@@ -20,7 +20,7 @@ from virtualizarr.parsers.zarr import (
     get_strategy,
     join_url,
 )
-from virtualizarr.tests import requires_pyarrow
+from virtualizarr.tests import requires_minio, requires_pyarrow
 
 pytestmark = requires_pyarrow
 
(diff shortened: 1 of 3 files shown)


virtualizarr/tests/test_parsers/test_zarr.py
```diff
@@ -545,3 +545,52 @@ def test_sharded_array_raises_error(tmpdir):
...
     ):
         parser(url=filepath, registry=registry)
+
+
+@requires_minio
+@pytest.mark.xfail(
+    reason="ZarrParser does not yet support buckets without list permissions"
+)
+def test_zarr_parser_nolist_bucket(minio_nolist_bucket):
+    """Test that ZarrParser works with a bucket that does not allow list operations."""
```
TomNicholas, line 555: I'm fine to merge this as-is, but I wonder if there is a simpler test that doesn't require minio.

The simplest example of a zarr url that we can't list is surely a http store? I was imagining there would be a class like that in zarr-python, but apparently there isn't. 

Obstore does have a HTTP store, but I'm confused why it says it supports list? 

https://developmentseed.org/obstore/latest/api/store/http/#obstore.store.HTTPStore.head_async

maxrjones: obstore's HTTPStore is designed around webdav (xref https://github.com/developmentseed/obspec-utils/issues/63).

It's a trade-off between code complexity with minio and relying on the network. For this version, you only need docker locally. I prefer local-first testing but am not planning to implement the fix so you could close this and build what works for you.

TomNicholas: I see. Sounds harder than I had anticipated, so minio seems fine.

PR #925, opened 2026-03-12
fix: add shutdown methods to executors

### LINKED ISSUE #926: Lithops FunctionExecutor memory leaks: atexit handler + unbounded futures list
## Summary

Lithops' `FunctionExecutor` has two memory leak sources that affect VirtualiZarr when using the lithops parallel backend. We've fixed both in #925 (the `executor-cleaning` branch), but documenting the upstream issues here for reference.

## Issue 1: Unbounded `futures` list (the main problem)

Every `map()` call on a `FunctionExecutor` appends `ResponseFuture` objects to `self.futures`. Each future caches its deserialized result in `_call_output`. Lithops **never clears this list internally**, so memory grows linearly with the number of operations — even when reusing a single executor.

This is the primary memory leak in typical VirtualiZarr usage (e.g., repeated `open_virtual_dataset` calls with a long-lived executor).

**Fix:** Clear `_call_output` on each future and clear the futures list during shutdown.

## Issue 2: `atexit` handler prevents garbage collection

In [`FunctionExecutor.__init__()`](https://github.com/lithops-cloud/lithops/blob/3.6.3/lithops/executors.py#L111), lithops registers:

```python
atexit.register(self.clean, clean_cloudobjects=False, clean_fn=True, on_exit=True)
```

`self.clean` is a bound method that holds a strong reference back to the executor. The `atexit` module keeps this reference alive for the entire process lifetime, so the `FunctionExecutor` can **never be garbage collected** — even after all user references are deleted.

This is by design in lithops: the atexit handler is a safety net to clean up cloud storage artifacts (cached functions, intermediate data) at process exit. It's harmless in their intended usage pattern (one executor per process), but causes a leak in VirtualiZarr's pattern of creating and discarding executors.

**Fix:** Call `atexit.unregister(self.lithops_client.clean)` during shutdown.

### MRE for the atexit issue

```python
import atexit
import gc
import weakref


class FakeFunctionExecutor:
    """Mimics lithops.executors.FunctionExecutor."""

    def __init__(self):
        # https://github.com/lithops-cloud/lithops/blob/3.6.3/lithops/executors.py#L111
        atexit.register(self.clean)

    def clean(self):
        pass


# --- Leak ---
executor = FakeFunctionExecutor()
ref = weakref.ref(executor)
del executor
gc.collect()
print(f"Object still alive: {ref() is not None}")  # True — leaked!

# --- Fix ---
executor2 = FakeFunctionExecutor()
ref2 = weakref.ref(executor2)
atexit.unregister(executor2.clean)
del executor2
gc.collect()
print(f"Object still alive: {ref2() is not None}")  # False — freed!
```

## Consideration for cloud executors

The `atexit.unregister` call removes lithops' safety net for cleaning cloud storage artifacts. For localhost executors this is harmless, but for cloud function executors (AWS Lambda, IBM Cloud Functions), if a process crashes without calling `shutdown()`, temporary cloud artifacts could be orphaned. A potential improvement would be to replace the strong atexit reference with a weakref-based callback:

```python
import weakref

def _weak_atexit_clean(weak_client):
    client = weak_client()
    if client is not None:
        client.clean(clean_cloudobjects=False, clean_fn=True, on_exit=True)

atexit.unregister(self.lithops_client.clean)
atexit.register(_weak_atexit_clean, weakref.ref(self.lithops_client))
```

This allows GC while still running cleanup if the executor is alive at exit.

## Upstream

There are no existing lithops issues about either of these problems. The only related issue ([lithops-cloud/lithops#1409](https://github.com/lithops-cloud/lithops/issues/1409)) was about thread growth, not memory leaks. We may want to file an upstream issue as well.

jbusecke:
@TomNicholas and I have been mulling over a complex native zarr ingestion job for a few days now. We were ingesting many batches of large (~1TB) native zarr stores, and saw a steady increase of memory which indicated that 'something' was holding onto memory in between batches. This PR adds tests to catch this behavior and a fix for the lithops executor that did fix our problem for now. 

- [x] The dataset we are using is currently not public. I would like to demonstrate the base issue fully reproducibly. If anyone knows a ~1TB/300k chunks native zarr store in an anon bucket, please let me know. -> Moved this concern to https://github.com/jbusecke/virtualizarr-benchmark

- [x] Closes #926 (confirmed in https://github.com/jbusecke/virtualizarr-benchmark)
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_parallel.py
+++ b/virtualizarr/tests/test_parallel.py
@@ -97,6 +97,6 @@ class TestExecutorShutdown:
 
 @requires_lithops
 def test_lithops_executor_data_cleaner_disabled():
-    """data_cleaner must be False to prevent atexit registration of lithops' clean method."""
+    """Data_cleaner must be False to prevent atexit registration of lithops' clean method."""
     with LithopsEagerFunctionExecutor() as executor:
         assert executor.lithops_client.data_cleaner is False


virtualizarr/parallel.py
```diff
@@ -372,4 +373,18 @@ def shutdown(self, wait: bool = True, *, cancel_futures: bool = False) -> None:
         wait
             Whether to wait for pending futures.
         """
+        # Free cached results from lithops ResponseFuture objects before shutdown.
```
TomNicholas: where are the shutdown methods for all the other executors? I swear I saw them on your PC earlier !

virtualizarr/tests/test_parallel.py
```diff
@@ -41,3 +48,99 @@ def test_get_executor_process_pool_mode():
 
     assert ctx is not None, "Expected executor to have a multiprocessing context"
     assert ctx.get_start_method() == "forkserver"
+
+
+@requires_lithops
+class TestLithopsExecutorShutdown:
```
TomNicholas: I would like to see a test like this but parametrized over all executors, checking that all of them clear up whatever type of futures they hold.

virtualizarr/parallel.py
chuckwondo: I removed the ones that only had a `pass` statement.

TomNicholas: No I meant Julius (who is physically in the same office as me) showed me this code on his laptop earlier and it showed him adding `.shutdown` methods for all the executors that actually did more than just `pass`.

TomNicholas: IIUC all these custom executors should theoretically have shutdown methods.

virtualizarr/tests/test_parallel.py
```diff
@@ -41,3 +48,99 @@ def test_get_executor_process_pool_mode():
...
+
+ALL_EXECUTORS = [
+    _make_executor(SerialExecutor),
+    _make_executor(DaskDelayedExecutor),
+    _make_executor(LithopsEagerFunctionExecutor),
+]
+
+
+@pytest.mark.parametrize("executor_cls", ALL_EXECUTORS)
+class TestExecutorMemory:
```
TomNicholas: I'm not sure if either of these tests will be reliable enough - curious of @chuckwondo 's thoughts.

virtualizarr/parallel.py
jbusecke: I initially remove them since they were not failing https://github.com/zarr-developers/VirtualiZarr/blob/d925fb7fed0b7659f39313908cc7c5d7dd6865f6/virtualizarr/tests/test_parallel.py#L99 (while lithops **did** fail). But I added them back in now.

virtualizarr/parallel.py
```diff
@@ -372,4 +379,18 @@ def shutdown(self, wait: bool = True, *, cancel_futures: bool = False) -> None:
...
+        # in _call_output. Without this, memory accumulates across repeated calls.
+        for f in self.lithops_client.futures:
+            f._call_output = None
+        self.lithops_client.futures.clear()
+        self._futures.clear()
+
+        # Lithops registers self.clean as an atexit handler (executors.py __init__),
+        # which prevents the FunctionExecutor from ever being garbage collected.
+        # Unregister it so the executor can be freed after shutdown.
+        atexit.unregister(self.lithops_client.clean)
```
TomNicholas: This is absolutely wild and deserves raising upstream

virtualizarr/parallel.py
chuckwondo: I think we might want to add this at the top of `shutdown`:

```python
if wait:
    self.lithops_client.wait(show_progressbar=False)
```

That will do quite a bit of additional cleanup, via calling `clean` on itself.

virtualizarr/tests/test_parallel.py
```diff
@@ -41,3 +47,88 @@ def test_get_executor_process_pool_mode():
...
+        "LithopsEagerFunctionExecutor": [requires_lithops],
+    }
+    return pytest.param(
+        executor_cls,
+        id=executor_cls.__name__,
+        marks=marks.get(executor_cls.__name__, []),
+    )
+
+
+ALL_EXECUTORS = [
```
TomNicholas: This deliberately excludes `concurrent.futures.ThreadPoolExecutor` and `concurrent.futures.ProcessPoolExecutor`

```suggestion
ALL_CUSTOM_EXECUTORS = [
```

virtualizarr/tests/test_parallel.py
```diff
@@ -41,3 +47,88 @@ def test_get_executor_process_pool_mode():
...
+            assert len(executor.lithops_client.futures) == 0
+            assert all(f._call_output is None for f in lithops_futures)
+
+        # Testing idempotency
+        executor.shutdown()
+        assert len(executor._futures) == 0
+
+
+@pytest.mark.parametrize("executor_cls", ALL_EXECUTORS)
+class TestExecutorMemory:
```
chuckwondo: I'm not sure these really belong here. They seem like tests that should occur upstream. If we see memory leaks in the upstream executors, we should probably be opening bugs against the appropriate repositories, no?

virtualizarr/parallel.py
chuckwondo: Probably so.

TomNicholas: see https://github.com/zarr-developers/VirtualiZarr/issues/926

virtualizarr/tests/test_parallel.py
TomNicholas: I agree, see https://github.com/zarr-developers/VirtualiZarr/issues/926 for discussion of what we should do or not do to clean up

virtualizarr/parallel.py
TomNicholas: That sounds like a good idea. Does `lithops_client.wait` internally call `lithops_client.clean`?

chuckwondo: > That sounds like a good idea. Does `lithops_client.wait` internally call `lithops_client.clean`?

Yes

jbusecke: Thanks @chuckwondo, testing that now.

jbusecke:
Ok Tom and I actually worked on an alternative approach where we change the lithops config to set `lithops.data_cleaner` to false (this is true by default and triggers the [atexit registration](https://github.com/lithops-cloud/lithops/blob/5becdfaeb8b786c445b3762575f8118a1994060e/lithops/executors.py#L111)). Combined with the added `.shutdown()` method on the Lithops exec this solves the problem in #926 and seems a bunch nicer than the original approach. 

I have limited this to when the backend is `localhost` so that we leave the serverless behavior untouched for now. We could easily extend this if a user finds this error with other backends.

virtualizarr/tests/test_parallel.py
jbusecke: I think you are right in principal, but I would propose to keep this around as at least an optional test due to the significant work that was needed to get to the bottom of this.

TomNicholas: Yeah @chuckwondo I see these tests as hopefully-temporary, but unfortunately important.

virtualizarr/parallel.py
jbusecke: I added the wait now, but alone it did not fix the problem.

virtualizarr/parallel.py
```diff
@@ -230,6 +233,9 @@ def map(
         # Compute all tasks
         return iter(dask.compute(*delayed_tasks))
 
+    def shutdown(self, wait: bool = True, *, cancel_futures: bool = False) -> None:
+        self._futures.clear()
```
chuckwondo, line 237: Can we just get rid of `self._futures`? It's not clear why we even hold the list of futures to begin with?

jbusecke: Um that is actually a good question @chuckwondo. @TomNicholas do you know why this was necessary in the case of dask?

chuckwondo: Why is this the case for _any_ of the executors?

PR #927, opened 2026-03-13
Remove pyarrow dependency from ZarrParser

### LINKED ISSUE #922: Update ZarrParser to remove pyarrow dependency.
From some feedback from @kylebarron:

```
I'd strongly suggest not tying to pyarrow

    You should very easily be able to make your code generic and not tied to pyarrow
    pyarrow doesn't have any internal type checking, so typing as pa.StringArray or pa.Uint64Array means absolutely nothing to the user (it might mean something to the developer)
&&

Feel free to disregard, but ideally you shouldn't need to depend on both dependencies. Pyarrow is very large. I looked recently and it looks like it's gotten even larger


50MB compressed is huge.
```

TomNicholas:
### What I did

Started as an attempt to close #922, ended up rewriting the entire parser. The "strategy" thing was way overcomplicated, it's much more sane and functional now.

### Acceptance criteria:

- [x] Closes #922
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage
- [ ] ~~New functions/methods are listed in an appropriate `*.md` file under `docs/api`~~


virtualizarr/parsers/zarr.py
```diff
@@ -518,6 +52,54 @@ def _run_async(coro: Coroutine[Any, Any, T]) -> T:
...
+    @property
+    def metadata_key_names(self) -> tuple[str]:
+        match self:
+            case ZarrFormat.V2:
+                return (".zarray", ".zattrs", ".zgroup", ".zmetadata")
+            case ZarrFormat.V3:
+                return ("zarr.json",)
+
+    @property
+    def scalar_chunk_key_name(self) -> str:
```
norlandrhagen, line 78: nice

virtualizarr/parsers/zarr.py
```diff
@@ -518,6 +52,54 @@ def _run_async(coro: Coroutine[Any, Any, T]) -> T:
...
+    @property
+    def chunks_dir_prefix(self) -> str:
+        match self:
+            case ZarrFormat.V2:
+                return ""
+            case ZarrFormat.V3:
+                return "c/"
+
+
+def join_url(base: str, key: str) -> str:
```
norlandrhagen, line 94: Does it make sense to use cloudpathlib or upathlib or something instead of r/l strips?

virtualizarr/parsers/zarr.py
```diff
@@ -619,21 +201,317 @@ def __call__(
...
+async def construct_manifest_array(
+    zarr_array: zarr.AsyncArray[Any], path: str
+) -> ManifestArray:
+    """Construct a ManifestArray from a zarr array."""
+    array_v3_metadata = metadata_as_v3(zarr_array.metadata)
+
+    # This is the only restriction on what Zarr Arrays cannot be virtualized
+    if any(isinstance(codec, ShardingCodec) for codec in array_v3_metadata.codecs):
+        raise NotImplementedError(
+            f"Zarr V3 arrays with sharding are not yet supported, but array {path} uses the ShardingCodec."
```
norlandrhagen, line 274: Out of scope of this PR. But what would be needed to get sharded zarr working?

virtualizarr/parsers/zarr.py
```diff
@@ -619,21 +201,317 @@ def __call__(
...
+            f"uses {type(array_v3_metadata.chunk_grid).__name__}."
+        )
+
+    # The on-disk format determines how chunks are stored (e.g. V2 has no c/ prefix),
+    # which differs from the always-V3 metadata we use internally.
+    on_disk_zarr_format = ZarrFormat(zarr_array.metadata.zarr_format)
+    on_disk_separator = (
+        zarr_array.metadata.chunk_key_encoding.separator
+        if on_disk_zarr_format == ZarrFormat.V3
+        else "."
```
norlandrhagen, line 292: Should this use some method from ZarrFormat.V2 instead of '.'?

virtualizarr/parsers/zarr.py
```diff
@@ -619,21 +201,317 @@ def __call__(
...
+def metadata_as_v3(metadata: ArrayV3Metadata | ArrayV2Metadata) -> ArrayV3Metadata:
+    """Convert metadata to V3 format with normalized chunk_key_encoding."""
+
+    if isinstance(metadata, ArrayV2Metadata):
+        v3_dict = _convert_v2_to_v3_dict(metadata)
+    else:
+        v3_dict = metadata.to_dict()
+
+    # Normalize chunk_key_encoding to DefaultChunkKeyEncoding with "." separator.
+    # The ManifestStore expects dot-separated keys (e.g. "0.0.0"), so we enforce
```
norlandrhagen, line 317: Probably out of scope as well, but is it worth updating the ManifestStore keys from dot to slash separated keys at some point to more closely mirror Zarr V3?

virtualizarr/parsers/zarr.py
```diff
@@ -619,21 +201,317 @@ def __call__(
...
+    """
+
+    chunk_grid_shape = determine_chunk_grid_shape(
+        metadata.shape, metadata.chunk_grid.chunk_shape
+    )
+    total_size = math.prod(chunk_grid_shape)
+
+    # Handle scalar arrays
+    if metadata.shape == ():
+        # Can only contain a single chunk, so just GET that instead of LISTing a whole directory unnecessarily
```
norlandrhagen, line 403: Nice!

virtualizarr/parsers/zarr.py
```diff
@@ -619,21 +201,317 @@ def __call__(
...
+
+    # join batches into one 1D array for all initialized chunks
+    all_paths = np.concatenate(path_batches)
+    all_sizes = np.concatenate(size_batches)
+
+    # strip the prefix to get chunk keys like "0.0.0"
+    stripped_keys = np.strings.slice(all_paths, len(array_chunks_prefix), None)
+
+    # construct full URIs for each chunk
+    full_paths = np.strings.add(store_base_uri + "/", all_paths)
```
norlandrhagen, line 524: Similar to the pathlib comment, but is it worth using some tool like that here to be potentially less brittle with string concatenating?

norlandrhagen (approved):
Great rewrite!

virtualizarr/parsers/zarr.py
TomNicholas: parsing the shard index I think

virtualizarr/parsers/zarr.py
TomNicholas: Yeah I thought about adding this logic to the `ZarrFormat` enum but it would have to be a method that accepts `metadata` which didn't feel much cleaner.

virtualizarr/parsers/zarr.py
TomNicholas: Interesting question ...VirtualiZarr's choice is arbitrary, and the default in V3 is still dot, so that seems like the safer option?

virtualizarr/parsers/zarr.py
TomNicholas: Ideally zarr-python would do all this for us, and provide some type of class for that, but I don't trust it. And we are dealing with a very well-defined set of possibilities here (i.e. what the zarr spec says about the on-disk format). Also we only use this function like twice. 🤷

virtualizarr/parsers/zarr.py
TomNicholas: Problem is with that approach then by virtualizing we are kind of undoing the sharding. And anyone using sharding probably knew what they were doing...

PR #938, opened 2026-03-20
feat: generalize ChunkManifest to hold inline chunks

### LINKED ISSUE #851: Generalize `ChunkManifest` to hold native chunks as well as virtual refs
Generalizing the `ChunkManifest` class to hold native chunks as well as virtual refs would unlock several features.

It's needed for:
1. Writing an `IcechunkParser` (https://github.com/earth-mover/icechunk/issues/104)
2. Concatenating virtual and non-virtual data in memory (Icechunk already supports doing this on-disk via `append_dim`)
3. `ManifestStore.to_icechunk()/kerchunk()`, and thereby making xarray an optional dependency (#521)

The problem is that the current implementation of `ChunkManifest` uses a clever trick: it's just 3 numpy arrays in a trenchcoat. This gives us loads of stuff for free:
- Efficient contiguous in-memory representation
- Efficient handling of Variable-length strings (via the numpy 2 dtype)
- Efficient functions for iterating over every element
- Efficient multi-dimensional concat/stack functions for merging chunk manifests
- No unusual or non-python dependencies

But I don't know how to keep that design and also store non-virtual chunks in arbitrary locations within those arrays.

Some alternatives:
- Numpy object array (almost certainly very inefficient)- 
- Keep the 3 numpy arrays, but have an auxiliary buffer in which to store non-virtual chunks in memory (proposed on #794 by @maxrjones)
- Write our own in-memory chunk manifest class in rust

maxrjones:
This implements the design proposed in https://github.com/zarr-developers/VirtualiZarr/issues/851#issuecomment-3975263592 for supporting native chunks in ChunkManifest, which allows loading Kerchunk JSONs with in-lined references or eventually Icechunk stores.

I prefer this design over https://github.com/zarr-developers/VirtualiZarr/pull/794, but still don't really like the amount of `if native` forks needed. Still, this is such a long-overdue feature that it seems worth releasing soon even if it's not the perfect design. We can also improve the internals later on. Let's discuss at the dev meeting tomorrow. 

- Adds `_native: dict[tuple[int, ...], bytes]` to `ChunkManifest` for storing in-memory chunk data
- Extends `ChunkEntry` with optional `data: NotRequired[bytes]` field
- Native chunks propagate through concat/stack/broadcast, pickle, and `ManifestStore.get()`
- Kerchunk parser decodes base64 inlined refs into native chunks
- Writers serialize native chunks: base64 for kerchunk, `store.set()` for icechunk
- Docs page + tests including a kerchunk→icechunk→xarray roundtrip

Acceptance criteria:

- [ ] Closes #851
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


docs/inlined_references.md
TomNicholas, line 1: Can this go under the `Explanation` section of the docs instead? Most users won't need to know about this, it's basically an implementation detail.

virtualizarr/manifests/manifest.py
```diff
@@ -39,10 +39,17 @@ class ChunkEntry(TypedDict):
...
 
     @classmethod  # type: ignore[misc]
     def with_validation(
-        cls, *, path: str, offset: int, length: int, fs_root: str | None = None
+        cls,
+        *,
+        path: str,
+        offset: int,
+        length: int,
+        data: bytes | None = None,
```
TomNicholas: More informative name?
```suggestion
        native_chunk_buffer: bytes | None = None,
```

virtualizarr/manifests/manifest.py
```diff
@@ -190,6 +201,7 @@ class ChunkManifest:
     _paths: np.ndarray[Any, np.dtypes.StringDType]
     _offsets: np.ndarray[Any, np.dtype[np.uint64]]
     _lengths: np.ndarray[Any, np.dtype[np.uint64]]
+    _native: dict[tuple[int, ...], bytes]
```
TomNicholas: How is a scalar native array handled? An empty tuple? Worth having a test for that case.

virtualizarr/manifests/manifest.py
```diff
@@ -213,6 +225,11 @@ def __init__(
                 "0.1.1": {"path": "s3://bucket/foo.nc", "offset": 400, "length": 100},
             }
             ```
+
+            Entries may also include a ``data`` key with raw bytes for native (in-memory) chunks::
+
+                {"0.0": {"path": "", "offset": 0, "length": 4, "data": b"\\x00\\x01\\x02\\x03"}}
```
TomNicholas, line 219: So empty path `""` means both "chunk is missing" and "chunk is native" - find it in the buffer? I think we should disambiguate those.

virtualizarr/manifests/manifest.py
```diff
@@ -234,26 +251,39 @@ def __init__(
         offsets = np.empty(shape=shape, dtype=np.dtype("uint64"))
         lengths = np.empty(shape=shape, dtype=np.dtype("uint64"))
 
+        native: dict[tuple[int, ...], bytes] = {}
+
         # populate the arrays
         for key, entry in entries.items():
-            if not isinstance(entry, dict) or len(entry) != 3:
+            if not isinstance(entry, dict) or len(entry) not in (3, 4):
```
TomNicholas: What is this? Seems like some very unreliable form of input validation?

TomNicholas: Should we make the `entries` into `TypedDicts`?

TomNicholas (changes requested):
This is awesome, thanks @maxrjones .

My main concern about this PR is that it changes private internals and public behaviour in one go. The more neat way to do it would be to leave any changes to the Kerchunk and Icechunk readers/writers to follow-up PRs.

virtualizarr/manifests/manifest.py
maxrjones: I improved this in [`e4ebc28` (this PR)](https://github.com/zarr-developers/VirtualiZarr/pull/938/commits/e4ebc28377ab3ed2ae088a72c071f5544c7688c6). I think the change to TypedDict would be worthwhile, but could be separate from this PR

virtualizarr/manifests/manifest.py
maxrjones: I disambiguated those in [`04a420f` (this PR)](https://github.com/zarr-developers/VirtualiZarr/pull/938/commits/04a420f21b1868dedbae6afb92989f4fa6309dc3)

virtualizarr/manifests/manifest.py
maxrjones: added a test in [`8fc3de3` (this PR)]([`8fc3de3` (this PR)](https://github.com/zarr-developers/VirtualiZarr/pull/938/commits/8fc3de361e244329b3358abdcc7ba58b3627a5f6), and yes - https://github.com/maxrjones/VirtualiZarr/blob/8fc3de361e244329b3358abdcc7ba58b3627a5f6/virtualizarr/tests/test_manifests/test_manifest.py#L520-L521

virtualizarr/manifests/manifest.py
maxrjones: I renamed to `inline_data` in [`8516604` (this PR)](https://github.com/zarr-developers/VirtualiZarr/pull/938/commits/85166043f64c70871f78f5796a62ad9aaf5ef74d)

TomNicholas:
> Good point, I pulled out the Kerchunk reader portion. I think the writers need to know what to do with inlined data immediately, otherwise you could get buggy serialized data if someone were to write a parser that uses the inlined data feature.

I mean you could just update the writers to raise `NotImplementedError` if there are any native chunks, but the splitting you've already done is good, thanks you.

> Parsers chose which chunks are inlined. 

Cool. So it is really an implementation detail. It's only relevant for Parser authors. So doesn't that mean the docs for it should only be under "Writing a custom parser"?

TomNicholas:
@maxrjones I hope you don't mind but I took over this PR.

I responded to my own review by making these changes:
  - Reverted all Icechunk writer changes — kept strictly to the `ChunkManifest` internals                                                                          
  - Docs moved out of a standalone page into `data_structures.md` under `## Chunk Manifests`                                                                         
  - Tests added for `concat` / `stack` / `broadcast` of manifests with inlined chunks, plus `ManifestStore` reads (byte-range variants, mixed inlined/virtual, `list_dir`) 
  - Fixed a broadcast bug: `_broadcast_manifest` wasn't replicating inlined entries along expanded axes — a `(1,2) → (3,2)` broadcast would leave the replicated positions missing from `_inlined` even though `_paths` claimed `__inlined__` there. Now replicates by reference, no byte copies.                                     
  - Fixed an equality bug: `elementwise_eq` didn't compare inlined bytes, which broke an invariant in `ManifestArray.__eq__` and tripped `RuntimeWarning("Should not be possible to get here")` for two manifests with identical paths/offsets/lengths but different inlined bytes.

maxrjones:
> @maxrjones I hope you don't mind but I took over this PR.
> 
> I responded to my own review by making these changes:
> 
> * Reverted all Icechunk writer changes — kept strictly to the `ChunkManifest` internals
> * Docs moved out of a standalone page into `data_structures.md` under `## Chunk Manifests`
> * Tests added for `concat` / `stack` / `broadcast` of manifests with inlined chunks, plus `ManifestStore` reads (byte-range variants, mixed inlined/virtual, `list_dir`)
> * Fixed a broadcast bug: `_broadcast_manifest` wasn't replicating inlined entries along expanded axes — a `(1,2) → (3,2)` broadcast would leave the replicated positions missing from `_inlined` even though `_paths` claimed `__inlined__` there. Now replicates by reference, no byte copies.
> * Fixed an equality bug: `elementwise_eq` didn't compare inlined bytes, which broke an invariant in `ManifestArray.__eq__` and tripped `RuntimeWarning("Should not be possible to get here")` for two manifests with identical paths/offsets/lengths but different inlined bytes.

this is great, thank you very much for taking this on! Those changes are all very helpful. I haven't done a line-by-line review, but would be happy with you merging whenever you feel it's ready.

Issue #896, opened 2026-02-27
Deduplicate paths in ChunkManifest


maxrjones:
## Problem

`ChunkManifest` stores chunk file paths in an `NDArray[StringDType]` — one full path string per chunk. When many chunks reference the same source file (the common case), the path is duplicated for every chunk.

For a file with 1000 chunks pointing to `s3://my-bucket/data/ERA5/temperature_2024_01.nc` (~55 chars), the `_paths` array stores that string 1000 times. With `_offsets` (8 bytes/chunk) and `_lengths` (8 bytes/chunk), the per-chunk cost is roughly:

- **Current**: ~55 (StringDType) + 8 + 8 = **~71 bytes/chunk**
- **Proposed**: 4 (uint32 path_id) + 8 + 8 = **~20 bytes/chunk** + one copy of the path string

At scale:
| Chunks | Current | Proposed | Savings |
|--------|---------|----------|---------|
| 10K | 710 KB | 200 KB | ~500 KB |
| 1M | 71 MB | 20 MB | ~51 MB |
| 50M | 3.5 GB | 1 GB | ~2.5 GB |

## Proposal

Replace `_paths: NDArray[StringDType]` with two fields:

```python
_path_ids: NDArray[np.uint32]  # chunk_grid_shape, index into _paths
_paths: list[str]              # deduplicated path lookup table
```

### Missing chunks

Use sentinel `path_id = 0xFFFFFFFF` (uint32 max, define as `MISSING_CHUNK` constant). The `_paths` list only contains real paths.

### Construction

**`__init__(entries, shape)`**: Build path dedup during construction. Collect unique paths, assign IDs, populate `_path_ids`.

**`from_arrays(paths, offsets, lengths)`**: Accept full-path StringDType array (existing API), deduplicate internally:
```python
unique_paths, inverse = np.unique(paths, return_inverse=True)
self._path_ids = inverse.reshape(paths.shape).astype(np.uint32)
self._paths = unique_paths.tolist()
```

### Concat (lazy dedup)

When concatenating manifests, offset-append path tables rather than deduplicating:

```python
new_paths = left._paths + right._paths
new_path_ids = np.concatenate([
    left._path_ids,
    right._path_ids + np.uint32(len(left._paths))
], axis=axis)
```

This is O(1) for the path table (just list concat) and avoids scanning for duplicates during the hot concat path. After many concats, duplicates may accumulate in `_paths`.

[...]

Issue #246, opened 2024-10-03
Make Manifests' in-memory reference structure pluggable


TomNicholas:
Currently the `ChunkManifest` is hardcoded to use numpy arrays underneath to store the paths/offsets/byte ranges. However there are a few cases where we might want to use another format:

- sparse grids of chunks
- "algorithmically inflatable" grids of chunks (see #238)
- using even more efficient reference storage formats (https://github.com/zarr-developers/VirtualiZarr/issues/23)

We could definitely imagine making this pluggable. The main question I have is whether the other manifest implementations should implement their own `ChunkManifest` class (e.g. `SparseChunkManifest`), i.e. `ManifestArray` becomes a `Generic` in the type of the `.manifest` attribute; or use virtualizarr's `ChunkManifest` class but wrap a different array type, i.e. `ChunkManifest` becomes a `Generic` in the type of the `.paths`/`.offset`/`.lengths` attributes.

Right now the latter should be pretty straightforward, but the former would require some refactoring because the `ChunkManifest` abstraction is leaky in that the [implementation of `concatenate`](https://github.com/zarr-developers/VirtualiZarr/blob/47a5e8702e44f71bb355bcba0ff6214fe6d09d83/virtualizarr/manifests/array_api.py#L88) for `ManifestArrays` [accesses private internals](https://github.com/zarr-developers/VirtualiZarr/blob/47a5e8702e44f71bb355bcba0ff6214fe6d09d83/virtualizarr/manifests/array_api.py#L129) of the wrapped `ChunkManifest` (i.e. the wrapped numpy arrays).

A related consideration is providing some kind of interface for iterating over all the references in the Manifest that doesn't make assumptions about how the references are actually stored under the hood. That's currently another place where the abstraction is a bit leaky, e.g. 

https://github.com/zarr-developers/VirtualiZarr/blob/47a5e8702e44f71bb355bcba0ff6214fe6d09d83/virtualizarr/manifests/manifest.py#L356

PR #939, opened 2026-03-20
refactor: add iteration helpers to ChunkManifest

maxrjones:
This PR fixes the leaky abstraction noted in https://github.com/zarr-developers/VirtualiZarr/issues/246, whereby the implementation of concatenate for ManifestArrays accesses private internals of the wrapped ChunkManifest. While not essential, I think it's worthwhile because it's a simple change that keeps subsequent PRs from making the codebase even more leaky:

- #851
- #896 (if we chose to do it)
- #860

For example, I think the changes in https://github.com/maxrjones/VirtualiZarr/compare/encapsulate-manifest-internals...maxrjones:VirtualiZarr:native-chunks-on-encapsulated?expand=1 are preferable to those in https://github.com/zarr-developers/VirtualiZarr/pull/938 because there are fewer cross-module accesses of private attributes.

Changes:

- Move concat/stack/broadcast operations into `ChunkManifest` classmethods
- Add `get_entry()`, `elementwise_eq()`, `iter_nonempty_paths()`, and `iter_refs()` methods
- Remove all direct access to `_paths`/`_offsets`/`_lengths` from outside `manifest.py`

Acceptance criteria:

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/array_api.py
+++ b/virtualizarr/manifests/array_api.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Callable, Union, cast
+from typing import TYPE_CHECKING, Callable, Union
 
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


TomNicholas (commented):
I like the iteration helpers a lot but I don't really see why `concat` should be a classmethod. It's basically just a function.

I understand you want to avoid ever calling private `ChunkManifest` internals outside of the `.manifest.py` file, but the file is somewhat arbitrary, and it's nice to define the data structure in one place and the operations on the data structure in other places.

I think it makes more sense to think in terms of public and private API than in terms of files. It's fine for a set of public functions to use private API, and that's what we're already doing here: `ChunkManifest` and `concat` are both listed as public API

https://virtualizarr.readthedocs.io/en/stable/api/developer.html#array-api

virtualizarr/manifests/manifest.py
```diff
@@ -435,6 +435,47 @@ def __eq__(self, other: Any) -> bool:
...
+        length = self._lengths[indices]
+        return ChunkEntry(path=path, offset=offset, length=length)
+
+    def elementwise_eq(self, other: "ChunkManifest") -> np.ndarray:
+        """Return boolean array where True means that chunk entry matches."""
+        return (
+            (self._paths == other._paths)
+            & (self._offsets == other._offsets)
+            & (self._lengths == other._lengths)
+        )
```
TomNicholas, line 453: This will have to be updated in #938, to compare the values of inlined chunks.

virtualizarr/manifests/array_api.py
TomNicholas, line 1: The changes in this file seem a bit pointless now?

maxrjones: I prefer functions for large blocks of code that share a goal, but am fine removing the changes

TomNicholas: I mean sure, happy to accept this. Just pointing out that the changes are now very minimal.

TomNicholas (approved):

virtualizarr/manifests/array_api.py
maxrjones: yeah fair, I wouldn't have opened a PR for just this scope, lesson learned for next time

PR #952, opened 2026-03-30
fix: preserve shard shape when writing to icechunk

TomNicholas:
## Summary
- Fixes the icechunk writer using `metadata.chunks` (inner chunk shape) instead of `metadata.chunk_grid.chunk_shape` (shard shape) when creating arrays, causing sharded arrays to lose their shard configuration
- Adds a regression test that creates a sharded zarr store, virtualizes it, writes to icechunk, and verifies both shard and chunk shapes are preserved

Closes #951

## Test plan
- [x] New test `test_sharded_array_roundtrip_icechunk` passes
- [x] Existing icechunk writer tests unaffected (pre-existing `test_write_loadable_variable` failure is unrelated)

### DIFF
--- a/virtualizarr/writers/icechunk.py
+++ b/virtualizarr/writers/icechunk.py
@@ -581,14 +581,6 @@ def write_manifest_virtual_refs(
         for grid_index, entry in manifest.iter_refs()
     ]
 
-    print(f"  write_manifest_virtual_refs: arr_name={arr_name}, num_specs={len(virtual_chunk_spec_list)}")
-    if arr_name == "embeddings":
-        target = [0, 0, 48, 97]
-        match = [s for s in virtual_chunk_spec_list if s.index == target]
-        print(f"  chunk [0,0,48,97]: found={len(match)}")
-        if match:
-            print(f"  location={match[0].location!r}, offset={match[0].offset}, length={match[0].length}")
-
     store.set_virtual_refs(
         array_path=key_prefix,
         chunks=virtual_chunk_spec_list,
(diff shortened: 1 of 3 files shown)


virtualizarr/writers/icechunk.py
```diff
@@ -518,7 +518,7 @@ def write_virtual_variable_to_icechunk(
         arr = group.require_array(
             name=name,
             shape=metadata.shape,
-            chunks=metadata.chunks,
+            chunks=metadata.chunk_grid.chunk_shape,
```
maxrjones: @TomNicholas the current change will lose the inner chunk structure. I cannot drop in-line suggestion to L517, but I think this code block should look like:
```python
        filters, serializer, compressors = extract_codecs(metadata.inner_codecs)
        arr = group.require_array(
            name=name,
            shape=metadata.shape,
            chunks=metadata.chunks,
            shards=metadata.shards,
            dtype=metadata.data_type.to_native_dtype(),
            filters=filters,
            compressors=compressors,
            serializer=serializer,
            dimension_names=var.dims,
            fill_value=metadata.fill_value,
        )
```

- metadata.chunks / metadata.shards tells require_array both the inner chunk shape and shard shape, so it constructs the ShardingCodec correctly
- metadata.inner_codecs — extracts the codecs inside the sharding codec (e.g. BytesCodec, ZstdCodec)

maxrjones (changes requested):

virtualizarr/writers/icechunk.py
TomNicholas: oh interesting - thanks. Apparently this seemed to work even without this change, due to some subtlety in zarr-python? I/Claude couldn't find a way to make a test that failed before adding your suggestion but passed after.

But I'm made this change now anyway as it's much more explicit.

maxrjones (approved):

Issue #12, opened 2024-03-08
Support variable-length chunks


TomNicholas:
Requires implementation upstream first

TomNicholas:
See also https://github.com/earth-mover/icechunk/issues/386

TomNicholas:
This is now implemented upstream in Zarr-Python, so the to-do list is:
- [x] Support in Zarr-Python (done in https://github.com/zarr-developers/zarr-python/pull/3802(
- [ ] Support in `ManifestArray` (see https://github.com/zarr-developers/VirtualiZarr/pull/954)
- [ ] Support region-writing to Icechunk from VirtualiZarr
- [ ] Support reading back data in Xarray (see https://github.com/pydata/xarray/pull/11592)
- [ ] Support in parsers that could potentially return rectilinear chunks
  - [ ] `ZarrParser`
  - [ ] `ZippedZarrParser`
  - [ ] `IcechunkParser`

I don't know of any other formats that support variable-length chunks within a single file - HDF5 and TIFF both don't.

TomNicholas:
This is a big problem that needs to be addressed: https://github.com/zarr-developers/zarr-extensions/issues/74

PR #954, opened 2026-04-02
Add support for rectilinear chunk grids

maxrjones:
This is a companion PR to https://github.com/zarr-developers/zarr-python/pull/3802 that adds support for rectilinear chunk grids.

Acceptance criteria:

- [x] Closes part of #12
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


norlandrhagen:
Super excited for this! Just ran into a use-case today that it'll be great for.

maxrjones:
> Super excited for this! Just ran into a use-case today that it'll be great for.

nice! I'll try to get back to this tomorrow. I got a bit behind on development due to travel. 

Are you able to share anything about the use-case? I'm supposed to demo this feature at our project wrap-up next week and if it's related to NASA data it could be a good collaboration opportunity.

norlandrhagen:
> Are you able to share anything about the use-case? I'm supposed to demo this feature at our project wrap-up next week and if it's related to NASA data it could be a good collaboration opportunity.

Yeah happy to share a gist or w/e! It's just some subset of CMIP data, so I'm guessing not super great for a demo? But if it's helpful for debugging / real world testing of rectilinear chunk grids, can definitely share it.

sjperkins:
I've started investigating a VirtualiZarr implementation after getting motivated here:

- https://github.com/earth-mover/icechunk/issues/2303

A rectilinear chunk grid would be very helpful for some RLE-encoded data in a legacy Radio Astronomy format.

TomNicholas:
I've taken over this PR - thanks for starting it @maxrjones !

It now works locally - CI is failing because it requires Xarray support for reading back the rectilinear chunks. That means we should really release Xarray with https://github.com/pydata/xarray/pull/11592 before we release VirtualiZarr with this feature.

One non-trivial design decision I've made here is when to actually use rectilinear chunk grids. Generally we probably don't want to just use rectilinear chunk grids by default, since they are often not needed, the metadata scaling is worse, and fewer tools understand the result. But we also don't want concatenation of two different-length chunks to silently change the chunk grid type. So what I've chosen to do for now is only ever use rectilinear chunk grids if the zarr experimental config flag is switched on, and otherwise error as before.

Note there are some other VZ to-do's in #12, but I will stack those on top of this PR.

PR #957, opened 2026-04-08
Add compatibility with Zarr-Python 3.2.0

maxrjones:
There are some internal changes coming in Zarr-Python that impact VirtualiZarr, which are addressed by this PR.

Acceptance criteria:

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/manifests/array.py
+++ b/virtualizarr/manifests/array.py
@@ -3,7 +3,12 @@ from typing import Any, Callable, Union, cast
 
 import numpy as np
 import xarray as xr
-from zarr.core.metadata.v3 import ArrayV3Metadata, RegularChunkGrid
+from zarr.core.metadata.v3 import ArrayV3Metadata
+
+try:
+    from zarr.core.metadata.v3 import RegularChunkGridMetadata
+except ImportError:
+    from zarr.core.metadata.v3 import RegularChunkGrid as RegularChunkGridMetadata
 
 import virtualizarr.manifests.utils as utils
 from virtualizarr.manifests.array_api import (
@@ -50,7 +55,7 @@ class ManifestArray:
             # try unpacking the dict
             _metadata = ArrayV3Metadata(**metadata)
 
-        if not isinstance(_metadata.chunk_grid, RegularChunkGrid):
+        if not isinstance(_metadata.chunk_grid, RegularChunkGridMetadata):
             raise NotImplementedError(
                 f"Only RegularChunkGrid is currently supported for chunk size, but got type {type(_metadata.chunk_grid)}"
             )
--- a/virtualizarr/manifests/array.py
+++ b/virtualizarr/manifests/array.py
@@ -1,16 +1,9 @@
 import warnings
-from typing import Any, Callable, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, Union, cast
 
 import numpy as np
 import xarray as xr
 from zarr.core.metadata.v3 import ArrayV3Metadata
-
-try:
-    from zarr.core.metadata.v3 import RegularChunkGridMetadata  # zarr-python>3.1.6
-except ImportError:
-    from zarr.core.metadata.v3 import (
-        RegularChunkGrid as RegularChunkGridMetadata,  # zarr-python<=3.1.6
-    )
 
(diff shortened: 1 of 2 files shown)


norlandrhagen (approved):

virtualizarr/manifests/array.py
```diff
@@ -3,7 +3,14 @@
...
 import xarray as xr
-from zarr.core.metadata.v3 import ArrayV3Metadata, RegularChunkGrid
+from zarr.core.metadata.v3 import ArrayV3Metadata
+
+try:
+    from zarr.core.metadata.v3 import RegularChunkGridMetadata  # zarr-python>3.1.6
+except ImportError:
+    from zarr.core.metadata.v3 import (
+        RegularChunkGrid as RegularChunkGridMetadata,  # zarr-python<=3.1.6
+    )
```
TomNicholas, line 13: Maybe we should actually look for specific versions of zarr-python instead of the blind try...except?

TomNicholas (approved):
Released, so as long as this passes (particularly on `min-deps`!) then lets go

virtualizarr/manifests/array.py
maxrjones: Zarr-Python uses hatch-vcs, which builds versions from git tags. If a user installs from a git source that does not have the tags fetched, the version can display as a low number despite being based on the most recent commits to main, which breaks the code with an obscure error message. This try...except leads to informative error messages by describing exactly what is needed rather than implicitly relying on the version.

TomNicholas:
Running the upstream tests shows a breaking change - it seems that zarr now forbids chunks of length 0, which it didn't forbid before?

https://github.com/zarr-developers/VirtualiZarr/actions/runs/24685336933/job/72193091799

AFAICT the Zarr spec does not forbid length-0 chunks, so it's zarr-python that is wrong here?

https://zarr-specs.readthedocs.io/en/latest/v3/chunk-grids/regular-grid/index.html

maxrjones:
> Running the upstream tests shows a breaking change - it seems that zarr now forbids chunks of length 0, which it didn't forbid before?
> 
> [zarr-developers/VirtualiZarr/actions/runs/24685336933/job/72193091799](https://github.com/zarr-developers/VirtualiZarr/actions/runs/24685336933/job/72193091799)
> 
> AFAICT the Zarr spec does not forbid length-0 chunks, so it's zarr-python that is wrong here?
> 
> [zarr-specs.readthedocs.io/en/latest/v3/chunk-grids/regular-grid/index.html](https://zarr-specs.readthedocs.io/en/latest/v3/chunk-grids/regular-grid/index.html)

This was discussed in https://github.com/zarr-developers/zarr-python/pull/3802#discussion_r3030161543. We leaned towards this being a mistake in the core spec since modular arithmetic is undefined for 0. Please raise an upstream issue if you still think this should be addressed in zarr-python rather than the spec.

LDeakin:
More context on that
- https://github.com/zarrs/zarrs/issues/352
- https://github.com/zarr-developers/zarr-python/issues/3711

TomNicholas:
Thanks for the context guys.

@sharkinsspatial you wrote this test and the [fixture](https://github.com/zarr-developers/VirtualiZarr/blame/7a98c011307a8d40bcfda3e67aec71210fb777a9/virtualizarr/tests/test_parsers/conftest.py#L33) it uses - do you have any concerns with HDF compatibility if I simply change the behaviour of the `HDFParser` to raise on length-0 chunks?

TomNicholas:
As it seems that zarr does allow 0-length shape, https://github.com/zarr-developers/VirtualiZarr/pull/977 just has the `HDFParser` set the chunk lengths to be 1 instead of 0 if not found, and that seems to be enough to handle empty HDF5 datasets. I'll merge that then update this branch to unblock us - @sharkinsspatial please shout if you think that's the wrong way to handle this edge case.

TomNicholas:
That seemed to work, but then revealed another breaking change in zarr-python: https://github.com/zarr-developers/zarr-python/issues/3922.

EDIT: Fixed by https://github.com/zarr-developers/zarr-python/pull/3924

PR #967, opened 2026-04-10
perf: use set_virtual_refs_arr for faster icechunk writes

TomNicholas:
## Summary

- Replaces the Python loop that creates per-chunk `VirtualChunkSpec` objects with a single call to `store.set_virtual_refs_arr()`, passing the manifest's numpy arrays directly to Rust
- ~2.8x faster for writing virtual refs against object storage (12.4s → 4.5s for 6 vars × 520k chunks each against Arraylake)

Depends on earth-mover/icechunk#2049.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -17,7 +17,7 @@
 ### Internal changes
 
 - Use `set_virtual_refs_arr` for ~3x faster virtual ref writing to icechunk.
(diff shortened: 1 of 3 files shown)


TomNicholas:
Failures are because I didn't think about the empty paths case 😞 

Requires https://github.com/earth-mover/icechunk/pull/2080 to be merged and released in IC

PR #969, opened 2026-04-15
build!: drop Python 3.11 support, require Python >= 3.12

TomNicholas:
## Summary

- Drop Python 3.11 support, require Python >= 3.12
- icechunk 2.x requires Python 3.12+, so VirtualiZarr must match
- Removes py311 pixi feature, test-py311 CI environment
- Updates minimum-versions env to use py312

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


TomNicholas:
@maxrjones I want this to support Icechunk 2.0.2, so that I can use the new API in #967.

maxrjones (approved):

PR #988, opened 2026-05-04
Correctly handle HDF5 fillvalue for string dtype arrays.

### LINKED ISSUE #878: Failure to read HDF5 data with no _FillValue attribute
Hello,
I am exploring VirtualiZarr/IceChunk for our workflow. I followed VirtualiZarr's example of reading one of our NetCDF files into VirtualiZarr:

```
import obstore
import virtualizarr as vz

bucket = 's3://its-live-data'
granule = "velocity_image_pair/landsatOLI/v02/S80W170/LC08_L1GT_020121_20231013_20231102_02_T2_X_LC09_L1GT_020121_20231106_20231106_02_T2_G0120V02_P084.nc"

store = obstore.store.from_url(bucket, region="us-west-2", skip_signature=True)
registry = ObjectStoreRegistry({bucket: store})
parser = HDFParser()

vds = vz.open_virtual_dataset(
    url=f"{bucket}/{granule}",
    parser=parser,
    registry=registry,
    loadable_variables=[],
    # drop_variables=['img_pair_info', 'mapping']
)
```

and getting the following failure:
```
virtualizarr/parsers/hdf/hdf.py:82
fill_value = dataset.fillvalue.item()
AttributeError: 'bytes' object has no attribute 'item'
```

It appears to be linked to the data variables that don't have` _FillValue `attribute specified. I tried to exclude such two variables (please see commented out  `drop_variables` in the code above), but that does not seem to make any difference.

I am using 
obstore 0.8.2
virtualizarr 2.4.0

Is it a bug? Any advice would be greatly appreciated.

Thanks,
Masha

sharkinsspatial:
# What I did
Updated the HDF5 parser to support string dtype arrays with a `bytes` or `str` `fillvalue`. 

Acceptance criteria:

- [x] Closes #878
- [x] Tests added
- [x] Tests passing


virtualizarr/parsers/hdf/hdf.py
```diff
@@ -64,6 +80,11 @@ def _construct_manifest_array(
     attrs = _extract_attrs(dataset)
     dtype = dataset.dtype
 
+    # HDF5 variable-length strings use numpy object dtype, which zarr v3 cannot
+    # resolve automatically. Map to StringDType which zarr maps to VariableLengthUTF8.
+    if dtype.kind == "O":
+        dtype = np.dtypes.StringDType()
```
maxrjones:
```suggestion
    # Discriminate against other object-kind HDF5 dtypes (vlen arrays, object/
    # region references) that would silently be coerced to StringDType and
    # produce garbage downstream — those aren't supported yet, so fail loudly.
    if h5py.check_string_dtype(dtype) is not None:
        dtype = np.dtypes.StringDType()
    elif dtype.kind == "O":
        raise NotImplementedError(
            f"HDF5 object dtype {dtype!r} is not a variable-length string and "
            f"is not yet supported by HDFParser. h5py exposes vlen arrays "
            f"(`h5py.vlen_dtype`) and object/region references "
            f"(`h5py.ref_dtype`, `h5py.regionref_dtype`) as numpy object dtype; "
            f"please open an issue if your file needs one of these."
        )
```
I don't think all hdf5 object-kind dtypes are variable-length strings

TomNicholas (approved):

TomNicholas:
Good to merge now so I can include it in the release?

maxrjones:
> Good to merge now so I can include it in the release?

let me take a quick look at the codecs changes. Assigning vlen bytes to all object dtypes seems a bit weird to me.

virtualizarr/codecs.py
```diff
@@ -106,7 +106,9 @@ def convert_to_codec_pipeline(
     arrayarray_codecs, arraybytes_codec, bytesbytes_codecs = extract_codecs(zarr_codecs)
 
     if arraybytes_codec is None:
-        if dtype.byteorder == ">":
+        if isinstance(dtype, np.dtypes.StringDType) or dtype.kind == "O":
```
maxrjones:
```suggestion
        if isinstance(dtype, np.dtypes.StringDType):
```
We should not apply the `VLenUTF8Codec` to any object dtype, since that could also reference vlen-bytes, vlen-array, pickle, json2, or msgpack2. Parsers should just use `np.dtypes.StringDType` directly. The HDF5 parser already coerces vlen strings to StringDType (and raises for other object kinds), so this was both unneeded and spec-incompliant.

maxrjones: would be good to get @TomNicholas, @sharkinsspatial, or @abarciauskas-bgse to double-check/apply this suggestion, then this PR should be good to go.

virtualizarr/parsers/hdf/hdf.py
```diff
@@ -35,6 +35,25 @@
...
 
+def _get_fill_value(dataset: H5Dataset):
+    """
+    Extract the fill value from an h5py dataset, handling string/bytes dtypes
+    that don't return numpy scalars from dataset.fillvalue.
+    """
+    try:
+        raw = dataset.fillvalue
+    except RuntimeError:
+        return np.ma.default_fill_value(dataset.dtype)
```
abarciauskas-bgse, line 46: I'm wondering if this could be introducing a different type of bug where now we are introducing a value for the zarr metadata `fill_value` that could differ from what is in the dataset's original metadata. I realized today that for GPM IMERG HH precipitation the `attrs` from the h5py Dataset include both `_FillValue` and `CodeMissingValue` which both are `-9999.0`:

```python
h5ds = h5py.File(s3fs.open(gpm_imerg_s3_url))
h5ds['/Grid/precipitation'].attrs['CodeMissingValue']
# np.bytes_(b'-9999.9')
h5ds['/Grid/precipitation'].attrs['_FillValue']
# np.float32(-9999.9)
```

Whereas this fallback returns `1e+20`:

```
np.ma.default_fill_value(h5ds['/Grid/precipitation'].dtype)
# 1e+20
```

I recognize that `CodeMissingValue` is not a standard attribute (as far as I can tell), but should the zarr `fill_value` attribute fall back to the `_FillValue` attribute picked up by h5py?

Looking at https://github.com/NASA-IMPACT/veda-odd/issues/371 would this be handled by the more general purpose phase 3 "`fill_value_defined()` distinction: stop propagating h5py-default fills to zarr storage."

abarciauskas-bgse: @maxrjones curious what you think

virtualizarr/codecs.py
abarciauskas-bgse: Your logic seems sound but I am not an expert. Is the non-compliance aspect simply that Zarr doesn't support the object data type?

virtualizarr/parsers/hdf/hdf.py
sharkinsspatial: @abarciauskas-bgse For clarification, `fill_value` and `_FillValue` are distinct semantic concepts https://virtualizarr.readthedocs.io/en/stable/custom_parsers.html#fill-values. `fill_value` is always set by `HDF5` but as you noted above, not all types of `fill_value` that `HDF5` can produce are supported by Zarr v3.

abarciauskas-bgse: Right ok, thanks for reminding me about that distinction -- I recalled there were 2 semantic concepts but forgot about the details and that excellent documentation. In that case, since this function is for the first fill value concept -- `Value for uninitialized chunks - (e.g., Zarr fill_value)` -- then probably this fallback is reasonable?

And I recognize that the `h5ds['/Grid/precipitation'].attrs['_FillValue']` is for the 2nd concept `Sentinel value - (e.g., CF _FillValue ))`

maxrjones: The `CodeMissingValue` is a different issue, but this is a good flag since Zarr's default fill values could differ from numpy or support dtypes not defined in `np.ma.default_fill_value`

@sharkinsspatial why did you chose to use numpy's default fill values here?

abarciauskas-bgse: That was a change I made, to be clear, and in discussion with @maxrjones it sounds like there may be a Zarr native fallback to use to look into

sharkinsspatial: @maxrjones @abarciauskas-bgse I'm just returning to this now (I mistakenly thought this had been altered and merged).  I just did some investigation and found that 
- Very old versions of h5py would throw on `fillvalue` for `vlen` strings and object types but this was fixed by https://github.com/h5py/h5py/pull/2132. 
- I ran tests with Claude for a wide range of `dtypes` and was unable to force an exception.
- It is theoretically possible for a non-HDF5 library (Matlab, netcdf-4) to create a non-compliant HDF dataset that has no `fillvalue` but I think the likelihood of encountering this is very low.
- As @maxrjones note it might be best for us to avoid injecting our own default `fillvalue` here in favor of just erroring on this actually exceptional case.

If you're ok I will revert the addition from your PR.

abarciauskas-bgse: FWIW I encountered this issue (where the fillvalue attribute was absent on an HDF5 dataset, causing the runtime error) while working on the virtual icechunk store for the GPM IMERG HH dataset (see https://github.com/virtual-zarr/gpmimerghh-virtualizarr-data-pipeline/blob/main/lambda/virtualizarr-processor/pyproject.toml#L10). 

I don't know about the likelihood of encountering other HDF datasets with _no_ fillvalue attribute. But just to be clear I wasn't trying to handle cases 1 and 2 from the bullets above.

It's fine with me if you need to revert the change, but I am curious what the path is for HDF datasets which have no fillvalue attribute.

sharkinsspatial: Only HDF datasets that are variable length arrays, object references or region references will return `None` for `fillvalue`.  In `h5py` these are all represented as the numpy kind `O` with no way to distinguish between them.  @maxrjones 's suggestion added a check first for string dtypes (because unfortunately vlen strings are also represented as `O`) and the throws `NotImplementedError` for any other type of dataset.  

Theoretically we can expose `vlen_dtype` and `ref_dtype` through v3's `VariableLengthBytes` but this will likely be pretty complex and we can defer for a separate PR.

abarciauskas-bgse: I _think_ what I encountered was different, where the fillvalue attribute was missing entirely, i.e.:

```python
# when fillvalue is `None`
hasattr(dataset, 'fillvalue') => True
# when fillvalue is missing
hasattr(dataset, 'fillvalue') => False
```

So one workaround could be to collapse both scenarios to None:

```python
raw = getattr(dataset, 'fillvalue', None)
```

however I realize the missing attribute scenario may not be likely or the purview of this PR (and like you both mentioned we may not want to inject the numpy default in this case) so fine with me if you want to revert this change.

sharkinsspatial: @abarciauskas-bgse In this case we are unconcerned with the arbitrary `attrs` and only concerned with the HDF 5 dataset's `fillvalue` property.  In the `attrs` case `fillvalue` is semantically meaningless as the correct CF convention `attr` for sentinel masking would be `_FillValue` https://virtualizarr.readthedocs.io/en/stable/explanation/custom_parsers.html#fill-values   The parser has separate logic for handling and encoding the `_FillValue` `attr` so that it can be used by xarray and other downstream libs for sentinel masking https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/parsers/hdf/hdf.py#L74-L76

sharkinsspatial:
@TomNicholas This is ready for final review.

PR #991, opened 2026-05-12
feat(parsers): add IcechunkParser for converting icechunk sessions to ManifestStores

### LINKED ISSUE #817: Open icechunk with manifests array
We would need an IcechunkParser to modify things on xarray-level and store it again with references only (no data).

Best and thanks,
Fabi

TomNicholas:
Closes #817 and https://github.com/earth-mover/icechunk/issues/104

## Summary

Adds `virtualizarr.parsers.IcechunkParser`, which converts a live icechunk `Session`'s chunk references into a VirtualiZarr `ManifestStore`. The parser walks zarr groups via the standard zarr-python API to enumerate arrays and recover their `zarr.json` metadata, then for each array async-iterates a new columnar accessor on the icechunk store (`array_chunk_iterator`) and scatters batches into the dense numpy arrays VirtualiZarr's `ChunkManifest.from_arrays` expects.

## Mapping

- IC virtual ref (`s3://`, `gs://`, `vcc://` — all resolved by icechunk) → VZ virtual ref, URL preserved.
- IC native (managed) chunk → VZ virtual ref with path `f\"{native_chunks_prefix}/{chunk_id}\"`. The prefix is supplied at parser construction time; icechunk returns bare chunk ids.
- IC inline chunk → VZ inline chunk (bytes carried in `ChunkManifest._inlined`).

## Usage

```python
import icechunk
from virtualizarr.parsers import IcechunkParser

repo = icechunk.Repository.open(storage=...)
session = repo.readonly_session(branch=\"main\")

parser = IcechunkParser(native_chunks_prefix=\"s3://my-bucket/my-repo/chunks\")
manifest_store = parser(session.store, registry=ObjectStoreRegistry({}))
```

## Design notes

The parser is a thin scatter on top of the icechunk-side iterator. All path rendering (native prefix concatenation) and VZ-specific sentinels (`INLINED_CHUNK_PATH`) are applied parser-side, so the icechunk API surface stays free of VirtualiZarr vocabulary.

Memory characteristics measured on a 10M-chunk fixture (single 1-D array, virtual refs to a fake S3 URL, disk-backed icechunk):
- ~0.92 μs/chunk parse time (~1.08 M refs/s on an M2 MBA)
- ~64 B/chunk of Python-side memory (uint64 offsets + uint64 lengths + StringDType path slot)

A 3-D cube fixture of similar chunk count benched at ~0.96 μs/chunk and ~65 B/chunk — within ~3% of the 1-D case.

## Requires

[earth-mover/icechunk#2124](https://github.com/earth-mover/icechunk/pull/2124) — adds `IcechunkStore.array_chunk_iterator`, which this parser consumes. Cannot be merged before that.

## Test plan

- [x] 5 new tests in `virtualizarr/tests/test_parsers/test_icechunk.py`:
  - end-to-end: returns a ManifestStore
  - manifest values for inline + virtual + missing slots
  - native-chunk prefix application (and trailing-slash tolerance)
  - `skip_variables` filter
- [x] Other parser test suites unaffected (no regressions in `test_zarr.py`).
- [x] Update docs once the icechunk PR lands and ships a release.

## Followups (not in this PR)

- Sharded array handling (likely just works — IC's chunk grid maps to inner zarr chunks — but unverified).
- Checksum preservation: IC virtual refs may carry ETag / LastModified checksums; VZ has no slot for these so they're dropped silently.
- Nested subgroup walking: matches `ZarrParser`'s current single-group-level behavior.


virtualizarr/parsers/icechunk.py
```diff
@@ -0,0 +1,365 @@
...
+        cfg = store.config or {}
+        return icechunk.s3_storage(
+            bucket=cfg["bucket"],
+            prefix=full_prefix or None,
+            region=cfg.get("region"),
+            endpoint_url=cfg.get("endpoint"),
+            access_key_id=cfg.get("access_key_id"),
+            secret_access_key=cfg.get("secret_access_key"),
+            session_token=cfg.get("session_token"),
+            anonymous=cfg.get("skip_signature", False) or None,
```
tylanderson: In store.config, skip_signature is stored as a string, which breaks here

virtualizarr/writers/icechunk.py
```diff
@@ -9,7 +9,7 @@
 from xarray.backends.zarr import encode_zarr_attr_value
 from zarr import Array, Group
 from zarr.core.buffer import default_buffer_prototype
-from zarr.core.chunk_key_encodings import ChunkKeyEncoding
+from zarr.core.chunk_key_encodings import DefaultChunkKeyEncoding
```
TomNicholas, line 12: This is a drive-by bugfix that ensures that we use a chunk key separator of `/` when setting IC chunks

d-v-b: drive-by comment / question: should these be public API if yall are using them?

TomNicholas: Maybe? This one in particular is super unimportant though, it saves me like 3 lines of code and probably isn't worth it...

PR #997, opened 2026-05-18
ci: decouple upstream dev-version testing from the pixi workspace

### LINKED ISSUE #995: Docs CI on Read the Docs fails intermittently with SIGSEGV during `pixi install --environment docs`
The Read the Docs build is failing intermittently. Looking at the last 15 builds for `virtualizarr`:

- 5 succeeded
- 5 failed with `SIGSEGV` during a wheel build in the pixi solve phase
- 4 failed with other `upstream`-environment solve errors
- 1 failed in the `minimum-versions` solve

The SIGSEGV is reported against *different* build backends across runs — `hatchling.build.build_wheel` (obspec-utils) in [build 32741525](https://app.readthedocs.org/projects/virtualizarr/builds/32741525/), `mesonpy.build_wheel` in [build 32696583](https://app.readthedocs.org/projects/virtualizarr/builds/32696583/), and others with no specific package named. So the package being built isn't the cause — it's whichever subprocess is unlucky.

**Likely cause.** `pixi install --environment docs` triggers a full workspace lockfile solve. The solve covers all configured platforms (`linux-64`, `osx-64`, `osx-arm64`) and includes the `upstream` dependency group, which contains 11 `git+https://…` URLs (xarray, zarr, numcodecs, ujson, icechunk, astropy, kerchunk, s3fs, universal_pathlib, obspec_utils, virtual_tiff). For each git source pixi/uv runs an isolated wheel build to extract metadata, several in parallel. The simultaneous native-code builds (Rust for icechunk, C/Cython for numcodecs/astropy/ujson) plausibly push the RTD container past its memory cap; cgroup-OOM under those conditions can surface as SIGSEGV with a core dump rather than SIGKILL.

Standalone reproduction tests against obspec-utils all pass:
- `python -m build --wheel` on the local checkout
- `uv pip install "obspec_utils @ git+https://github.com/virtual-zarr/obspec-utils"`
- `pixi install --environment docs` locally (pixi 0.66.0, osx-arm64)

RTD uses pixi 0.68.1, which may also be a factor.

**Workaround.** Rerun. ~1-in-3 odds it passes.

**Possible longer-term fixes, in order of effort:**

1. Trim the `upstream` group — `obspec_utils` and `virtual_tiff` already have PyPI releases pinned in the main deps, and the `upstream` group is primarily intended to catch breakage from *external* dev versions (xarray, zarr, numcodecs, etc.). Dropping the `virtual-zarr/*` git entries here would reduce the parallel-build count without losing meaningful coverage.
2. Reduce the `upstream` group more aggressively — keep only the truly external bleeding-edge deps (xarray, zarr, numcodecs).
3. Bump the RTD plan or move docs to a GitHub Actions runner with more memory.
4. Investigate whether pixi 0.68.x changed git-source build isolation behavior versus 0.66.x.

TomNicholas:
Closes #995.

Move the 11 `git+https://…` dev-version sources out of `pyproject.toml`'s `[dependency-groups].upstream` into a standalone `ci/upstream-overrides.txt`. The Upstream CI job now pip-overlays them on top of a regular `upstream` pixi env via `pip install --upgrade --no-deps -r ci/upstream-overrides.txt`.

## Why

Pixi solves a single workspace-wide lockfile covering every environment defined in `[tool.pixi.environments]`. With `git+https` sources in the `upstream` dependency group, every solve — `docs`, `minimum-versions`, even `test` — had to build a wheel for each git source as part of metadata extraction. On memory-constrained CI runners (Read the Docs in particular), the parallel native-code builds (Rust for icechunk, C/Cython for numcodecs/astropy/ujson, hatchling for the rest) intermittently SIGSEGV'd or hit `ETXTBSY` races, blocking docs and other unrelated environments. See #995 for the data.

Keeping git sources out of the pixi workspace means every other env's solve becomes trivial (PyPI/conda-forge only). The `upstream` job still tests against bleeding-edge dev versions — just via a post-install pip overlay rather than during solve.

## Trade-offs

- The pixi solver no longer sees the dev versions of those 11 packages, so it can't catch a transitive incompatibility (e.g. dev xarray needing a newer numpy than pixi picked). That would surface as an import-time error in the upstream job instead. Acceptable for a canary job.
- `pip` had to be added to the upstream env (new `upstream-overlay` feature) since pixi envs don't include it by default.

## Test plan

- [x] `pixi install --environment docs` succeeds locally (was the broken case)
- [x] `pixi install --environment upstream` succeeds locally with the new layout
- [x] `pixi run -e upstream pip install --no-deps --dry-run "obspec_utils @ git+…"` resolves and would install
- [ ] Read the Docs build for this branch passes (will know after push)
- [ ] Upstream CI job runs to completion on a labeled PR (the workflow's trigger requires the `test-upstream` label or a scheduled run)

### DIFF
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -38,9 +38,17 @@ jobs:
           pixi-version: v0.59.0
           environments: upstream
 
-      - name: List installed libraries
+      - name: Install pixi upstream environment
         run: |
           pixi install --environment upstream
+
+      - name: Overlay dev versions of upstream libraries
+        # See ci/upstream-overrides.txt for why these aren't in pyproject.toml.
+        run: |
+          pixi run -e upstream pip install --upgrade --no-deps -r ci/upstream-overrides.txt
+
+      - name: List installed libraries
+        run: |
           pixi list --environment upstream
 
       - name: Cache tutorial data
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Overlay dev versions of upstream libraries
         # See ci/upstream-overrides.txt for why these aren't in pyproject.toml.
         run: |
-          pixi run -e upstream pip install --upgrade --no-deps -r ci/upstream-overrides.txt
+          pixi run -e upstream install-upstream-overrides
 
       - name: List installed libraries
         run: |
--- a/ci/upstream-overrides.txt
+++ b/ci/upstream-overrides.txt
@@ -1,0 +1,21 @@
+# Dev versions of upstream libraries used by the `Upstream` CI job to test
+# VirtualiZarr against bleeding-edge releases of its dependencies.
+#
+# Installed via `pip install --upgrade --no-deps -r ci/upstream-overrides.txt`
+# on top of an already-resolved pixi `upstream` env, so pip only swaps the
+# named packages and does not re-resolve the dependency graph.
+#
+# These deliberately do NOT live in pyproject.toml: keeping git+https sources
+# out of the pixi workspace prevents every other env's lockfile solve (docs,
+# minimum-versions, etc.) from having to build them. See issue #995.
+xarray @ git+https://github.com/pydata/xarray
+universal_pathlib @ git+https://github.com/fsspec/universal_pathlib
+numcodecs @ git+https://github.com/zarr-developers/numcodecs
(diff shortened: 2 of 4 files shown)


maxrjones (approved):
I'd probably add a task rather than needing to type `pip install --upgrade --no-deps -r ci/upstream-overrides.txt` after `pixi run -e upstream` but that's just a nit. Seems worth trying this

PR #1013, opened 2026-06-10
ci: exclude GOES-16 ingest notebook from lint hooks

TomNicholas:
Follow-up to #1012. The `examples/V2/goes-16-ingest.ipynb` notebook was ported verbatim from the [GOES-16 blog post](https://www.earthmover.io/blog/virtual-zarr) and trips the repo's lint hooks (F404/F821 in narrative cells, formatting, and a couple of codespell hits in prose). Rather than rewrite the author's notebook, exclude it from linting:

- **`pyproject.toml`** — add the notebook to `[tool.ruff] exclude` (covers both `ruff` lint and `ruff format`), and set `force-exclude = true` so the exclude is honoured when pre-commit passes the file explicitly.
- **`.pre-commit-config.yaml`** — add the notebook to codespell's `-S` skip list.

`pre-commit run --files examples/V2/goes-16-ingest.ipynb pyproject.toml .pre-commit-config.yaml` now passes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
     rev: v2.4.2
     hooks:
       - id: codespell
-        args: ["-L", "fo,ihs,kake,te", "-S", "fixture"]
+        args: ["-L", "fo,ihs,kake,te", "-S", "fixture,*goes-16-ingest.ipynb"]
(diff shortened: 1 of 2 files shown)


TomNicholas:
(I shouldn't have merged the other PR before CI completed 🤦 )

PR #1014, opened 2026-06-16
Revert warning about ProcessPoolExecutor

jbusecke:
Reverting #890 as it has been fixed by #899


jbusecke:
Can also tone down the lithops section about non-threadsafe parsers if desired?

TomNicholas (approved):
I think this is fine, but pinging @chuckwondo to double-check.

chuckwondo (approved):
LGTM!

PR #1015, opened 2026-06-16
fix: raise a clear error when concatenating along a scalar coord

maxrjones:
@hrodmn hit the obscure missing chunk manager error today, this time because xarray tried to load data when generating indexes for concatenating along a dimension that has scalar coordinates. This PR raises a clear error with informative guidance (load the variable).

This is a distinct from existing chunk manager issues so it doesn't close anything, but is related to #114.

Acceptance criteria:

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/tests/test_xarray.py
+++ b/virtualizarr/tests/test_xarray.py
@@ -1009,6 +1009,30 @@ class TestOpenVirtualMFDataset:
         )
         assert combined_vds.sizes["time"] == 2
 
+    def test_nested_concat_dim_none_merges_without_raising(
+        self, netcdf4_file, local_registry
+    ):
+        # combine="nested" with concat_dim=None is the merge-only path: datasets are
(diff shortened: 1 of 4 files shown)


TomNicholas (changes requested):
Thanks, let's de-claude the tests before merging.

virtualizarr/tests/test_xarray.py
```diff
@@ -968,6 +969,118 @@ def test_parallel_open(
             )
             xrt.assert_identical(combined_vds, expected_vds)
 
+    def test_nested_concat_dim_scalar_virtual_coord_raises(
```
TomNicholas, line 972: I think we should delete all the other tests from this PR except the first part of this one. Every other test here either tests a private internal, or tests a behaviour that should already be tested.

Issue #382, opened 2025-01-22
Concatenation of virtual datasets fails due to missing Chunk Manager


observingClouds:
**What I did**
I created kerchunk reference files following #381 and tried to combine them via `xr.merge`.

```python
import xarray as xr
import tempfile
import json
from pathlib import Path
import virtualizarr as vz
import os

#Create xarray dataset
ds1 = xr.Dataset(
    {
        "a": (("x", "y"), [[1, 2], [3, 4]]),
        "b": (("x", "y"), [[10, 20], [30, 40]]),
    },
    coords={"x": [10, 20], "y": [1, 2]},
)

ds2 = xr.Dataset(
    {
        "c": (("x", "y"), [[5, 6], [7, 8]]),
        "d": (("x", "y"), [[50, 60], [70, 80]]),
    },
    coords={"x": [1, 2], "y": [1, 2]},
)

ref1 = ds1.virtualize.to_kerchunk()
ref2 = ds2.virtualize.to_kerchunk()

## 'Outlining' all references as inlined data is not yet supported
"""NotImplementedError: Outlining all references as inlined data is not yet supported. [ToDo]"""
tempdir1 = Path(tempfile.TemporaryDirectory().name)
tempdir2 = Path(tempfile.TemporaryDirectory().name)
def outline_references(ref: dict, folder: Path = None) -> dict:
    """
    Virtualizarr currently does not support inlined references.
    To open references with virtualizarr, the references must be written to a file.
    Except the .zarray, .zattrs and .zgroup files, all references are written to disk.
    """
    refs = ref["refs"]
    for k, v in refs.items():
        if os.path.basename(k).startswith('.'):
            continue
        elif isinstance(v, str):
            file = folder / k
            if not os.path.exists(os.path.dirname(file)):
                os.makedirs(os.path.dirname(file))
            with open(folder / k, "w") as f:
                f.write(v)
            refs[k] = [str(file), 0, v.__sizeof__()]
    return ref

ref1 = outline_references(ref1, tempdir1)
ref2 = outline_references(ref2, tempdir2)


## Write references to disk (open_virtual_dataset expects a string)
with open("ref1.json", "w") as f:
    json.dump(ref1, f)

with open("ref2.json", "w") as f:
    json.dump(ref2, f)

# Note this section requires the modification in #381 
vds1 = vz.open_virtual_dataset("ref1.json", filetype='kerchunk')
[...]

Issue #354, opened 2024-12-17
Xarray assertions can trigger loading


TomNicholas:
This should not happen, instead either attempting to load should trigger a clear `NotImplementedError`, or perhaps in this case xarray's assert functions should dispatch to virtualizarr's equals method instead of trying to coerce to a numpy array.

```python
    def test_lithops(self, netcdf4_files_factory):
        # by default this will use the lithops LocalHost executor
    
        filepath1, filepath2 = netcdf4_files_factory()
    
        # test combine nested without in-memory indexes
        combined_vds = open_virtual_mfdataset(
            [filepath1, filepath2],
            combine="nested",
            concat_dim="time",
            coords="minimal",
            compat="override",
            indexes={},
            parallel="lithops",
        )
        vds1 = open_virtual_dataset(filepath1, indexes={})
        vds2 = open_virtual_dataset(filepath2, indexes={})
        expected_vds = xr.concat(
            [vds1, vds2], dim="time", coords="minimal", compat="override"
        )
    
>       xrt.assert_identical(combined_vds.coords['lat'], expected_vds.coords['lat'])

virtualizarr/tests/test_backend.py:508: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Users/tom/miniconda3/envs/numpy2.0_released/lib/python3.11/site-packages/xarray/core/formatting.py:973: in diff_array_repr
    temp = [wrap_indent(short_array_repr(obj), start="    ") for obj in (a, b)]
/Users/tom/miniconda3/envs/numpy2.0_released/lib/python3.11/site-packages/xarray/core/formatting.py:973: in <listcomp>
    temp = [wrap_indent(short_array_repr(obj), start="    ") for obj in (a, b)]
/Users/tom/miniconda3/envs/numpy2.0_released/lib/python3.11/site-packages/xarray/core/formatting.py:634: in short_array_repr
    array = to_duck_array(array)
/Users/tom/miniconda3/envs/numpy2.0_released/lib/python3.11/site-packages/xarray/namedarray/pycompat.py:129: in to_duck_array
[...]

PR #1016, opened 2026-06-16
fix!: remove ManifestArray.chunks to prevent load attempts

### LINKED ISSUE #114: Ugly error from xarray chunkmanager when loading ManifestArray

maxrjones:
This is a breaking change to the developer API, but I wanted to propose it because I think it's worth considering as a solution to the class of bugs related to the opaque "Could not find a Chunk Manager" `TypeError` from xarray.

Xarray identifies a chunked array by [looking for the `.chunks` attribute](https://github.com/pydata/xarray/blob/77ec40a202620ce600b3e4971bef3b16b5ee1699/xarray/namedarray/pycompat.py#L91-L92). By defining `ManifestArray.chunks`, we advertise that we fulfill xarray's chunked-array protocol. However, `ManifestArray`s are not computable chunked arrays (they hold virtual references, not deferred computations). Because Xarray thinks they're chunked arrays, xarray routes every load/compute/coerce path through a chunk manager that doesn't exist for `ManifestArray`. We can resolve these issues by removing `ManifestArray.chunks` so that xarray no longer treats `ManifestArray`s as chunked arrays, and instead point people to `ManifestArray.metadata.chunks`.

I checked that we're not losing functionality. concat/merge dispatch through NEP-18 `__array_function__` (plus the general duck-array surface), which is independent of whether xarray classifies the array as chunked. The only test changes are `marr.chunks` -> `marr.metadata.chunks`.

## Migration

`ManifestArray` no longer has a `.chunks` attribute. Read the Zarr chunk shape from `.metadata.chunks` instead:

```python
marr.chunks            # ❌ AttributeError
marr.metadata.chunks   # ✅ e.g. (4, 4)
```

This is the same one-line change applied internally throughout VirtualiZarr. Note that `xarray.Dataset.chunks` / `Variable.chunksizes` on a virtual dataset now report no dask-style chunking (previously they returned a malformed value), which is the correct behavior for a non-computable array.

Closes #114, #354, #382.

cc @TomNicholas for thoughts on this potential approach

- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


TomNicholas:
> previously they returned a malformed value

I thought the reason we used `.chunks` was so that it showed up in the repr?

maxrjones:
> > previously they returned a malformed value
> 
> I thought the reason we used `.chunks` was so that it showed up in the repr?

We don't need `ManifestArray.chunks` for the repr. The repr is from `ManifestArray__repr__`, not from xarray inspecting `.chunks`. 

The repr in this branch is unchanged:
image

norlandrhagen:
Nice @maxrjones. I ran into this error a lot. This will be a nice QOL change.

chuckwondo (approved):
Thanks @maxrjones. Seems reasonable to me.

TomNicholas (approved):
Thanks @maxrjones . This seems very clever - I wish I had thought of doing it this way originally!

Maybe we should add a test for the new expected behavior?

PR #1029, opened 2026-06-25
Add `nrefs()` method to VirtualiZarrDatasetAccessor

### LINKED ISSUE #573: Add `.nrefs` method to accessor that returns number of virtual chunk references
It would be neat to add a method `vds.nrefs()` to the accessor, which returns the total number of virtual references in the dataset. It's useful to know how many millions of references you're about to try and write to icechunk (or kerchunk json...)

```python
class VirtualDatasetAccessor
    def nrefs(self: xr.Dataset) -> int:
        """Count the number of virtual references in the dataset"""
    
        # this just ignores non-virtual variables
        return sum(
            len(var.data.manifest) for var in vds.variables.values() if isinstance(var.data, ManifestArray)
        )
```

maitriupadhyay03-cell:
# What I did

Adds a `nrefs()` method to `VirtualiZarrDatasetAccessor` that returns the total number of virtual chunk references in a dataset, ignoring non-virtual (loadable) variables. Useful for knowing how many references you're about to write to Icechunk or Kerchunk before actually doing it.

The implementation follows the same pattern as the existing `.nbytes` property.

Acceptance criteria:

- [x] Closes #573
- [x] Tests added
- [ ] Tests passing
- [x] Full type hint coverage
- [x] New functions/methods are listed in an appropriate `*.md` file under `docs/api`

### DIFF
--- a/virtualizarr/tests/test_xarray.py
+++ b/virtualizarr/tests/test_xarray.py
@@ -1174,14 +1174,14 @@ def test_nrefs(simple_netcdf4, local_registry):
         registry=local_registry,
         parser=parser,
     ) as vds:
-        # simple_netcdf4 has one virtual variable 'air' with a single chunk
+        # simple_netcdf4 has one virtual variable 'foo' with a single chunk
         assert vds.vz.nrefs == 1
 
     with open_virtual_dataset(
         url=simple_netcdf4,
         registry=local_registry,
         parser=parser,
-        loadable_variables=["air"],
+        loadable_variables=["foo"],
     ) as vds:
         # when the only variable is loadable (non-virtual), nrefs should be 0
         assert vds.vz.nrefs == 0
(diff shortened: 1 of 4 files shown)


TomNicholas (approved):
Great, thank you!

virtualizarr/tests/test_xarray.py
```diff
@@ -1163,3 +1163,25 @@ def test_isel_iterative_append_simulation(self, array_v3_metadata):
...
+@requires_imagecodecs
+def test_nrefs(simple_netcdf4, local_registry):
+    parser = HDFParser()
+    with open_virtual_dataset(
+        url=simple_netcdf4,
+        registry=local_registry,
+        parser=parser,
+    ) as vds:
+        # simple_netcdf4 has one virtual variable 'foo' with a single chunk
+        assert vds.vz.nrefs == 1
```
TomNicholas: You defined it as a method not a property - so I think you need the braces

```suggestion
        assert vds.vz.nrefs() == 1
```

Issue #883, opened 2026-02-24
Concatenation in time of (time, y, x)-dimentional data fails


mliukis:
I am trying to concatenate our (time, y, x)-dimensional data in `time`:
* All data variables,  except for `x` and `y` coordinates, have consistent chunking size of (1, 512, 512) in each dataset. 
* `x` and `y` coordinates have chunk size set to None. 
* `x` and `y` coordinates differ across multiple NetCDF files.

The code fails with NotImplementedError for the unsupported indexer. Do `x` and `y` coordinates need to be the same across multiple files that I am trying to concatenate?

Here is the code to reproduce the issue with 2 NetCDF files:
```
import obstore
import os
import virtualizarr as vz
from virtualizarr import open_virtual_dataset, open_virtual_mfdataset
from virtualizarr.parsers import HDFParser
from obspec_utils.registry import ObjectStoreRegistry
import xarray as xr

nc_bucket = 's3://its-live-data'
nc_dir = 'test-space/virtual-cubes'

granules = [
    "LC08_L1GT_020121_20231013_20231102_02_T2_X_LC09_L1GT_020121_20231106_20231106_02_T2_G0120V02_P084.nc",
    "LC08_L1GT_020120_20201121_20210315_02_T2_X_LC08_L1GT_020120_20210124_20210305_02_T2_G0120V02_P051.nc",
]

store = obstore.store.from_url(nc_bucket, region="us-west-2", skip_signature=True)
registry = ObjectStoreRegistry({nc_bucket: store})
parser = HDFParser(drop_variables=['mapping'])

all_vds = []

for granule in granules:
    nc_vds = vz.open_virtual_dataset(
        url=os.path.join(nc_bucket, nc_dir, granule),
        parser=parser,
        registry=registry,
        loadable_variables=['time', 'y', 'x'],
        decode_times=True,
    )
    # Delete data variable that needs a special handling - interested in some attributes values only
    del nc_vds['img_pair_info']
    all_vds.append(nc_vds)

combined_ds = xr.concat(
    all_vds,
    dim=["time"],
    join='outer',
    coords='minimal',
    compat='override',
    combine_attrs='override'
)
```

Raises exception:
```
File [.../lib/python3.13/site-packages/virtualizarr/manifests/indexing.py:177](http://localhost:8888/.../lib/python3.13/site-packages/virtualizarr/manifests/indexing.py#line=176), in apply_selection_1d(marr, indexer_1d, length)
[...]

PR #1031, opened 2026-06-29
Add lazy reindexer that can operate on ManifestArrays

### LINKED ISSUE #882: Support for sparsely tiled spatial grids?
I've been trying to use virtualizarr to virtualize a dataset that uses the [MODIS Sinusoidal Grid](https://modis-land.gsfc.nasa.gov/MODLAND_grid.html) and I'm wondering if the maintainers of this project have considered this problem before and have any advice. Essentially the problem I have is that I want to virtualize a dataset that is split into multiple files on a sparse spatial grid for a single time step. The nature of the MODIS grid is that the number of tiles varies between rows so it's not trivial to concatenate along the X and Y dimensions for the data. 

### Current Behavior / Steps to Reproduce
Let me try to explain this with an example (sorry if this is not very "minimal" but this is a fairly complex use case):

```python
import earthaccess
import re
import virtualizarr as vz
import xarray as xr

from virtualizarr.parsers import DMRPPParser

# Get some data that uses the MODIS Sin Grid
earthaccess.login()
results = earthaccess.search_data(
    short_name="VNP43MA4",
    temporal="2026-01-27",
)
parsed_url = urlparse(results[0].data_links(access="indirect")[0])
domain = parsed_url.netloc
granule_dmrpp_urls = [
    granule.data_links(access="indirect")[0] + ".dmrpp" for granule in results
]

http_store = HTTPStore.from_url(
    f"https://{domain}",
    client_options={
        "default_headers": {
            "Authorization": f"Bearer {token}",
        },
    },
)
obstore_registry = ObjectStoreRegistry({f"https://{domain}": http_store})

# This is maybe unnecessary, but I prefer to virtualize one row of tiles at a time
# so I parse the urls based on their grid index e.g., h02v07 etc.
tile_dict = {}
for url in granule_dmrpp_urls:
    match = re.search(r'\.h(\d{2})v(\d{2})\.', url)
    if match:
        h, v = int(match.group(1)), int(match.group(2))
        tile_dict[(h, v)] = url

row_vals = sorted(set(v for h, v in tile_dict.keys()))

vds_rows = []
for row in row_vals:
    row_urls = [url for key, url in sorted(tile_dict.items()) if key[1] == row]
    vds_row = vz.open_virtual_mfdataset(
        urls=row_urls,
        registry=obstore_registry,
        parser=DMRPPParser(
            group="/HDFEOS/GRIDS/VIIRS_Grid_BRDF/Data_Fields",
            skip_variables=["Projection"],
        ),
        combine="nested",
        concat_dim="XDim",
        parallel="dask",
        loadable_variables=["XDim", "YDim"]
    )
    vds_rows.append(vds_row)
vds = xr.concat(vds_rows, dim="YDim", join="outer")
```

This code will throw an error along these lines:
```
xarray.structure.alignment.AlignmentError: cannot reindex or align along dimension 'XDim' because of conflicting dimension sizes: {4800, 21600, 16800, 28800, 33600, 9600, 12000, 25200, 27600, 34800, 32400, 30000, 13200}
```

This is expected behavior since the nested concat won't fill the empty parts of the grid with nan. I tried doing this another way by assuming the tile size a priori, loading all the datasets one at a time, and then using combine_by_coords, but I ran into a NotImplementedError:

```python
TILE_SIZE = 1200

gridded_vds = []
# As an aside, this loop will be extremely slow for a full size grid of 36x18 tiles
for (h, v), url in sorted(tile_dict.items()):
    tile_vds = vz.open_virtual_dataset(
        url,
        registry=obstore_registry,
        parser=DMRPPParser(
            group="/HDFEOS/GRIDS/VIIRS_Grid_BRDF/Data_Fields",
            skip_variables=["Projection"],
        ),
    )

    # Assign global pixel coordinates based on tile position
    x_start = h * TILE_SIZE
    y_start = v * TILE_SIZE

    tile_vds = tile_vds.assign_coords({
        'XDim': np.arange(x_start, x_start + TILE_SIZE),
        'YDim': np.arange(y_start, y_start + TILE_SIZE),
    })
    gridded_vds.append(tile_vds)

combined = xr.combine_by_coords(
    gridded_vds,
    combine_attrs='override',
    fill_value=np.nan,
)
```

This throws an error as well:
```
/app/venv/lib/python3.13/site-packages/virtualizarr/manifests/indexing.py", line 177, in apply_selection_1d
    raise NotImplementedError(
        f"Unsupported indexer. So-called 'fancy indexing' via numpy arrays is not supported, but received {indexer_1d}"
    )
NotImplementedError: Unsupported indexer. So-called 'fancy indexing' via numpy arrays is not supported, but received [  -1   -1   -1 ... 1197 1198 1199]
```

### Expected Behavior

The proper way to support this use case is probably the latter example, in which case [this feature](https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/manifests/indexing.py#L155) of ManifestArray needs to be implemented. I am able to do the combine_by_coords operation from the second example when I download the data locally and work with it that way. Are you all aware of any workarounds that would allow me to virtualize a dataset like this in the mean time?

sharkinsspatial:
# What I did
Added lazy `reindex` and `reindex_like` accessor methods that align a virtual dataset to new coordinate labels without materializing data, padding missing positions in the new target grid with null-path manifest entries so that the missing chunks are read as `fill_value`.

Acceptance criteria:

- [x] Closes #882 #382 #883 https://github.com/nasa-jpl/its_live_production/discussions/59
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage
- [x] New functions/methods are listed in an appropriate `*.md` file under `docs/api`

### DIFF
--- a/virtualizarr/tests/test_xarray.py
+++ b/virtualizarr/tests/test_xarray.py
@@ -17,8 +17,14 @@ from virtualizarr import (
     open_virtual_datatree,
     open_virtual_mfdataset,
 )
-from virtualizarr.manifests import ChunkManifest, ManifestArray
+from virtualizarr.manifests import (
+    ChunkManifest,
+    ManifestArray,
+    ManifestGroup,
+    ManifestStore,
+)
 from virtualizarr.manifests.indexing import SubChunkIndexingError
+from virtualizarr.manifests.utils import create_v3_array_metadata
 from virtualizarr.parsers import HDFParser
 from virtualizarr.tests import (
     requires_dask,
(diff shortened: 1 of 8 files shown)


virtualizarr/tests/test_xarray.py
```diff
@@ -377,6 +383,298 @@ def test_combine_by_coords_keeping_manifestarrays(
...
+class TestAlignment:
+    """
+    Outer-style alignment introduces coordinate labels an array doesn't have and
+    fills the new positions with NaN. The NaN fill is *not* fundamentally a
+    materializing operation: a missing chunk in a manifest (``MISSING_CHUNK_PATH``)
+    already reads back as ``fill_value``, so the NaN can live purely as a chunk
+    reference and only be realized when a reader asks for those bytes (see
+    ``test_missing_chunk_reads_back_as_virtual_nan``).
+
+    What actually breaks today is xarray's *reindex machinery*, not the manifest
```
TomNicholas, line 395: What would it take to support this type of reindexing in xarray?

sharkinsspatial: I did a few aborted investigations with Claude.  Once I finalize this Draft I'll add some more context in the initial comment but I should start by saying my understanding of xarray internals is very limited.  I'll have to look back through the results of my initial Claude session last week (I think I serialized as a memory) but IIUC 
1. xarray's reindex operates at the element level rather than the array level so there would be no duck array hook for reindex.
2. Internally reindex is using a gather operation that operates at the element level.  This element level gather could move elements within a chunk.  We can only operate lazily at the chunk level without materializing the chunks so this would probably be undesirable. 
3. Internally reindex also uses an element level `where` call that would require materializing.

My guess is that there may be some path to potentially updating xarray internals to handle this but I couldn't grok how that would be handled.  Maybe you or @dcherian have insights here.

In the interim, users can call this accessor reindexer to mimic what would happen internally in xarray and then use xarray's native combination methods on the pre-aligned data.

sharkinsspatial: I'll actually probably try to add an additional `align` accessor method which would be 
1. Calculate the union of the both ds indexes.
2. Reindex each ds onto these unioned indexes

To make the the user experience a bit more seamless.

dcherian: Xarray will provide the backing array library with an array indexer of ints with `-1` indicating missing. What you can do is take that indexer, partition it by chunks and then for each partition assert either the partition has no `-1` or is all `-1` (that is the case you can support). if so, you should be able to translate that indexer to whatever approach is used here. If you support this, you will get Xarray's reindex & alignment machinery for free.

sharkinsspatial: @dcherian 🙇‍♂️ Thanks for this. I couldn't make the leap of logic for partitioning the indexer by chunk boundary.  I'll try to investigate tomorrow, hopefully I can get this working for the simple case before we land https://github.com/zarr-developers/VirtualiZarr/pull/954 and then someone else will be on the hook for rectilinear partitioning 😆

PR #1032, opened 2026-06-30
Reindex via indexer

### LINKED ISSUE #882: Support for sparsely tiled spatial grids?

sharkinsspatial:
# What I did

Thanks to @dcherian for recommending this approach 🙇 

Added support for ManifestArray to handle xarray's reindex/alignment indexer at the array layer, so vanilla `xr.reindex`, `xr.align(join="outer")` that would introduce `nan`s work with virtual arrays and stay lazy. Missing positions become null-path manifest entries that Zarr returns as the array's fill_value — no materialization. 

- indexing.py — route a "fancy" integer-array indexer to a chunk-grid remap instead of rejecting it, `_collapse_outer_indexer` reshapes broadcast (vectorized) indexers back to their 1D per-axis form.
- reindex.py (new) — `chunk_map_from_indexer` translates xarray's per-element reindex indexer (e.g. [0,1,2,-1,-1], where -1 marks a missing label) into a per-chunk plan. It slices the indexer into chunk-sized blocks and requires each block to be either all -1 → emit an empty null-path chunk, or a contiguous run that exactly fills one source chunk → reuse that chunk's reference. Anything else (a block mixing real and missing positions, or a partial/reordered chunk) would mean materializing, so it raises. The result is one source-chunk index or null chunk per target chunk slot.
- array_api.py —` np.where` returns the gathered array unchanged when the fill mask matches its null chunks.
- array.py — adds `ManifestArray._reindex_axis` and admits ndarray in `__array_function__`.

This replaces the bespoke manifest surgery people need today to align virtual datasets onto a common grid when the alignment would result in `nan`s.

By design there's no chunk splitting / within-chunk reordering — target labels must land on chunk boundaries. Because align sorts the union ascending, a descending coordinate (e.g. ITS_LIVE y coord) requires explicit reindex with a descending target. 

- [x] Closes #882 #382 #883 https://github.com/nasa-jpl/its_live_production/discussions/59
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


virtualizarr/tests/test_manifests/test_reindex.py
TomNicholas, line 1: Every test here seems to test private API. Isn't the public `ManifestArray` API that these tests rely upon actually the implementation of `where`? So we should test that instead?

virtualizarr/tests/test_native_align.py
```diff
@@ -0,0 +1,223 @@
...
+CODECS = [{"configuration": {"endian": "little"}, "name": "bytes"}]
+F32 = {
+    1.0: b"\x00\x00\x80\x3f",
+    2.0: b"\x00\x00\x00\x40",
+    3.0: b"\x00\x00\x40\x40",
+}
+I32 = {10: b"\x0a\x00\x00\x00", 20: b"\x14\x00\x00\x00", 30: b"\x1e\x00\x00\x00"}
+
+
+def _ds_1d(x, path, dtype="float32", fill_value=float("nan"), chunk=1, name="foo"):
```
TomNicholas: Surely we have fixtures like this already

virtualizarr/tests/test_native_align.py
```diff
@@ -0,0 +1,223 @@
...
+        assert isinstance(result["foo"].data, ManifestArray)
+        assert list(result.x.values) == [0, 1, 2, 3, 4]
+        assert result["foo"].data.manifest.dict() == {
+            "0": {"path": "file:///a.nc", "offset": 0, "length": 4},
+            "1": {"path": "file:///a.nc", "offset": 4, "length": 4},
+            "2": {"path": "file:///a.nc", "offset": 8, "length": 4},
+        }
+
+    def test_reindex_float_reads_back_nan(self):
+        import obstore as obs
```
TomNicholas: Tell Claude not to do this! Mine always forgets too

virtualizarr/tests/test_native_align.py
TomNicholas, line 1: What does "native" mean here?

TomNicholas (commented):
Excited to get this to work, but I think we'll have to go back and forth re-organizing and polishing before I feel confident about this.

TomNicholas:
Also it would be quite useful to have an issue/docs example / both showing how this actually solves user's NaN-padding and alignment problems.

virtualizarr/tests/test_native_align.py
sharkinsspatial: I can change this nomenclature.  I was using it originally to distinguish from the accessor based approach I was using before while I was testing both but it doesn't really make sense by itself 👍

sharkinsspatial: Closed with https://github.com/zarr-developers/VirtualiZarr/pull/1032/changes/28e241071e1055ae4188db29a5886c9a472eee20

virtualizarr/tests/test_native_align.py
sharkinsspatial: Closed with https://github.com/zarr-developers/VirtualiZarr/pull/1032/changes/059248399fcca46788ac72ff82e36e1c200f12d7

sharkinsspatial:
Tests are failing due to an upstream GRIB parser issue.

sharkinsspatial:
@TomNicholas This is fully ready for review now.

examples/V2/its_live.ipynb
TomNicholas, line 1: Lets include the cell outputs!

virtualizarr/tests/test_manifests/test_reindex.py
```diff
@@ -0,0 +1,205 @@
...
+        codecs=[{"configuration": {"endian": "little"}, "name": "bytes"}],
+        fill_value=float("nan"),
+        dimension_names=dims,
+    )
+    return ManifestArray(
+        metadata=metadata, chunkmanifest=ChunkManifest(entries=entries)
+    )
+
+
+class TestReindexGatherKeepsChunks:
```
TomNicholas, line 38: These are great, thanks

docs/how_to/validation.md
```diff
@@ -255,6 +254,16 @@ print(combined)
 The combined variable is still a single `ManifestArray` of shape `(2, 3, 4)`, but only the first timestep references real chunks.
 Reading data from the second timestep would return the array's `fill_value`.
 
+In the previously mentioned case when a dataset is "inherently sparse at a global level, whilst still being dense at a regional level", when users want to concatenate these sparse datasets they can use normal xarray concatenation patterns with outer join.
+
+```
+xr.concat([ds_1, ds_2], dim="time", join="outer")
+```
```
TomNicholas, line 261: Can you think of a scenario when a user would want to explicitly call `.reindex`? If so we should document that too.

virtualizarr/tests/test_manifests/test_reindex.py
```diff
@@ -0,0 +1,205 @@
...
+xarray's reindex/alignment machinery conforms an array to new labels by indexing
+it with an integer "gather" indexer that has ``-1`` at every missing position
+(e.g. ``[0, 1, 2, -1, -1]``). ManifestArray supports that indexer at the array
+level -- ``marr[indexer]`` -- by remapping the chunk grid: a run of ``-1`` that
+covers a whole chunk becomes a null-path chunk (reads back as ``fill_value``),
+and a chunk-aligned run of real indices keeps its chunk reference. Nothing that
+would split a chunk is allowed.
+
+These tests drive that public indexing path directly rather than the private
+``chunk_map_from_indexer``/``_reindex_axis`` helpers that implement it.
```
TomNicholas: superfluous comments
```suggestion
```

virtualizarr/manifests/array.py
```diff
@@ -333,6 +336,71 @@ def with_fill_value_only(self, fill_value: Any) -> "ManifestArray":
         )
         return ManifestArray(metadata=new_metadata, chunkmanifest=empty_manifest)
 
+    def _reindex_axis(
+        self, axis: int, chunk_map: list[int | None], new_size: int
+    ) -> "ManifestArray":
```
TomNicholas: The only place this private method is called is within `indexing.py`, so I think it should become a function and live in `indexing.py` instead (or in `reindex.py` and imported).

virtualizarr/manifests/array.py
```diff
@@ -333,6 +336,71 @@ def with_fill_value_only(self, fill_value: Any) -> "ManifestArray":
...
+        Each entry of ``chunk_map`` is the source chunk index to copy into that
+        target chunk slot, or ``None`` for a missing (null-path) chunk that reads
+        back as ``fill_value``. ``new_size`` is the new length of ``axis`` in
+        elements; the chunk size is unchanged.
+        """
+        from virtualizarr.manifests.manifest import (
+            MISSING_CHUNK_PATH,
+            ChunkManifest,
+        )
+        from virtualizarr.manifests.utils import copy_and_replace_metadata
```
TomNicholas: Imports should be moved to the top of files where possible

virtualizarr/manifests/array_api.py
```diff
@@ -54,6 +69,43 @@ def result_type(*arrays_and_dtypes: Union["ManifestArray", np.dtype]) -> np.dtyp
...
+    raise NotImplementedError(
+        "np.where on a ManifestArray is only supported for the reindex/alignment "
+        "fill pattern (filling an array's own missing chunks); general masking "
+        "would require materializing values."
+    )
+
+
+def _missing_element_mask(marr: "ManifestArray") -> np.ndarray:
+    """Boolean element-mask (shape == marr.shape), True at missing (null) chunks."""
+    from virtualizarr.manifests.manifest import MISSING_CHUNK_PATH
```
TomNicholas: Move to top of file

TomNicholas (approved):
I'm not going to pretend I understand every line of the logic but the tests look good enough now.

Just some minor nits then we can merge!

docs/how_to/validation.md
sharkinsspatial: @TomNicholas I'm not aware of a case where users would directly use `reindex` but my `xarray` foo is fairly limited 😆

TomNicholas: If they wanted to reorder chunks along a dimension they might. But I can't think of a sane example of an archive for which this would be needed. We can worry about this later, if this ever comes up.

Issue #947, opened 2026-03-26
Optimization idea: Files as shards?


TomNicholas:
**Can we optimize reading of archival data containing small chunks by pretending that files are zarr shards**?

Maximum read throughput from S3 is achieved for objects above a certain size (maybe a few MBs? This number depends on various implementation details, and so is lower in rust than python), but often archival data has chunks < that size, but files > that size. If we can read e.g. one TIFF/HDF5 file in one request, then read throughput from the virtual store could sidestep the bottleneck of the overhead of reading many tiny chunks.

Whether or not this would actually be faster overall than just fetching chunks individually depends on the density of chunks within our "shard" (which can be visualized with @maxrjones' [tool](https://huggingface.co/spaces/maxrjones/nisar-manifest-explorer)). If they are tightly packed then the single shard request would be fetching only data it needs, and this would be a strict improvement. But if they are too spaced out for any reason (e.g. because the file-shard contains multiple variables, with interleaved chunks, which HDF5 definitely does), you might end up fetching too much data that isn't relevant for that array, and it be less efficient overall.

Sharding is Zarr's implementation of internal chunking, where one shard is stored as one object, but contains many chunks, and also contains the shard index, telling you where exactly inside the shards each chunk is. Zarr v3 implements sharding as a codec, so in theory maybe we could change our parsers to return a codec chain containing the `ShardingCodec`, with the shard index pointing at the internal TIFF/HDF5 chunks?

Unfortunately I don't think that will work, because where would we put the shard index? 
- In native zarr v3 it is part of the object, but in virtual stores we can't change the object. So does it have to go in memory? Does that then mean it needs to get serialized into icechunk separately somehow? It's annoying that sharding is a codec and not part of the zarr spec...
- We can't [treat the file itself as the shard](https://www.slideshare.net/slideshow/using-a-hierarchical-data-format-v5-file-as-zarr-v3-shard/281967847) either because the `ShardingCodec` assumes a particular format for the index within the object, which won't match pre-existing archival TIFF/HDF5 files.

This just leaves me thinking that zarr should not have implemented sharding as a codec...

cc @sharkinsspatial @jsignell

jsignell:
I like Max's idea of batching reads in ways other than sharding - I guess you always know which chunks are contained within the same file?

But Davis' idea of a custom codec that implements sharding for an index that is expressed slightly differently is pretty compelling. It seems like this might be better able to understand how the chunks are linearized since it would be customized.

[...]

PR #1033, opened 2026-07-01
Coalescing ManifestStore.get_many

TomNicholas:
Implements the chunk-coalescing idea from #947: teach `ManifestStore` to satisfy a batch of chunk requests with fewer, larger reads by deriving an "effective shard index" from the manifests at read time.

`ManifestStore.get_many` (overriding the new cross-key `Store.get_many`) resolves the requested chunk keys to `(source file, byte range)`, groups requests by source file, coalesces each group into runs, and serves each run with one ranged read that is sliced back into per-chunk buffers. No file-format assumptions, no spec changes. Keys that aren't plain manifest-backed chunks (metadata, inlined, missing) fall back to per-key `get`.

Coalescing uses the object_store / async-tiff two-knob model: references merge into one read when their gap is `<= coalesce_max_gap_bytes` and the merged read stays `<= coalesce_max_bytes`.

### Why `coalesce_max_gap_bytes` defaults to 0

Benchmarking a 2D map-tile query (5×5 chunks of `air_temperature`) against a remote Met Office file on Azure, versus the status quo of one `get` per chunk:

| config | reads | fetched | over-read | median | vs baseline |
|---|---|---|---|---|---|
| no coalescing (per-chunk get) — *status quo* | 25 | 621 KiB | 1.00× | 273 ms | 1.00× |
| **gap=0 (merge only adjacent)** | 5 | 621 KiB | 1.00× | **97 ms** | **2.8× faster** |
| gap=1 MiB, cap=8 MiB | 1 | 1964 KiB | 3.16× | 456 ms | 1.7× slower |
| gap=1 MiB, cap=256 KiB | 5 | 621 KiB | 1.00× | 99 ms | 2.8× faster |
| gap=256 KiB, cap=8 MiB | 5 | 621 KiB | 1.00× | 96 ms | 2.8× faster |

A 2D spatial box is contiguous along one axis but strided along the other (row-major chunk layout), so a large gap bridges the row stride and pulls in the chunks between rows — reading ~3× the needed bytes and running *slower than no coalescing*. Merging only adjacent references is a pure win: zero over-read, ~2.8× faster (collapsing 25 requests into 5). Hence the default gap of 0; `coalesce_max_bytes` (8 MiB) just bounds the size of any single read. A non-zero gap is available for callers who know their access is contiguous.

Depends on the `Store.get_many` API and its use in `BatchedCodecPipeline` in zarr-python (zarr-developers/zarr-python#4112, zarr-developers/zarr-python#4113); until those are released this is a dormant override, tested in isolation.

xref #947

Draft — feedback welcome on the approach and the config surface.

### DIFF
--- a/virtualizarr/manifests/store.py
+++ b/virtualizarr/manifests/store.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
+import asyncio
 import warnings
-from collections.abc import AsyncGenerator, Iterable
+from collections.abc import AsyncGenerator, Iterable, Sequence
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Literal, TypeAlias
+from typing import TYPE_CHECKING, Literal, NamedTuple, TypeAlias
 from urllib.parse import urlparse
 
 from obspec_utils.registry import ObjectStoreRegistry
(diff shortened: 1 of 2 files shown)


betolink:
This is really cool work @TomNicholas, I like that this will be handled at the chunk manifest level since you have a map of the byte ranges. I'm curious on why this should be a Zarr responsibility.

TomNicholas:
> I'm curious on why this should be a Zarr responsibility.

As opposed to what?

For context there is a rationale here: https://github.com/zarr-developers/VirtualiZarr/issues/947#issuecomment-4859947806

maxrjones:
> > I'm curious on why this should be a Zarr responsibility.
> 
> As opposed to what?
> 
> For context there is a rationale here: [#947 (comment)](https://github.com/zarr-developers/VirtualiZarr/issues/947#issuecomment-4859947806)

Probably as opposed to a lower level, e.g., https://developmentseed.org/obstore/latest/api/get/#obstore.get_ranges has a coalesce argument or a wrapper store for coalescing could be added to obspec_utils - https://obspec-utils.readthedocs.io/en/latest/.

It does seem this is a more general problem than ManifestStore, which mostly exists to handle indirection.

TomNicholas:
See https://github.com/zarr-developers/VirtualiZarr/issues/947#issuecomment-4867733151 for why I don't think that can work - I want to coalesce requests over multiple independent keys, which means I need to be able to request multiple zarr keys at once.

> `ManifestStore`, which mostly exists to handle indirection.

`ManifestStore` (and `IcechunkStore`, which would implement the same algorithm in rust) is the only place that knows both about the zarr keys the users wants and also about the storage layout (via the manifests).

PR #1036, opened 2026-07-08
Fix ContainsGroupError when writing a DataTree with region/append_dim

aaronspring:
## Problem

Writing a virtual `DataTree` with `region` or `append_dim` (forwarded to each node) always failed with `zarr.errors.ContainsGroupError`. `virtual_datatree_to_icechunk` created every node's group unconditionally with `Group.from_store`, but `region`/`append_dim` require the group and its arrays to **already** exist — so the write blew up before it ever got to writing the region/append.

This is the same class of failure as #1001, but for the DataTree path.

## Fix

Mirror the branch that `virtual_dataset_to_icechunk` already had on `main`: open the existing group for `region`/`append_dim`, create it otherwise.

```python
for store_path, vds in paths_and_virtual_datasets:
    if kwargs.get("append_dim") or kwargs.get("region"):
        group = Group.open(store=store_path, zarr_format=3)   # must already exist
    else:
        group = Group.from_store(store=store_path, zarr_format=3)  # create
```

## Test

`test_write_datatree_region` was marked `xfail` with the reason *"This doesn't work in xarray either, maybe not even intended?"* — but that was a misdiagnosis. On `main`, running it with `--runxfail` fails with `ContainsGroupError` at the group-creation step, not in any xarray region logic. With this fix the test runs to completion and its assertions pass, so the `xfail` marker is removed.


aaronspring:
CI note: the 3 red build jobs (`test-py312`, `test-py313`, `minimum-versions`) fail only on `test_parsers/test_grib.py` (`test_grib_virtual_dataset`, `test_grib_only_variables`, `test_grib_chunk_decodes_via_codec`) — the GRIB/gribberish parser returning empty `data_vars` / a `NodeTypeValidationError`. These are pre-existing and unrelated to this PR, which touches only the icechunk DataTree writer; all 550 other tests pass. The same 3 failures appear on #1034 (a pre-commit.ci autoupdate that changes no source), #1032, #1033, and on scheduled `main` runs.

TomNicholas:
Thanks @aaronspring !

PR #1038, opened 2026-07-13
perf: add fast path for local hdf5

ianhi:
# What I did
Added a fast path for parsing HDF files that are local. For these the Blockreader can cause massive perforamnce hits. Especially for a larger files (40 GB), you can actually end up reading more data than the file size! especially for b-trees where you end up revisting nodes.

Behavior changes:
- `reader_factory` now defaults to `None`. Passing any factory, including `BlockStoreReader`, opts back into the reader path for local files.
- Opening a file by path makes HDF5 take a file lock, which the reader path never did. If another process holds the file open for writing, the parser retries with `locking=False`. It doesn't disable locking up front because HDF5 rejects reopening a file this process already has open with a different locking setting.

attn @ilan-gold  curious if this helps with the performance issues here: https://annbatch.readthedocs.io/en/stable/preshuffling.html

Acceptance criteria:

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/docs/api/parsers/hdf5.md
+++ b/docs/api/parsers/hdf5.md
@@ -1,3 +1,32 @@
 # HDF5/NetCDF4
 
+!!! note "Chunk-dense files over the network"
+
+    Local files are walked with HDF5's own index-aware driver, so building a
+    manifest reads only the chunk index. For **remote** chunk-dense files (many
+    small chunks — e.g. sparse single-cell `.h5ad`), the default block reader
+    can instead fetch a large fraction of the whole file just to recover that
+    index.
+
+    The robust fix is to **download the file once and parse it locally**, which
+    takes the native fast path automatically.
+
+    If downloading isn't practical, tune the block size via the `reader_factory`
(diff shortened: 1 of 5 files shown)


TomNicholas:
@maxrjones does this seems sensible to you?

docs/api/parsers/hdf5.md
```diff
@@ -1,3 +1,32 @@
 # HDF5/NetCDF4
 
+!!! note "Chunk-dense files over the network"
+
+    Local files are walked with HDF5's own index-aware driver, so building a
+    manifest reads only the chunk index. For **remote** chunk-dense files (many
+    small chunks — e.g. sparse single-cell `.h5ad`), the default block reader
+    can instead fetch a large fraction of the whole file just to recover that
+    index.
+
```
ilan-gold, line 10: Not relevant here @ianhi but this is exactly what I was saying should be a target for doing in rust, especially over many files simultaneously

ilan-gold:
This fix seems reasonable!

PR #1055, opened 2026-07-24
Fix ZarrParser crash on S3 directory-marker objects

### LINKED ISSUE #1052: ZarrParser crashes on stores with zero-byte S3 'directory marker' objects (invalid literal for int())
## Description

`ZarrParser` crashes with `ValueError: invalid literal for int() with base 10: '<name>'` when the source store contains zero-byte "directory marker" objects — a very common S3/GCS artifact created by tools like `aws s3 sync`, `s3fs`, or `boto3.put_object(Key=prefix + "/")`.

Root cause: `obstore`'s S3 client strips the trailing `/` from listed object keys before they reach VirtualiZarr. `build_1d_chunk_mapping` (in `virtualizarr/parsers/zarr.py`) relies on `np.strings.endswith(paths_np, "/")` to filter these marker objects out of the chunk-key list:

```python
is_directory = np.strings.endswith(paths_np, "/")
```

Since the trailing slash is already gone by the time this check runs, it never matches, and the marker object's key survives into `stripped_keys`. `build_chunk_manifest` then tries to parse it as chunk coordinates and crashes:

```python
coords = np.array(
    [[int(c) for c in key] for key in split_keys], dtype=np.int64
).T
```

I confirmed with a raw `ListObjectsV2` call that the object really does have a key ending in `/` on S3 (e.g. `precip.zarr/x/`, size 0), but by the time `obstore.list()` returns it to Python, the path has already been normalized to `precip.zarr/x` (no trailing slash) — so `endswith(path, "/")` is dead code for any obstore-backed store.

This isn't a rare edge case: in a sweep of 152 real-world public Zarr v2 stores (NOAA/NCAR/weatherbench2 datasets on AWS Open Data), 26 failed with this exact error.

## Reproduces against real public data

```python
from obstore.store import S3Store
import virtualizarr as vz
from virtualizarr.parsers import ZarrParser

store = S3Store("noaa-nwm-retrospective-2-1-zarr-pds", region="us-east-1", skip_signature=True)
registry = ObjectStoreRegistry({"s3://noaa-nwm-retrospective-2-1-zarr-pds": store})
vz.open_virtual_dataset(
    url="s3://noaa-nwm-retrospective-2-1-zarr-pds/precip.zarr",
    registry=registry,
    parser=ZarrParser(),
)
# ValueError: invalid literal for int() with base 10: 'precip'
```

Other public stores that fail the same way: `s3://noaa-nwm-retrospective-2-1-zarr-pds/{gwout,lakeout,ldasout,rtout}.zarr`, `s3://noaa-nwm-retrospective-3-0-pds/{Alaska,CONUS,Hawaii}/zarr/chrtout.zarr`, `s3://ncar-cesm-lens/...`, `s3://ncar-cesm2-lens/...`, `s3://ncar-na-cordex/...`, `gs://gcp-public-data-arco-era5/...`, `gs://weatherbench2/...` — all Zarr v2 stores carrying per-array zero-byte "folder" objects from their original upload.

## Offline minimal reproduction (correction)

**Correction:** I originally posted an "offline" repro here using `obstore.store.MemoryStore` with a marker object written via `put_async(prefix + "/", b"")`. That repro does **not** actually reproduce the bug — I was wrong. `MemoryStore` normalizes the trailing slash away on *both* the write and the list side, so a listing for `prefix + "/"` never even returns the marker in the first place (unlike real S3, where the write and the list go through separate systems: the object genuinely has a trailing slash in its key on S3, is matched server-side by `ListObjectsV2`'s literal prefix comparison, and only loses its trailing slash in `obstore`'s client-side parsing of the *response*). So no in-memory or local `obstore` backend reproduces this — it's specific to the S3 (and presumably GCS) client path.

A faithful, dependency-free reproduction has to fabricate that exact listing shape directly, bypassing `obstore` entirely:

```python
import asyncio

import numpy as np
import zarr

from virtualizarr.parsers.zarr import ZarrFormat, build_chunk_manifest, metadata_as_v3


class _FakeColumn:
    def __init__(self, values):
        self._values = values

    def to_numpy(self):
        return np.array(self._values)


class _FakeBatch:
    def __init__(self, paths, sizes):
        self._columns = {"path": paths, "size": sizes}

    def column(self, name):
        return _FakeColumn(self._columns[name])


class _FakeObstoreStore:
    """Simulates the exact listing shape a real S3 store returns for a
    directory containing both a zero-byte marker object and a real chunk:
    the marker's path already has its trailing slash stripped."""

    def list_async(self, prefix, return_arrow=True):
        async def _stream():
            yield _FakeBatch(
                paths=["arr/c", "arr/c/0.0"],  # "arr/c" is the marker
                sizes=[0, 4],
            )

        return _stream()


metadata = metadata_as_v3(
    zarr.create(
        shape=(10, 10),
        chunks=(5, 5),
        dtype="int8",
        store=zarr.storage.MemoryStore(),
        zarr_format=3,
    ).metadata
)

manifest = asyncio.run(
    build_chunk_manifest(
        obs_store=_FakeObstoreStore(),
        array_path="arr",
        store_base_uri="test://path",
        metadata=metadata,
        on_disk_zarr_format=ZarrFormat.V3,
        on_disk_separator=".",
    )
)
print(manifest.dict())
```

```
ValueError: invalid literal for int() with base 10: 'arr/c'
```

## Suggested fix

Since `obstore` doesn't preserve the trailing slash, `build_1d_chunk_mapping`/`build_chunk_manifest` can't rely on `endswith("/")` to detect directory markers. A robust fix: any listed key that doesn't literally *start with* `array_chunks_prefix` can't be a valid nested chunk key (only genuine descendants literally start with the full prefix string, including its trailing slash) — a directory marker fails this test regardless of whether its trailing slash survived. I have a working patch (adds `is_not_nested = ~np.strings.startswith(paths_np, array_chunks_prefix)` to the existing filter mask) plus a regression test using the fake-store approach above, and will open a PR shortly.

## Environment

- `virtualizarr==2.7.1` (latest release)
- `zarr==3.2.1`
- `obstore==0.8.2`
- Python 3.13

TomNicholas:
## Summary

Fixes #1052.

`obstore` strips the trailing slash from listed S3 object keys, so `build_1d_chunk_mapping`'s `endswith("/")` check can never filter out zero-byte "directory marker" objects (created by e.g. `aws s3 sync`, `s3fs`, or `boto3.put_object(Key=prefix + "/")`) out of the discovered chunk keys. The marker then fails `int()` parsing in `build_chunk_manifest`, crashing with e.g. `ValueError: invalid literal for int() with base 10: 'precip'`.

This isn't a rare edge case — of 152 real-world public Zarr v2 stores I swept (NOAA/NCAR/weatherbench2 datasets on AWS Open Data), 26 failed with this exact error.

## Fix

Filter on whether each listed key actually *starts with* the full `array_chunks_prefix` instead of relying on the (now ineffective) trailing-slash check. A directory marker is keyed exactly as the prefix with its trailing slash removed, so it's one character shorter than the prefix and can never satisfy `startswith(prefix)` — this holds regardless of whether `obstore` has already stripped the marker's own trailing slash, so it's robust to the underlying cause rather than papering over one specific symptom.

## Test plan

- Added `test_build_chunk_manifest_ignores_directory_marker_object` in `virtualizarr/tests/test_parsers/test_zarr.py`. Since no in-memory or local `obstore` backend reproduces the S3-specific slash-stripping quirk itself (writes and lists are both path-normalized consistently for those backends, so a marker written via `MemoryStore.put_async` never even collides with a nested-chunk listing), the test fabricates the exact listing shape a real S3 store returns and feeds it to `build_chunk_manifest` directly.
- Confirmed the new test fails with the reported `ValueError` against the pre-fix code, and passes after the fix.
- Full existing test suite passes locally (`pytest virtualizarr/tests`), `ruff check`/`ruff format` clean, `mypy` clean on the changed file.
- Also verified the fix against the real public bucket from the original bug report (`s3://noaa-nwm-retrospective-2-1-zarr-pds/precip.zarr`), which now opens successfully instead of raising.

[This is Claude Code on behalf of TomNicholas]


TomNicholas:
Closing in favour of #1054, which fixes the same bug (#1052).

Issue #1060, opened 2026-07-31
kerchunk writer breaks with zarr 3.3.0: BytesCodec.endian is now a str, not an enum


aladinor:
### What happens

With `virtualizarr` 2.7.1 and `zarr` 3.3.0, kerchunk export raises:

```
AttributeError: 'str' object has no attribute 'value'
```

`convert_v3_to_v2_metadata` reads `codec.endian.value`, assuming `BytesCodec.endian` is an enum. In zarr 3.3.0 that attribute is a plain `str`.

```
virtualizarr/writers/kerchunk.py:83   in dataset_to_kerchunk_refs
virtualizarr/writers/kerchunk.py:132  in variable_to_kerchunk_arr_refs
virtualizarr/utils.py:127             in convert_v3_to_v2_metadata
>       and codec.endian.value == "big"  # type: ignore[attr-defined]
        for codec in v3_metadata.codecs
```

Since `virtualizarr` requires `zarr>=3.1.6` with no upper bound, a fresh resolve pairs 2.7.1 with 3.3.0.

### Reproducer

Dependencies are declared inline ([PEP 723](https://peps.python.org/pep-0723/)) — save as `mre.py` and run `uv run mre.py`. No data files or virtual dataset needed; a plain `zarr.create_array` yields metadata with the `BytesCodec` that triggers it.

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "virtualizarr==2.7.1",
#     "zarr==3.3.0",
# ]
# ///
"""    uv run mre.py                       # zarr 3.3.0 -> AttributeError
    uv run --with 'zarr==3.2.1' mre.py  # zarr 3.2.1 -> OK
"""
import virtualizarr
import zarr
from virtualizarr.utils import convert_v3_to_v2_metadata

print(f"zarr {zarr.__version__} | virtualizarr {virtualizarr.__version__}")

arr = zarr.create_array(store={}, shape=(4,), chunks=(4,), dtype="float32")
codec = next(c for c in arr.metadata.codecs if type(c).__name__ == "BytesCodec")
print(f"BytesCodec().endian = {codec.endian!r}  type={type(codec.endian).__name__}  "
      f"has .value={hasattr(codec.endian, 'value')}")

convert_v3_to_v2_metadata(arr.metadata)   # <-- virtualizarr/utils.py:127
print("convert_v3_to_v2_metadata: OK")
```

zarr 3.3.0:

```
[...]

PR #1061, opened 2026-07-31
fix: restore compatibility with zarr 3.3.0

d-v-b:
This should close out #1060, sorry for the breakage!

🤖 AI text below 🤖

Two independent breakages under zarr 3.3.0 (566 passing -> 43 failures):

1. zarr 3.3.0 deprecated its string-valued codec enums, so `BytesCodec.endian` is now a plain string rather than an `Endian` member. `convert_v3_to_v2_metadata` did `codec.endian.value`, raising `AttributeError: 'str' object has no attribute 'value'` on every Kerchunk write path (40 failures). Unwrap the enum only when present so both forms work across the supported `zarr>=3.1.6` range.

2. zarr 3.3.0 requires chunk edges to be positive. `kerchunk.hdf4` uses a variable's dimensions as its chunk shape, so a zero-length variable -- e.g. the `FP_*` fire-pixel variables of a MODIS granule that detected no fires -- asks zarr for `chunks=[0]` and raises `ValueError: Chunk size must be positive, got 0` (3 failures). Coerce such an edge to 1 for the duration of the translation, matching what `from_kerchunk_refs` already does when reading those references back.

Verified green against zarr 3.3.0, 3.2.1, and 3.1.6.

Assisted-by: ClaudeCode:claude-opus-5

# What I did

Acceptance criteria:

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/virtualizarr/parsers/hdf4.py
+++ b/virtualizarr/parsers/hdf4.py
@@ -1,11 +1,52 @@
-from collections.abc import Iterable
+import threading
+from collections.abc import Iterable, Iterator
+from contextlib import contextmanager
 from pathlib import Path
+from typing import Any
 
 from obspec_utils.registry import ObjectStoreRegistry
 
 from virtualizarr.manifests import ManifestStore
 from virtualizarr.parsers.kerchunk.translator import manifestgroup_from_kerchunk_refs
 from virtualizarr.types.kerchunk import KerchunkStoreRefs
+
+# Serializes the monkeypatch in _positive_chunk_edges, which mutates zarr.Group
+# for the duration of a single kerchunk translation.
+_patch_lock = threading.Lock()
+
+
+@contextmanager
+def _positive_chunk_edges() -> Iterator[None]:
+    """
+    Clamp zero-sized chunk edges requested by `kerchunk.hdf4` to 1.
+
+    HDF4 granules routinely hold zero-length variables — a MODIS fire-mask granule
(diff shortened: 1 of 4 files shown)


virtualizarr/parsers/hdf4.py
```diff
@@ -1,12 +1,53 @@
...
 from virtualizarr.parsers.kerchunk.translator import manifestgroup_from_kerchunk_refs
 from virtualizarr.types.kerchunk import KerchunkStoreRefs
 
+# Serializes the monkeypatch in _positive_chunk_edges, which mutates zarr.Group
+# for the duration of a single kerchunk translation.
+_patch_lock = threading.Lock()
+
+
+@contextmanager
+def _positive_chunk_edges() -> Iterator[None]:
```
d-v-b, line 19: I can definitely see downsides for this approach. happy to chase this down in kerchunk, if yall don't want it addressed here via monkeypatching

TomNicholas: Yeah this is kind of awful haha. I would prefer we merge the endian fix (I will trim this PR), and worry about the HDF4 edge case separately.

TomNicholas: Turns out there was already an open PR for the endian fix (#1048) so I merged that instead of this PR.

d-v-b: sounds good! happy to close this

PR #1062, opened 2026-07-31
fix: return empty dict from ChunkManifest.dict() for zero-length arrays

d-v-b:
This fixes a bug I observed when working on #1061 

"🤖 AI text below 🤖"

A chunk grid shape containing a 0 made np.nditer raise "ValueError: Iteration of zero-sized operands is not enabled", which also broke keys()/values()/items() and writing such a variable to Kerchunk references. A zero-length array legitimately has no chunks, so dict() now short-circuits to {}. iter_refs() had the same problem and now yields nothing.

Assisted-by: ClaudeCode:claude-opus-5

# What I did

Acceptance criteria:

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


TomNicholas (approved):
Thanks!

Issue #604, opened 2025-05-26
Virtualize ZIP archives?


TomNicholas:
[This article](https://medium.com/@mojodna/tapalcatl-cloud-optimized-tile-archives-1db8d4577d92#:~:text=ZIPs%20facilitate%20random,file%20of%20interest.) is fascinating:

> ZIPs facilitate random access and cheap updates as part of their design.
>
> Random access is achieved by providing a [central directory](https://en.wikipedia.org/wiki/Zip_%28file_format%29#Structure), which contains a listing of all files contained in the archive (as well as their metadata and offsets). Once the central directory has been read, entry offsets can be used to seek to the portion of the file that contains a file of interest.

This very much makes it sound like we should be able to virtualize anything that was zipped! (As long as it was only zipped once - I've heard of double-zipped archives and that's insanity).

I'm not really sure what the API for this would look like though. IIUC ZIP can be applied to anything, so perhaps virtualizarr would need some kind of "`Unzipper`" that could be composed with any `Parser`?

mdsumner:
yml SOzip: https://github.com/sozip/sozip-spec

TomNicholas:
I went down this rabbit hole, and there are various cases here, some we can handle, some we can't. Buckle up.

## GZipped anything

GZip compression scrambles the byte locations within a file, meaning a single chunk is no longer a contiguous byte range on disk. This completely prevents virtualizing any gzipped file. (Data providers, please stop gzipping your files!!!)

## Zip

Zip takes a bunch of "members" and arranges them all into a single file, with a "central directory" at the end listing every member's name, sizes, compression method, and offset. Compression is _per-member_ and optional. This is basically a chunk manifest.

### Single NetCDF/GRIB etc. zipped with `STORED (method 0)`

A parseable file format (e.g. netcdf/grib) that has been zipped with the `STORED` option (i.e. no compression) is virtualizable - we just need to translate the central directory into `ManifestArrays`.

> so perhaps virtualizarr would need some kind of "Unzipper" that could be composed with any Parser?

I _think_ we can make something like this. We can take advantage of the flexibility of obstore/obspec/our registries/our parser protocol, and hopefully make a syntax like this work:

```python
parser = ZipParser(inner=HDFParser())
vds = vz.open_virtual_dataset("s3://bucket/archive.zip", registry=registry, parser=parser)
```

where `ZipParser` can be combined with any single-file parser.

[...]

PR #1065, opened 2026-08-03
Add ZippedZarrParser for virtualizing zipped Zarr stores

TomNicholas:
Part of #604 (the zipped-native-Zarr case).

## What this does

Zipped Zarr stores (e.g. `.zarr.zip` files) can now be virtualized without ever unzipping them, at ingestion or read time:

```python
from virtualizarr import open_virtual_datatree
from virtualizarr.parsers import ZippedZarrParser

vdt = open_virtual_datatree(
    url="s3://bucket/path/store.zarr.zip",
    registry=registry,
    parser=ZippedZarrParser(),
)
```

The zip format keeps a central directory at the end of the archive listing every member's name, size, and offset — effectively a pre-built chunk manifest. The parser reads it with a handful of range requests (EOCD record, ZIP64-aware, plus one batched read of the members' local file headers, whose extra-field lengths may legally differ from the central directory's), and each chunk becomes a virtual reference to a byte range *within the archive itself*. Readers need zero zip-awareness — `ManifestStore` reads and Icechunk virtual refs just issue ordinary range requests into the `.zarr.zip` object.

Notably this means ingestion requires ~3 range requests total regardless of store size — no per-key listing at all.

## Implementation notes

- `parse_zip_index` (central-directory reader) is deliberately format-agnostic and standalone: it's the shared foundation for a possible future generic `ZipParser(inner=...)` combinator for non-Zarr formats inside stored zips (the "Unzipper" idea from #604) — which is also why this class is *not* named `ZipParser`.
- Zarr metadata documents are read through a minimal read-only `zarr.abc.store.Store` over the zip index, so group-tree recursion and V2→V3 metadata conversion reuse the existing `construct_manifest_group_tree` / `metadata_as_v3` machinery unchanged. Supports Zarr V2 and V3, hierarchical groups, and `group=` / `skip_variables=` like `ZarrParser`.
- Only archives with uncompressed (`STORED`) members are supported — the default for `zarr.storage.ZipStore`, and what `zip -0` produces — since compressed members do not preserve chunk byte ranges. Compressed archives raise `NotImplementedError` naming the offending member. `DEFLATE`-compressed zipped Zarr is virtualizable in principle (zip compresses per-member, and member granularity == chunk granularity) but needs a raw-deflate codec registered wherever the refs are read, so it's left as future work.

## Testing

- 11 new tests: V2+V3 roundtrips, byte-for-byte chunk offset verification against `zipfile` ground truth, deflate/not-a-zip errors, datatree/nested groups, `group=`, `skip_variables=`, scalar arrays.
- Validated end-to-end against a real-world 164 MiB hierarchical Zarr V3 `.zarr.zip` (336 members) on Cloudflare R2: parsed with 3 range requests, and loading a variable through the resulting manifest returned correct data.

- [x] Closes part of #604
- [x] Tests added
- [x] Full type hint coverage

### DIFF
--- a/docs/about/releases.md
+++ b/docs/about/releases.md
@@ -8,6 +8,15 @@
     [ManifestStore.nbytes_virtual][virtualizarr.manifests.ManifestStore.nbytes_virtual] — mirroring `ds.vz.nbytes`. Closes
     [#798](https://github.com/zarr-developers/VirtualiZarr/issues/798).
 
+  - Zipped Zarr stores (e.g. `.zarr.zip` files) can now be virtualized via the new
+    [ZippedZarrParser][virtualizarr.parsers.ZippedZarrParser]. The archive's central directory is read with a
+    handful of range requests and each chunk becomes a virtual reference to a byte range within
+    the archive itself, so no unzipping is ever needed — at ingestion or read time. Supports Zarr
+    V2 and V3, hierarchical groups, and remote archives on object storage; only archives with
+    uncompressed (`STORED`) members are supported, which is the default for
+    `zarr.storage.ZipStore`. Part of [#604](https://github.com/zarr-developers/VirtualiZarr/issues/604).
+    By [Tom Nicholas](https://github.com/TomNicholas).
+
   ### Bug fixes
 
   - Fix `ZarrParser` raising `ValueError: invalid literal for int()` when a store contains zero-byte
--- a/docs/api/parsers/zip.md
+++ b/docs/api/parsers/zip.md
@@ -1,0 +1,3 @@
+# Zipped Zarr
+
+::: virtualizarr.parsers.ZippedZarrParser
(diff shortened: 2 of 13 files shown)


TomNicholas:
Now stacked on #1066 (a `ZarrParser` V2 `dimension_separator="/"` bug fix found while writing this parser) — the first commit here will disappear once that merges and this is rebased.

Also added a refactor commit pulling the duplicated ManifestArray-building logic out of `ZarrParser` and `ZippedZarrParser` into two shared helpers in `parsers/zarr.py`: `parse_array_layout` (metadata/format/separator/grid-shape prologue) and `chunk_entries_to_manifest` (the sparse-entries → dense-manifest scatter). Chunk *discovery* stays parser-specific. `IcechunkParser`'s streaming scatter loop is intentionally not unified — see the commit message.

PR #1071, opened 2026-08-05
fix: only remap variable-length HDF5 strings to StringDType

maxrjones:
This PR addresses a regression from #988. It scopes the dtype remapping from #988 to only variable-length strings in order to retain support for fixed-length strings. On main, fixed length strings became unreadable through the virtual store (`ValueError: corrupt buffer`). Fixed-length fill values are now kept as bytes rather than decoded to `str`.

It also warns at parse time that variable-length string variables cannot be read back through the resulting virtual store, since their chunks store references into the HDF5 global heap rather than the string data itself.

Sorry for not tuning in before #988 was merged.

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


TomNicholas:
(Claude on my behalf, since I'm just trying to get a release out quick)

Reviewed and tested locally — this does fix the regression: `test_fixed_length_bytes_roundtrip` fails on `main` and passes here, and the strict UTF-8 decode on the vlen path plus keeping fixed-length fill values as bytes both look right to me. Thanks for catching it.

I pushed one commit (7c816bc) rather than block the release on a round trip, shout if you disagree with it.

`h5py.check_vlen_dtype` reports `str` for a vlen string dtype with a utf-8 cset but `bytes` for one with an ascii cset, so `is str` misses ASCII vlen strings and they fall through to the object-dtype branch:

```
NotImplementedError: HDF5 object dtype dtype(O) is not a variable-length string
and is not yet supported by HDFParser. h5py exposes vlen arrays ...
```

which misdiagnoses the file, and — since it raises at parse time — makes the *whole* file unvirtualizable without `drop_variables`, where on `main` only that one variable was unreadable. netCDF4 string variables are utf-8 so they were unaffected, but plenty of hand-rolled/older HDF5 products use the ascii cset.

The commit widens the check to `h5py.check_vlen_dtype(dtype) in (str, bytes)` in both `_construct_manifest_array` and `_get_fill_value`, so ASCII vlen strings take the same warn-and-remap path, and adds an `ascii_vlen_string_hdf5_url` fixture plus a test asserting it warns rather than raising.

Also: the `test-py312` failure is a 300 s `pytest-timeout` in `test_numpy_arrays_to_inlined_kerchunk_refs`, unrelated to this change.

TomNicholas (approved):
Reviewed and tested locally — fixes the #988 regression, and the HDF suite is green. The one CI failure was the pre-existing kerchunk/zarr deadlock now tracked in #1072, and it passed on rerun. Thanks for the quick turnaround on this.

PR #1086, opened 2026-08-13
Re-implement netCDF3 parser without Kerchunk

### LINKED ISSUE #982: netcdf3 parser fails to preserve _FillValue attribute
`virtualizarr.parsers.netcdf3` doesn't copy the `_FillValue` attribute into the Zarr metadata. Consequently, after opening the virtual icechunk store in xarray, I see the sentinel value rather than NaNs.

I found this line in kerchunk, which explicitly excludes `_FillValue` from being copied over:
https://github.com/fsspec/kerchunk/blob/main/kerchunk/netCDF3.py#L221

According to https://hackmd.io/@zarr/SJkDiK9Teg , in Zarr V2 `Array.fill_value` and `attrs['_FillValue']` were conflated, which probably explains kerchunk's behavior, but in Zarr V3 they are distinct.

TomNicholas:
Replaces the current implementation of the `NetCDF3Parser` with a new standalone implementation which does not require kerchunk or scipy. This one is faster, can understand a wider range of netCDF3 format variants, fixes at least one bug (#982), and has fewer dependencies.

🤖 

NetCDF3Parser went through `kerchunk.netCDF3.NetCDF3ToZarr`, which subclasses `scipy.io._netcdf.netcdf_file` and overrides a private method. Only the file header is ever needed, and the netCDF classic header is a small, fully specified big-endian grammar that ends every variable's record with that variable's byte offset — exactly what a ChunkManifest needs.

Parsing it directly drops kerchunk and scipy as dependencies (virtualizarr[netcdf3] is now empty and no longer pulls in virtualizarr[remote]), reads the header through the ObjectStoreRegistry like every other native parser, and adds CDF-5 support, which scipy cannot read at all.

Record variables are interleaved on disk, so they produce one chunk reference per record per variable. These are now built as vectorized numpy arrays straight into a ChunkManifest instead of going through an intermediate kerchunk refs dict. On a 6.4 MB file with 20 record variables × 4000 records, open_virtual_dataset goes from ~997 ms to ~63 ms (15.8x), with all 80,000 chunk references identical to the kerchunk path.

Closes #982 — kerchunk excludes _FillValue from the attributes it copies, so sentinels survived into loaded data instead of decoding to NaN.

Tests cover all three CDF versions against what xarray reads natively, plus the corners the format hides: CDF-5's 8-byte dimension ids, the unpadded record stride when a file has exactly one record variable, streaming numrecs recovered from file size, char and scalar variables, and malformed headers.


docs/explanation/faq.md
```diff
@@ -251,7 +251,7 @@ Users of Kerchunk may find the following comparison table useful, which shows wh
...
 | From a netCDF4/HDF5 file                                                 | `kerchunk.hdf.SingleHdf5ToZarr`                                                                                                     | `open_virtual_dataset(..., parser=HDFParser())`                                                            |
-| From a netCDF3 file                                                      | `kerchunk.netCDF3.NetCDF3ToZarr`                                                                                                    | `open_virtual_dataset(..., parser=NetCDF3Parser())`, via `kerchunk.netCDF3.NetCDF3ToZarr`                                                                                     |
+| From a netCDF3 file                                                      | `kerchunk.netCDF3.NetCDF3ToZarr`                                                                                                    | `open_virtual_dataset(..., parser=NetCDF3Parser())` (reads the header directly; also supports CDF-5, which kerchunk/scipy cannot)                                                                                     |
```
TomNicholas:
```suggestion
| From a netCDF3 file                                                      | `kerchunk.netCDF3.NetCDF3ToZarr`                                                                                                    | `open_virtual_dataset(..., parser=NetCDF3Parser())` (also supports [CDF-5](https://docs.unidata.ucar.edu/nug/current/netcdf_introduction.html#:~:text=CDF%2D2%20files.-,NetCDF%2064%2Dbit%20Data%20Format%20(CDF%2D5),-To%20allow%20large), which kerchunk/scipy do not)                                                                                     |
```

kthyng:
@TomNicholas This is amazing. An hour after you made this PR I was about to start figuring out how to deal with CDF5 files in kerchunk, which this appears to fix (I just tried it out). Thank you so much!!!

PR #1103, opened 2026-09-22
Decode HDF5 bytes-array attributes to lists of strings

ianhi:
# What I did

`HDFParser` fails on any HDF5 file with an attribute stored as an array of fixed-length byte strings:

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

`_extract_attrs` converted these with `v.astype(str)`, which leaves a numpy array; the `v == "DIMENSION_SCALE"` check then compares elementwise and raises. (An array attribute would also fail later: zarr can't serialize an ndarray into `zarr.json`.) Instrument software such as Imaris writes every attribute this way, so `HDFParser` can't open those files at all.

Repro:

```python
import h5py, numpy as np, obstore

with h5py.File("/tmp/charattr.h5", "w") as f:
    f.create_dataset("data", data=np.zeros((4, 4)))
    f["data"].attrs["character_array"] = np.array([b"2", b"0", b"4", b"2"], dtype="|S1")

registry = ObjectStoreRegistry({"file:///": obstore.store.LocalStore("/")})
HDFParser()(url="file:///tmp/charattr.h5", registry=registry)  # ValueError on main
```

The fix decodes each element as UTF-8 (`astype(str)` decodes as ASCII) and returns a list of `str` with the array's shape: `["2", "0", "4", "2"]` above.

## Why a list, not a joined string

A virtual dataset's attributes should match what users get opening the file directly. For this attribute:

| Reader | Result |
| --- | --- |
| netCDF-C (`ncdump`) | `string data:character_array = "2", "0", "4", "2"` |
| netCDF4-python / xarray `engine="netcdf4"` | `['2', '0', '4', '2']` |
| h5netcdf / xarray `engine="h5netcdf"` | `[b'2', b'0', b'4', b'2']` |
| this PR | `['2', '0', '4', '2']` |

This also follows HDF5: a string datatype of size 1 with a `(4,)` dataspace is four one-byte strings, and nothing in the file says they form one. netCDF itself writes text as a scalar string of size *n*, and netCDF-C reads that form as text; this PR leaves the scalar path unchanged. Joining one-character arrays is a convention of particular writers (Imaris), so it's left to format-specific code.

Multi-dimensional attributes keep their shape as nested lists, as the numeric branch already does (netCDF4 flattens them instead).

- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


ianhi:
For example on an imaris file I found in the idr

```python

base = "https://ftp.ebi.ac.uk/pub/databases/IDR/"
url = base + "idr0066-voigt-mesospim/20190802-ftp/ExperimentB/data/ExpB_whole_cns_spinal_cord.ims"
registry = ObjectStoreRegistry({base: HTTPStore.from_url(base)})
store = HDFParser(group="DataSetInfo/Image")(url=url, registry=registry)
```

On main this raises the ValueError above. With this PR it returns X: ['2', '0', '4', '2'], Y: ['2', '0', '4', '9'], Z: ['1', '0', '2', '4'], Unit: ['u', 'm'].

Which should totally be concatenated into strings for what they mean. But this is the spec compliant interpretation, part of the virtualization here maybe should be to update them from `'2', '0', '4', '2'` to `'2042'`

virtualizarr/tests/test_parsers/conftest.py
```diff
@@ -158,6 +158,18 @@ def string_attributes_hdf5_url(tmpdir):
     return f"file://{filepath}"
 
 
+@pytest.fixture
+def bytes_attributes_hdf5_url(tmpdir):
+    filepath = f"{tmpdir}/bytes_attributes.nc"
+    f = h5py.File(filepath, "w")
```
TomNicholas: should we use our new context manager here?

TomNicholas (approved):

ianhi:
added the context manager to all the hdf opens to protect everywhere. shoulda done that on the last pr honestly

TomNicholas (approved):

PR #1105, opened 2026-09-23
Add docs about the registry

### LINKED ISSUE #629: Document the object store registry

### LINKED ISSUE #1050: Docs could explain the registry better
@ianhi said Kanai (new user at SciPy sprints) said "I'm confused what a registry is the docs don't explain this very well" which is totally fair, we should improve that.

ianhi:
# What I did
Added docs on what a registry is and why you need it. I used the GOES-East transition becuase that seemed reasonable, but I don't know that much about satteilltes so I may have done somethign non-sensical.

I was confused about the purpose of the registry. Together with the robot i have taught myself what I think it is, and why we need it. These docs are the result of that.

Acceptance criteria:

- [x] Closes #1050 
- [x] Closes #629


docs/explanation/registry.md
```diff
@@ -0,0 +1,238 @@
...
+    region="us-east-1",
+    # public bucket, so send anonymous (unsigned) requests
+    skip_signature=True,
+)
+```
+
+## 2. Building the time series
+
+A time series that spans the handover needs at least one file from each side of it.
+This example uses two files, taken at 12:00 UTC on 6 April and 8 April 2025, but the same code works for any number of files.
```
TomNicholas: Slight de-clauding
```suggestion
This example uses two files, taken at 12:00 UTC on 6 April and 8 April 2025, but the same code works for any number of files.
These files are chosen as examples of spanning the handover between the two data sources.
```

docs/explanation/registry.md
```diff
@@ -0,0 +1,238 @@
...
+    f"{goes19_bucket}/ABI-L2-SSTF/2025/098/12/"
+    "OR_ABI-L2-SSTF-M6_G19_s20250981200209_"
+    "e20250981259517_c20250981304588.nc"
+)
+```
+
+Each file has to be read through the store for its own bucket.
+If you opened the files one at a time in a loop, you could pick the store for each URL yourself.
+A registry does that matching for you, and it's the only way to do it when a single call opens many files.
+Register each store under the URL prefix it serves, usually its bucket, then pass the registry to [`open_virtual_mfdataset`][virtualizarr.open_virtual_mfdataset], which opens both files and joins them along time.
```
TomNicholas: I feel like most of the content from this point on is superfluous...

TomNicholas (commented):
I like this a lot, but think we could delete about half of it and it still be similarly useful.

docs/explanation/registry.md
ianhi: I think up until here is a nice high level "why", and the rest is definitely more in teh weeds. Reading it back I think I find it all useful (imagingin myself prior to understanding) but perhaps a bit verbose

I think at a minimum I would want to keep up through the open_mf_dataset, the reading back later, and at least a short line about local.

ianhi:
I took a pass over it, re-wrote (by hand!) some sections, and moved things around and deleted stuff

ianhi:
Attn @8bitbiscuit who I think is Kanai? Would these docs have been adeuqate for you at scipy?

PR #1107, opened 2026-09-23
Fix intermittent kerchunk test deadlock caused by leaked h5py objects

ianhi:
Fixes the intermittent timeout in `test_numpy_arrays_to_inlined_kerchunk_refs` (`minimum-versions` job).

`test_cf_fill_value` left an open `h5py.File` alive after `pytest.xfail`. When those objects were later garbage-collected on zarr's IO thread during kerchunk's `translate()`, h5py's global lock deadlocked. This closes the file and drops the reference before any `xfail`, so no h5py objects outlive the test.

The leak is old, but it only started causing hangs on Aug 5, when #988 and #1065 changed test counts and pytest-xdist began scheduling `test_hdf.py` on the same worker just before `test_integration.py`.


virtualizarr/tests/test_parsers/test_hdf/test_hdf.py
```diff
@@ -143,7 +133,12 @@ def test_scalar_fill_value(self, scalar_fill_value_hdf5_url):
 
     def test_cf_fill_value(self, cf_fill_value_hdf5_file):
         cf_fill_value_hdf5_url = f"file://{cf_fill_value_hdf5_file}"
-        dtype = _hdf5_dataset_dtype(cf_fill_value_hdf5_file, "data")
+        with h5py.File(cf_fill_value_hdf5_file) as f:
+            dtype = f["data"].dtype
+        # A pytest.xfail traceback would keep `f` alive in a reference cycle, and
+        # h5py objects freed by the cyclic GC on another thread can deadlock on
+        # h5py's global lock (e.g. during kerchunk's SingleHdf5ToZarr.translate).
+        del f
```
TomNicholas: Can't this go in a context manager? So that it can be used whenever we need to create a file with h5py?

Issue #799, opened 2025-09-19
Add in-memory arrays to `ManifestStore`


TomNicholas:
**tl;dr: I think we should add in-memory arrays to `ManifestStore`, probably by generalizing the wrapped type to `ManifestArray | np.ndarray`.**

### Context

This library has evolved from managing only "virtual" xarray datasets into a design where `ManifestStore` really should be the primary entity, with [xarray as an optional dependency](https://github.com/zarr-developers/VirtualiZarr/issues/521). Icechunk and Kerchunk are already optional dependencies, and `ManifestStore` should be able to stand alone as [independently useful](https://github.com/zarr-developers/VirtualiZarr/issues/603).

However, `ManifestStore` is currently lacking a way to represent "inlined" chunks. Comparing different representations of virtual/non-virtual chunks, this absence becomes conspicuous:

| - | `ManifestStore` | "virtual" `xarray.Dataset` | Icechunk format | Kerchunk format |
| --- | --- | --- | --- | --- |
| external references | `ManifestArray`  | `xr.Variable` wrapping `ManifestArray` (a so-called "virtual variable") | virtual chunk ref | kerchunk reference stored in JSON |
| "inlined" references | ❌ | `xr.Variable` wrapping (lazy) `np.ndarray` (a so-called "loadable variable") | native chunk ref (also potentially inlined into manifests, but this is an optimization that's an internal implementation detail) | encoded bytes stored in JSON |

This causes multiple problems:
- We can't roundtrip `kerchunk -> ManifestStore -> virtual xarray.Dataset -> kerchunk` if the original kerchunk file contains inlined references. (#489)
- Similarly it limits the utility of proposed `.to_icechunk`/`.to_kerchunk` methods on `ManifestStore`, as they can't handle all the cases that `ds.vz.to_icechunk/kerchunk` can. (#591)
- It tempts us to hack around this using a memorystore (see https://github.com/zarr-developers/VirtualiZarr/issues/636), but I think this is the wrong design...

### Proposal: Add `numpy.ndarray` as a wrapped type in `ManifestStore`

Currently `ManifestStore` holds `ManifestGroups`, which have `._members: Mapping[str, "ManifestArray | ManifestGroup"]`. If we generalize this to `._members: Mapping[str, "np.ndarray | ManifestArray | ManifestGroup"]` we can store inlined chunks in memory in the `ManifestStore` as numpy arrays.

[...]

PR #1108, opened 2026-09-23
Add `ManifestStore.to_icechunk` to write without xarray

ianhi:
## What I did

Adds `ManifestStore.to_icechunk()` and `ManifestGroup.to_icechunk()`, so a parser's output can go straight to Icechunk instead of through a virtual xarray dataset. Part of #521; picks up the idea of @TomNicholas's draft #591.

```python
HDFParser()(url=url, registry=registry).to_icechunk(session.store)
```

Each array is written with the metadata on its `ManifestArray` (dimension names, attributes, codecs and fill value), and each group with its `ManifestGroup` attributes. Nothing passes through xarray, so it's possible to write stores that are valid Zarr but not valid xarray, each covered by a test:

- arrays without dimension names
- sibling arrays sharing a dimension name at different lengths, like pyramid levels
- a subgroup reusing a parent's dimension name at a different length

For files that do fit xarray, the output matches `vds.vz.to_icechunk`. A test writes three netCDF files both ways, then compares every array's metadata and the data read back. The only difference is the `coordinates` attribute that xarray adds to each group.

**Behaviour change:** `mode="a"` now raises if an existing array of the same name has different metadata. zarr's `require_array` checks only shape and dtype, then keeps the stored metadata, so new references were decoded with the old codecs. Rewriting a little-endian array with big-endian references read back `[0, 16777216, 33554432, 50331648]` instead of `[0, 1, 2, 3]`. The xarray path had the same bug and now shares the check. Identical rewrites and attribute merges still work.

Not included:

- Kerchunk.
- `append_dim` and `region`, which are defined by dimension name and stay on the xarray API.
- Loadable variables. A follow-up can inline arrays in a `ManifestStore` using the inlined chunks from #851 (see #799).

`import virtualizarr` still loads xarray. Making it optional is the package split discussed in #521.

Docs: a usage section on writing without xarray, and a data structures section on how the Zarr and xarray data models differ.

- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


docs/about/releases.md
```diff
@@ -19,6 +27,13 @@
   rooted at the filesystem root (`file:///`). The store root URI was stripped of every
   trailing slash, mangling the scheme so Icechunk's virtual chunk container prefix never
   matched. By [Tom Nicholas](https://github.com/TomNicholas).
+- Writing to Icechunk with `mode="a"` now raises a `ValueError` when an array of the same
```
TomNicholas, line 30: Maybe we should split this out into a pre-requisite PR?

docs/explanation/data_structures.md
```diff
@@ -275,3 +275,40 @@ See the [usage guide on combining virtual datasets](../how_to/usage.md#combining
...
+| An array that repeats a dimension name, such as `("x", "x")` | Valid | Warns "Duplicate dimension names present", and many operations don't support it |
+| Sibling arrays that share a dimension name at different lengths, like the levels of a multiscale image pyramid | Valid: each array's shape is independent | A Dataset gives each dimension one length, so this raises "conflicting sizes for dimension" |
+| A subgroup that reuses one of its parent's dimension names at a different length | Valid | An [xarray.DataTree][] requires each node to align with its parents, so this raises "not aligned with its parents" |
+
+Parsers can still represent all of these, because a `ManifestStore` follows the Zarr model.
+Converting such a store to a virtual dataset or datatree either fails or, for repeated dimension names, warns. You can instead write it to Icechunk directly with [ManifestStore.to_icechunk][virtualizarr.manifests.ManifestStore.to_icechunk], as shown in [the usage guide](../how_to/usage.md#writing-to-icechunk-without-xarray).
+
+### Where each representation keeps metadata
```
TomNicholas, line 297: Remove this section, its basically implementation details.

docs/explanation/data_structures.md
```diff
@@ -275,3 +275,40 @@ See the [usage guide on combining virtual datasets](../how_to/usage.md#combining
...
+| CF encoding, such as `scale_factor`, `_FillValue` or time `units` | Array attributes, with values left encoded | Virtual variables: the same attributes, still not decoded. Loaded variables: decoded, with the encoding moved to `Variable.encoding` |
+
+When a `ManifestGroup` becomes a virtual dataset, each array's dimension names and attributes move onto its xarray variable, and the wrapped `ManifestArray` no longer carries them.
+xarray then keeps track of them as you combine datasets, and `vds.vz.to_icechunk` writes them from the variables.
+`ManifestStore.to_icechunk` instead writes each `ManifestArray`'s own metadata, so if you combine `ManifestArray` objects yourself with functions such as `np.concatenate`, the dimension names they hold are the ones that get stored.
+
+### Loaded variables only exist in xarray
+
+[Loading a variable](faq.md#why-would-i-want-to-load-variables-using-loadable_variables) replaces its `ManifestArray` with an in-memory array, which only an xarray Dataset can hold alongside virtual variables.
+A `ManifestStore` has no loaded variables. Everything it writes stays a reference, apart from any chunks the parser inlined.
```
TomNicholas, line 314: This is not really true, and is an important subtlety / edge case to understand / test.

Actually there are 3 types of chunk, virtual, native, and inline.

You already know how Icechunk stores these. In the Xarray virtual dataset representation virtual and inline refs are both handled by `ManifestArray`, and can be mixed arbitrarily. This is a relatively recent feature so may not be fully explained in the docs. "Loadable variables" are separate, they are another way to store array values in memory, but it's all or nothing, the whole array is either loaded or its not.

Before we added the inline chunks feature (see #938), there was no way to store in-memory array values on a `ManifestStore`. Now it can be done, as inline chunks. `ManifestStore` has no equivalent of the entire array being loaded as numpy. We may want to add that to `ManifestStore`, since it would be more efficient (see #799), but we can probably get away without it for now.

Regardless, we should test roundtripping `ManifestStores` that have inline chunks as well as pure virtual refs.

docs/how_to/usage.md
```diff
@@ -471,6 +471,79 @@ print(snapshot_id)
...
+```python exec="on" session="usage" source="material-block" result="code"
+manifest_store = HDFParser()(url=url, registry=registry)
+
+session = repo.writable_session("main")
+manifest_store.to_icechunk(session.store, group="without_xarray")
+snapshot_id = session.commit("Wrote the file without going via xarray")
+print(snapshot_id)
+```
+
+Nothing is loaded on this path, so references stay virtual and keep the file's own chunking, including for the coordinate variables that `open_virtual_dataset` loads by default.
```
TomNicholas, line 494: Delete this paragraph

docs/how_to/usage.md
```diff
@@ -471,6 +471,79 @@ print(snapshot_id)
...
+Appending and region writes are only available through the virtual dataset API, since both are defined in terms of dimension names.
+To write a single group rather than a whole store, use [ManifestGroup.to_icechunk][virtualizarr.manifests.ManifestGroup.to_icechunk].
+
+##### Loading arrays without xarray
```
TomNicholas, line 498: 5 levels of indentation? 🤔

docs/how_to/usage.md
```diff
@@ -471,6 +471,79 @@ print(snapshot_id)
...
+```
+
+Merging chunks works the same way: to load a coordinate stored as one chunk per file into a single chunk, pass its full length as `chunks=`.
+
+!!! important
+    `zarr.from_array` copies the fill value and attributes only from zarr 3.4 onwards.
+    With older versions, pass them yourself, as in `fill_value=lat_source.fill_value, attributes=lat_source.attrs.asdict()`.
+
+!!! note
+    `ManifestStore` and `ManifestGroup` are experimental, and their API may change without notice.
```
TomNicholas, line 545: This is not true I don't think

docs/how_to/usage.md
```diff
@@ -471,6 +471,79 @@ print(snapshot_id)
...
+Appending and region writes are only available through the virtual dataset API, since both are defined in terms of dimension names.
+To write a single group rather than a whole store, use [ManifestGroup.to_icechunk][virtualizarr.manifests.ManifestGroup.to_icechunk].
+
+##### Loading arrays without xarray
+
+Loading an array copies its data into Icechunk, so reading it no longer touches the archival file.
+This is helpful for small arrays that are read often, such as coordinates, especially when they are split into one chunk per archival file and would be better stored as a few larger chunks.
+
+To load an array, copy it out of the `ManifestStore` with [zarr.from_array][], passing `overwrite=True` to replace the virtual array.
+The copy has the same chunks, codecs and dimension names as the source:
```
TomNicholas, line 504: This needs checking in light of my above comment about inline chunks.

docs/about/releases.md
ianhi: done #1109

docs/explanation/data_structures.md
ianhi: this was very helpful thanks. It's a little bit head spinning that I can virutalize icechunk with inline chunks, that become inline in teh manifestarrays, but that if I then persist them to icechunk they will not necessarily become inline chunks (because the being persisted to store sets its own inline chunk threshold)

TomNicholas: Yeah it is complicated! Really we have 4 totally different models for storing inline/native/virtual chunks: Icechunk, Kerchunk, VirtualiZarr `ManifestStore`, and VirtualiZarr "virtual" `xarray.Dataset`.

